PR #48, opened 2022-09-19
chore: Do some cleanup and improve node_modules caching in the actions

martin-trajanovski:
## Description

This is just a cleanup PR that will remove unnecessary dependencies and files.

## Motivation and Context

The `cypress` under the root directory was copied with the last pull from the `frontend` repo and it is not needed. Also `faker-js` is not needed as a dependency in the `apps/user-office-frontend` anymore because the `e2e` tests are extracted as a separate app.


jekabs-karklins (approved):

jekabs-karklins (approved):
🆗

PR #49, opened 2022-09-19
fix: Add new status to the message that is sent to RabbitMQ (SWAP-2801)

jekabs-karklins:
## Description
This PR adds the `newStatus` field to the message, PROPOSAL_STATUS_CHANGED_BY_WORKFLOW, and PROPOSAL_STATUS_CHANGED_BY_USER. so that the consumer makes better decisions for its logic.

This PR also changes so that all events about proposal status changes are sent out to the Queue.
Message consumer itself is responsible to figure out if and how to react.

## Motivation and Context

Scichat wants to know not only that the status of the proposal has changed, but also what is the new status

## Fixes

https://jira.esss.lu.se/browse/SWAP-2801

## Depends on

N/A


apps/user-office-backend/src/eventHandlers/messageBroker.ts
```diff
@@ -145,33 +144,22 @@ export function createPostToRabbitMQHandler() {
         const proposalStatus =
           await proposalSettingsDataSource.getProposalStatus(proposal.statusId);
 
-        // if the new status isn't 'SCHEDULING' ignore the event
```
martin-trajanovski, line 148: This change will require a small modification in the scheduler messageBroker as well where we handle the events of type: `PROPOSAL_STATUS_CHANGED_BY_WORKFLOW` and `PROPOSAL_STATUS_CHANGED_BY_USER`. I think this is the proper way of working with events where the sender doesn't care for some specific cases and it is just sending the event with all needed information. The receiver is the one that should check if there is some action needed based on the event type or event message.

package.json
```diff
@@ -53,6 +53,7 @@
     "wait-on": "^6.0.1"
   },
   "devDependencies": {
+    "eslint": "^8.23.1",
```
martin-trajanovski, line 56: Not sure if we really need eslint here in the root repository. I think when lint script is executed it is actually calling lint in every app where we should have eslint installed already as their own dependency. Maybe I am wrong but I think this worked fine so far.

martin-trajanovski (commented):
I left some review comments that will need a small attention.

package.json
jekabs-karklins: I will remove it

martin-trajanovski (approved):

mutambaraf (approved):

PR #51, opened 2022-09-20
feat(auth): Implement OAuth with OpenId Connect authorization

jekabs-karklins:
## Description
This PR aims to add support for authentication via OAuth and OpenId connect.

## Motivation and Context
This PR moves the authentication responsibility out of the user's office because it is expected that each facility has its own user registry and user management is out of the user office software scope. Authentication logic is moved behind the "Authorization interface" (A), and a new type of implementation for OAuth has been added (B)
image
This way we can support any number of Authorization service types.

With this change, it is easy to add support for different third-party authorization services that support open standard oAuth 2.0. To add a new provider in the .env file you need to specify
- AUTH_DISCOVERY_URL (URL containing oAuth configuration (https://openid.net/specs/openid-connect-discovery-1_0.html#WellKnownRegistry))
- AUTH_CLIENT_ID
- AUTH_CLIENT_SECRET
An example of how it could be done for Google accounts can be seen in example.development.env

IMPORTANT:
After the merge, the following entries need to be added to your .env file
```
AUTH_DISCOVERY_URL=http://localhost:5000/.well-known/openid-configuration
AUTH_CLIENT_ID=useroffice
AUTH_CLIENT_SECRET=useroffice
```

## How Has This Been Tested
 Manual tests
 E2E tests
Fixes https://jira.esss.lu.se/browse/SWAP-2688

## Changes
- Removed feature flag `EXTERNAL_AUTH`, this is because the user office should not support user management, and all authentication/authorization should be done against external service
- Storing only a subset of user information in JWT. Until now almost all information from the user object was stored in the JWT. Many fields were not used, and as the object grew it hit the maximum size of the cookie 4096 bytes. There is now a new object UserJWT which is smaller.
- The classes that extend UserAuthorization all had duplicated constructor args. Resolving dependencies from the container instead of injecting this allowed us to clean up the duplicated code.
- ESS Authorization has been removed in favor of reusable OAuthAuthrozation
- After the removal of the EXTERNAL_AUTH feature E2E tests that were skipping tests based on that feature flag, needed adjustments. Now they are depending on different more precise feature toggles.


ACLay:
So this is an awful lot of code to try and review. Even ignoring the autogenerated `sdk.ts` `package-lock.json` `UOWSSoapInterface.ts` changes still leaves over 4000 lines...

So I guess I've a couple comments off the bat:
- Is there any way it can be broken up into smaller chunks?
- Why are there nearly 9000 lines of changes to `UOWSSoapInterface.ts`? I'm just not sure how you'd have done that since the npm command to regenerate it requires being on our network to access the live service.

jekabs-karklins:
> So this is an awful lot of code to try and review. Even ignoring the autogenerated `sdk.ts` `package-lock.json` `UOWSSoapInterface.ts` changes still leaves over 4000 lines...
> 
> So I guess I've a couple comments off the bat:
> 
> * Is there any way it can be broken up into smaller chunks?
> * Why are there nearly 9000 lines of changes to `UOWSSoapInterface.ts`? I'm just not sure how you'd have done that since the npm command to regenerate it requires being on our network to access the live service.

Hi, @ACLay ,

This PR is just transferred from the old repo, which you have already looked at so, there should not be much new, I have resolved some merge conflicts since then.
 
We both decided to make the PR bigger after you asked it, so no complaints accepted :)
https://github.com/UserOfficeProject/user-office-frontend/pull/977#pullrequestreview-1082512767

Generally, this PR has been in review for more than two months, and since the people change the development branch, they are changing code without this feature and conflicts are more likely, and therefore there are more changes. It takes time to keep PR mergeable.

The way forward I see is that we both could add some time this week to make sure it is in a good state for merging. Until Wednesday I will 
- make sure I pull the latest changes from development, 
- fix the conflicts, 
- make sure tests are passing 
- fix issues with`UOWSSoapInterface.ts`, 
- add details in the description of the main changes, and reasons for it

Will you have some time from Wednesday - Friday to look at this PR? So that we can merge it in and we do not have to return to this PR?

jekabs-karklins:
@ACLay  I've done the things I wrote in the morning, please look over them again once you have time :)

ACLay:
I'm looking through this now.

apps/user-office-backend/tsconfig.json
```diff
@@ -11,6 +11,8 @@
       "esnext.asynciterable"
     ] /* Specify library files to be included in the compilation. */,
     "allowJs": true /* Allow javascript files to be compiled. */,
+    "skipLibCheck": true,
+    "skipDefaultLibCheck": true,
```
ACLay, line 15: These don't sound like safe options to be adding.

[The docs for skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck) suggest using alternative dependency resolution tools instead of it.

[The docs for skipDefaultLibCheck](https://www.typescriptlang.org/tsconfig#skipDefaultLibCheck) say to use skipLibCheck instead, so at a minimum I don't think we need both.

ACLay (commented):
I still don't understand why there are multiple commits adding up to nearly 9000 lines of changes for `UOWSSoapInterface.ts` It's auto generated, set to be ignored by [prettier](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-backend/.prettierignore) & [eslint](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-backend/.eslintignore), and [it's generator](https://github.com/UserOfficeProject/user-office-lib/tree/master/packages/uows-client-generator) hasn't been altered in since July.

apps/user-office-frontend/cypress.json
```diff
@@ -2,7 +2,8 @@
   "video": false,
```
martin-trajanovski, line 2: Probably we don't need this cypress.json file. I think we use the one that is in the apps/user-office-frontend-e2e/ and this one is leftover after moving to the monorepo.

jekabs-karklins:
> I still don't understand why there are multiple commits adding up to nearly 9000 lines of changes for `UOWSSoapInterface.ts` It's auto generated, set to be ignored by [prettier](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-backend/.prettierignore) & [eslint](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-backend/.eslintignore), and [it's generator](https://github.com/UserOfficeProject/user-office-lib/tree/master/packages/uows-client-generator) hasn't been altered in since July.

Looks like it was prettier causing the differences, it's fixed now.

apps/user-office-backend/tsconfig.json
jekabs-karklins: `skipLibCheck` is added for performance. They have been set for `user-office-frontend` and `user-office-frontend-e2e` tsconfigs already. It does not cause problems with development experience.

You are right the skipDefaultLibCheck should be removed.

jekabs-karklins:
> > I still don't understand why there are multiple commits adding up to nearly 9000 lines of changes for `UOWSSoapInterface.ts` It's auto generated, set to be ignored by [prettier](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-backend/.prettierignore) & [eslint](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-backend/.eslintignore), and [it's generator](https://github.com/UserOfficeProject/user-office-lib/tree/master/packages/uows-client-generator) hasn't been altered in since July.
> 
> Looks like it was prettier causing the differences, it's fixed now.

It turned out that [.prettierignore is not taken into account if it is not in the root folder](https://github.com/prettier/prettier/issues/4081). For this reason,[ I have moved it up](https://github.com/UserOfficeProject/user-office-core/pull/51/commits/42ffbcd1d1cce5f659cd6ec70c504acc6973b801). So it works as expected.

apps/user-office-frontend/cypress.json
jekabs-karklins: done

apps/user-office-backend/tsconfig.json
jekabs-karklins: Done

.prettierignore
```diff
@@ -0,0 +1,12 @@
+**/package-lock.json
+**/node_modules/
+**/.github/workflows/
+**/README.md
+**/.vscode/
+**/src/generated/
+
+/apps/user-office/backend/src/datasources/stfc/UOWSSoapInterface.ts
+/apps/user-office-backend/src/datasources/stfc/UOWSSoapInterface.ts
```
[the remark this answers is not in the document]
jekabs-karklins: Thanks

apps/user-office-backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -226,15 +218,23 @@ export class StfcUserAuthorization extends UserAuthorization {
...
+        await client.logout(token).catch(() => {
+          logger.logWarn('Failed to log out user', { token });
 
-    return;
+          return rejection('Failed to log out user', { token });
+        });
+      } else {
+        return rejection('No external token found in JWT', { token });
+      }
+    } catch (error) {
```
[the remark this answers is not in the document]
ACLay: It looks like that external try/catch block is still there: https://github.com/UserOfficeProject/user-office-core/pull/51/files#diff-3203c957159c45948037f11a5138ff7deaa651464b70bd47c4dc0ef2f1fa6642R483

so I do think it should be removed from here.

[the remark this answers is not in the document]
jekabs-karklins: External try..catch has to be there because it calls 
`verifyToken(token);` which will throw if the token is invalid

I would also leave the inner try..catch because the logout is a public function and can be called from multiple places. The return type is void | Rejection, so from the API point of view, I would expect to receive a back Rejection, instead of the exception if something fails.

From what I can see the flow is kept the same as it was before, which was the point of this small refactor.

What do you think? Shall we keep it as is?

ACLay (approved):

martin-trajanovski (approved):
This seems quite OK to me.

PR #59, opened 2022-09-29
refactor: Re-fetch all proposals if instrument or SEP is selected and improve the selection in the proposals table

martin-trajanovski:
## Description

1. Re-query all the proposals in the officer proposal table after assigning to instrument or SEP. This was needed because of the proposal status change that can happen in the workflow which runs asynchronously.
2. Remove `nextproposalstatus` from the response queries and do some cleanup based on the changed approach.
3. Improve user officer proposal table selection because previously there was some weird logic that was selecting just first page of proposals + 1

## Motivation and Context

Proposal workflow is running asynchronously after the action is taken on a proposal so it wasn't so easy to return the correct next proposal status. Previously we made some predictions based on the proposal workflow and return the next status but it wasn't the nicest solution because we were kind of running the workflow twice for every proposal.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2812

## Changes

https://user-images.githubusercontent.com/6333063/196171828-922b6c44-e9a1-4015-bc4d-0ab00cd5b498.mp4


ACLay (commented):
What's caused so much change to the frontend's sdk.ts file? There's 11 lines of changes to the .graphql files but 30,307 lines of change in the sdk

apps/user-office-frontend/src/components/review/ProposalTableReviewer.tsx
```diff
@@ -213,19 +210,25 @@ const ProposalTableReviewer: React.FC<{ confirm: WithConfirmType }> = ({
...
+    const { review: freshReview } = await api().getReview({
+      reviewId: reviewId,
+    });
 
+    if (freshReview) {
       const userDataUpdated = {
         ...userData,
         reviews: userData?.reviews.map((review) => {
-          if (review.id === currentReview?.id) {
+          if (review.id === freshReview?.id) {
```
ACLay, line 226: It seems odd to me that this line is using a `?.` null check on `freshReview` when we're already inside an `if (freshReview)` block. Would `freshReview.id` be enough here?

martin-trajanovski:
> What's caused so much change to the frontend's sdk.ts file? There's 11 lines of changes to the .graphql files but 30,307 lines of change in the sdk

@ACLay As you can see from this commit (https://github.com/UserOfficeProject/user-office-core/pull/59/commits/2649ac04092f34d07f23397c36a4653d8754a6a5#diff-725cffd6237469fb511f59cdb6ed1bbaa104622f68ebf3b4ffbc584a19749ddc) there were just a few changes in the sdk. Once I merged this bigger PR for implementing OAuth it introduced lot of new changes as you can see here (https://github.com/UserOfficeProject/user-office-core/pull/51/files#diff-725cffd6237469fb511f59cdb6ed1bbaa104622f68ebf3b4ffbc584a19749ddc). But I guess we should not take care of the sdk anymore as it will be removed with the following PR https://github.com/UserOfficeProject/user-office-core/pull/69. It will be autogenerated in the actions and everyone should generate its own sdk locally but we wont keep it in the repository anymore.

ACLay (approved):

apps/user-office-frontend-e2e/cypress/integration/generic_templates.ts
```diff
@@ -32,6 +32,11 @@ context('GenericTemplates tests', () => {
   let workflowId: number;
   let createdQuestion1Id: string;
 
+  beforeEach(() => {
```
jekabs-karklins, line 35: I've noticed that if you have `before hook` in context AND in suite then they might cause a race conditions. Last week we made it so to only have `before hook` in the suite

image

jekabs-karklins (commented):
Looks pretty good
Nice cleanup 💯

jekabs-karklins (approved):

PR #62, opened 2022-09-30
fix(instrument availability time): Entering number grater than biggest 32-bit integer crashes the app

martin-trajanovski:
## Description

Prevent sending very big numbers for availability time that are not supported by the database.

## Motivation and Context

If you enter number bigger than 32-bit integer the app crashes and there is a backend error.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2519


jekabs-karklins (approved):
Looks good

PR #69, opened 2022-10-07
Auto SDK for frontend

bolmsten:
## Description

At the moment when a PR is merged that has changes in the SDK it will require all other PR's to be updated manually. This PR removes the sdk file from the repo and creates it on the fly in the CI/CD pipelines.


.github/workflows/github-tag-and-release.yml
```diff
@@ -55,6 +59,12 @@ jobs:
             dmsc/duo-backend:${{ steps.tag_version.outputs.new_tag }}
             ghcr.io/userofficeproject/user-office-backend:${{ steps.tag_version.outputs.new_tag }}
 
+      - name: Install Node.js dependencies
+        run: npm ci
```
martin-trajanovski, line 63: Not quite sure why we need to install the dependencies in this step.

martin-trajanovski (approved):
There is just a small comment but otherwise it looks good to me.

jekabs-karklins:
We will probably want another PR where we add the sdk to the .gitignore

jekabs-karklins (approved):

.github/workflows/github-tag-and-release.yml
bolmsten: to be able to run the generate SDK for the docker image to have the file

martin-trajanovski: Maybe I am missing something here but as I know the image is built using the `Dockerfile` that is inside frontend directory. I can't see any sdk generation in this step but just installation of the dependencies. Not sure what sdk file is used inside the docker image (if we remove the sdk from the repository code later how it will be added inside the image when we are not generating one). I might not see the whole picture here in this process but it is a bit confusing when I see only installation without sdk generation.

bolmsten: you are correct I had missed the SDK generation step, added it now

simonfernandes (approved):
This will be really nice 👍

PR #70, opened 2022-10-07
refactor: Refactor the proposal workflow event handler to be more generic one

martin-trajanovski:
## Description

Instead of using custom handler for each proposal event that we want to trigger a workflow engine run we use more generic way where we trigger the workflow to check for a status change on each proposal events. It recognizes a proposal event by proposal information keys in the event payload.

## Motivation and Context

Previously the proposal workflow handler was handling only events specified. We want the engine to be run on every event and check for a proposal status change instead of adding the event handlers manually.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2808


apps/user-office-backend/db_patches/0127_AddProposalStatusChangedByUserEvent.sql
```diff
@@ -0,0 +1,14 @@
+DO
+$$
+BEGIN
+	IF register_patch('AddProposalStatusChangedByUserEvent.sql', 'martintrajanovski', 'Add proposal_status_changed_by_user event', '2022-10-07') THEN
+        BEGIN
+            ALTER TABLE proposal_events
+            ADD COLUMN proposal_status_changed_by_user BOOLEAN DEFAULT FALSE,
+            ADD COLUMN proposal_status_changed_by_workflow BOOLEAN DEFAULT FALSE,
+            ADD COLUMN proposal_updated BOOLEAN DEFAULT FALSE;
```
jekabs-karklins, line 9: If proposal_status_changed_by_user and proposal_status_changed_by_workflow are mutually exclusive you could add the constraint.

And maybe don't have a default value.

```suggestion
            ALTER TABLE proposal_events
            ADD COLUMN proposal_status_changed_by_user BOOLEAN DEFAULT TRUE,
            ADD COLUMN proposal_status_changed_by_workflow BOOLEAN DEFAULT FALSE,
            ADD COLUMN proposal_updated BOOLEAN DEFAULT FALSE
            CONSTRAINT one_and_only_one_status_changed_by_must_be_true
               CHECK ( 
	               (proposal_status_changed_by_user OR proposal_status_changed_by_workflow) 
		        AND 
	               (proposal_status_changed_by_user AND proposal_status_changed_by_workflow) IS FALSE 
	         );
	         
	     ALTER TABLE proposal_events ALTER COLUMN proposal_status_changed_by_user DROP DEFAULT;
            ALTER TABLE proposal_events ALTER COLUMN proposal_status_changed_by_workflow DROP DEFAULT;
```

jekabs-karklins (approved):

jekabs-karklins:
Nice cleanup :100:

apps/user-office-backend/db_patches/0127_AddProposalStatusChangedByUserEvent.sql
martin-trajanovski: I think this is not the case in our situation. It can be either one or another or even both. I mean this events can be fired at any time. Proposal status could be changed by workflow and then officer decides that there is a manual change needed and it can fire proposal_status_changed_by_user. But thanks for the suggestion anyway.

mutambaraf (approved):
Nice refactoring

PR #72, opened 2022-10-11
Add PI column in Proposals table

RasmiaKulan:
## Description

The proposal system displays the Principal Investigator column in the Proposals table when accessed as **User Officer** and **Instrument Scientist**

## Motivation and Context

The proposal system did not have the option to view "Principal Investigator" within the **Show Columns** feature.

## How Has This Been Tested

- Manual tests
- E2E tests

## Fixes
Closes: UserOfficeProject/user-office-project-issue-tracker#642

## Changes
Modified the backend and frontend to fetch proposer details, and display it in a "Principal Investigator" column when the proposals are viewed as  **User Officer** and **Instrument Scientist**.

### Additional Changes
Modified the `localRoot` in the debug config to point at the backend.


apps/user-office-backend/src/resolvers/types/ProposalView.ts
```diff
@@ -87,3 +98,23 @@ export class ProposalView implements Partial<ProposalOrigin> {
   @Field(() => AllocationTimeUnits)
   public allocationTimeUnit: AllocationTimeUnits;
 }
+
+@Resolver((of) => ProposalView)
```
martin-trajanovski, line 102: @RasmiaKulan I have a small comment here. I think in the `ProposalView` we should avoid using the resolvers because the purpose of the view is to get combined tables data much quicker. Probably the right way to do it is to update the view containing the needed information and then just use that one directly. Here is an example: https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-backend/db_patches/0097_AddAllocationTimeUnitToProposalViewTable.sql. You should delete the view and recreate it with the new columns needed.

RasmiaKulan:
I will be on leave from 4th up until 18th November and wont be able to work on this issue as we are currently working on higher priority issues so its unlikely to be completed for a while. I have pushed all the changes I have made so far.
Branch name: user-office-project-issue-tracker-642-add_PI_column_to_proposals_table

PR #76, opened 2022-10-14
 add external logout and fix login with token

mutambaraf:
## Description

This will fix current login problem in develop and add external logout url for STFC

## Motivation and Context
The current develop branch don't login when using an external token and also the external logout url must be set for users to be able to successfully logout

## How Has This Been Tested

Manual testing

## Changes

.github/workflows/test-build.yml
apps/user-office-backend/src/config/stfc/configureSTFCEnvironment.ts
apps/user-office-frontend/src/components/user/ExternalAuth.tsx

## Depends on

https://github.com/isisbusapps/docker-orchestration/pull/315

### DIFF
--- a/apps/user-office-frontend/src/components/user/ExternalAuth.tsx
+++ b/apps/user-office-frontend/src/components/user/ExternalAuth.tsx
@@ -146,7 +146,8 @@ function ExternalAuth() {
     setView(<LoadingMessage />);
 
     const errorDescription = urlQueryParams.error_description;
-    const authorizationCode = urlQueryParams.sessionid ?? urlQueryParams.code;
+    const authorizationCode =
+      urlQueryParams.sessionid ?? urlQueryParams.code ?? urlQueryParams.token;
 
     if (errorDescription) {
       handleError(errorDescription);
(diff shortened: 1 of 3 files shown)


apps/user-office-frontend/src/components/user/ExternalAuth.tsx
```diff
@@ -146,7 +146,8 @@ function ExternalAuth() {
     setView(<LoadingMessage />);
 
     const errorDescription = urlQueryParams.error_description;
-    const authorizationCode = urlQueryParams.sessionid ?? urlQueryParams.code;
+    const authorizationCode =
+      urlQueryParams.sessionid ?? urlQueryParams.code ?? urlQueryParams.token;
```
jekabs-karklins, line 150: Do you need `sessionid` ? We only use the `code`

jekabs-karklins (approved):

RasmiaKulan (approved):

apps/user-office-frontend/src/components/user/ExternalAuth.tsx
simonfernandes: I don't think we do still use it but I'm not 100% sure. Removing this line and the login still works:

https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/user-office-frontend/src/components/App.tsx#L62

If we're in doubt then I think we could leave this condition as is for now and it shouldn't affect anything, right? As it'll just default to the others if null.

simonfernandes (approved):

PR #84, opened 2022-10-19
Update the logout function calling older in user context

mutambaraf:
## Description

When a user logout in stfc the system will still make calls to the backend creating invalid token logs.This is shown by the below screenshot.

## Motivation and Context

To get ready of unnecessary log failures. 

## How Has This Been Tested
Manually tested on local dev

## Fixes

Fix the following errors from being created.

Logout log errors

## Changes
The following file was updated.
apps/user-office-frontend/src/context/UserContextProvider.tsx


simonfernandes (approved):
I tested this locally and it fixes it for me too 🙂

PR #87, opened 2022-10-24
Add multi part selection component

keiranjprice101:
Key points.

Question configs are defined as key value pairs called `MultiPartSelectionPairs`. e.g.

```JavaScript
    {
        "a": [1, 2, 3]
        "b": [4,5,6]
    }
```

Will produce 2 dropdowns where the first dropdown has options `["a" , "b"]` and the second `[1, 2, 3]` or `[4, 5, 6]` depending on the selected option of dropdown one.

The answers are stored as type ` {partOneAnswer: string, partTwoAnswer: string}`

What needs to be done:

- [ ] Major UI improvement on form and template relation form. (Originally discussed moving to text field and comma seperated list)
- [ ] JSON postgres functions are not working for search
- [ ] Needs additional tests.
- [ ] An implimentation of `validate` should be created in the backened definition. (Not required to run)


TickleThePanda:
We don't have any use for this at the moment - we can look at reapplying something similar in the future if we need it.

PR #96, opened 2022-11-01
feat: Scheduler API access added to core if scheduler is enabled

martin-trajanovski:
## Description

There was a need to access scheduler by API access tokens. We added connection between the core and the scheduler where user office core can query all the mutations and queries from the scheduler and include them in the API access token generation if scheduler is enabled. Scheduler also implements all the logic needed to work with these tokens and validate them correctly, the same way as it is done in the core.

## Motivation and Context

The user office core allowed access with an API access token but scheduler wasn't included in this. Now we make scheduler to work with access token generated in the core with correct access rules.

## How Has This Been Tested

* Manual testing

## Fixes

* https://jira.esss.lu.se/browse/SWAP-2811
* https://jira.esss.lu.se/browse/SWAP-2875

## Changes

https://user-images.githubusercontent.com/6333063/199691326-f20e84aa-765d-43d7-bef7-afe29e48d9ba.mp4

## Depends on

https://github.com/UserOfficeProject/user-office-scheduler-backend/pull/191


apps/user-office-backend/src/queries/AdminQueries.ts
```diff
@@ -94,17 +95,60 @@ export default class AdminQueries {
         ];
 
       const proto = Object.getPrototypeOf(element);
-      const names = Object.getOwnPropertyNames(proto).filter(
-        (item) => item !== 'constructor'
-      );
+      if (!proto.constructor.name.startsWith('Admin')) {
```
jekabs-karklins, line 98: Would be nice to have a comment on why is are mutations starting with Admin excluded

apps/user-office-backend/src/queries/AdminQueries.ts
```diff
@@ -94,17 +95,60 @@ export default class AdminQueries {
...
+        const names = Object.getOwnPropertyNames(proto).filter(
+          (item) => item !== 'constructor'
+        );
 
-      const classNamesWithMethod = names.map(
-        (item) => `${proto.constructor.name}.${item}`
-      );
+        const classNamesWithMethod = names.map(
```
jekabs-karklins, line 103: Constructing the classNameWithMethod could be moved out to a function declaration to keep things DRY

apps/user-office-backend/src/queries/AdminQueries.ts
```diff
@@ -94,17 +95,60 @@ export default class AdminQueries {
...
+          (item) => `${proto.constructor.name}.${item}`
+        );
 
-      allMutationMethods.push(...classNamesWithMethod);
+        allMutationMethods.push(...classNamesWithMethod);
+      }
     });
 
-    return { queries: allQueryMethods, mutations: allMutationMethods };
+    // NOTE: If scheduler is disabled we get undefined as scheduler clientF
```
jekabs-karklins, line 111:
```suggestion
    // NOTE: If the scheduler is disabled we get undefined as scheduler client
```

jekabs-karklins (approved):

simonfernandes (approved):
Sorry for the delay on this Martin, I think this looks fine from our side 🙂

PR #115, opened 2022-12-13
Added navigation link

Junjiequan:
## Description

When specifying unit it would be nice to have a quick link so that you can quickly navigate to units page

## How Has This Been Tested

* manual testing

## Fixes

Screenshot 2022-12-13 at 14 42 48

## Changes

apps/user-office-frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx

### DIFF
--- a/apps/user-office-frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
+++ b/apps/user-office-frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
@@ -4,7 +4,6 @@ import FormControl from '@mui/material/FormControl';
 import Link from '@mui/material/Link';
 import MaterialTextField from '@mui/material/TextField';
 import makeStyles from '@mui/styles/makeStyles';
-// import clsx from 'clsx';
 import { Field } from 'formik';
 import { CheckboxWithLabel, TextField } from 'formik-mui';
 import React, { FC, useState } from 'react';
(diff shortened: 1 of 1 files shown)


apps/user-office-frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
```diff
@@ -81,24 +93,37 @@ export const QuestionNumberForm: FC<QuestionFormProps> = (props) => {
...
-                setSelectedUnits(newValue);
-                setFieldValue('config.units', newValue);
-              }}
-              value={selectedUnits ?? undefined}
-              data-cy="units"
-            />
+              <Link
+                href="/Units/"
+                target="_blank"
+                style={{ textAlign: 'right' }}
```
martin-trajanovski, line 118: Maybe you can extract this style in its own class instead of using styles. The same thing that you did with the icon. That way we keep the styling of the whole component in one place.

martin-trajanovski (changes requested):
I left a small comment but otherwise this looks great.

apps/user-office-frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
Junjiequan: Glad you mentioned it. I was a bit unsure whether should I use inline style, since its a small change. It definitely makes sense to put all stylings in one place.

martin-trajanovski (approved):

PR #125, opened 2023-01-04
feat: added create-new-workflow button in call modal

Junjiequan:
## Description

As a UO I want to be able to create stub Workflow so that I can quickly create new call

## Motivation and Context

Currently filling out the call after filling our first few fields turns out you can not continue because you have no Workflow created.

A solution is to add a new button "Create new workflow" which when created will create a new empty workflow and you can start using it but fill out it later.

## How Has This Been Tested

* manual test

## Fixes

https://user-images.githubusercontent.com/78078898/210516322-f28d6c41-7aea-4d78-915e-b38f7559bef8.mp4

## Changes

apps/user-office-frontend/src/components/call/CallGeneralInfo.tsx


ACLay:
I wonder if it'd be better to instead add a database patch that sets up a default workflow? We already have [one in place for adding a default template](https://github.com/UserOfficeProject/user-office-core/blob/93968ec68c5b4e01cd199180329c72c921db5bcf/apps/user-office-backend/db_patches/0026_Create_proposal_templates_table.sql) so I don't think it'd be unprecedented. 

I suggest this as it seems to me like there being no workflows would only be an issue the first time a call is created after the system is initialised, but if we present a "create workflow" button like this we introduce the risk of a user officer creating an empty workflow for any call, and if they don't go back to repopulate it or set it up properly, it would result in missed events and processing.

Junjiequan:
Our team discussed about your suggestion @ACLay. You made good point there, plus considering UI consistency this PR is closed.

PR #126, opened 2023-01-04
fix(tests): fix date to be in the future

jekabs-karklins:
## Description

Replace year 2023 with the year in the future

## Motivation and Context

The date must be in the future for tests to pass

## Fixes

https://jira.esss.lu.se/browse/SWAP-2955

## Depends on

N/A


Junjiequan (approved):
Awesome

PR #127, opened 2023-01-04
feat: call events must be retriggered when call dates are extended after being previously closed

mutambaraf:
## Description
CALL_ENDED or CALL_ENDED_INTERNAL events must be retriggered when either call end date or call end internal dates are updated.

## Motivation and Context

When the user officer opens a call after it has previously been closed and triggered the CALL_ENDED event. Proposals submitted after the call end date is updated will not have their workflow updated by the CALL_ENDED event because it would have been triggered already and will not be triggered when the new call ended date expires. The same issue also applies to CALL_ENDED_INTERNAL event.
Both CALL_ENDED or CALL_ENDED_INTERNAL event must be retriggered at the new dates after call dates updates.

## How Has This Been Tested

- manual tests

## Closes
https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/731

## Changes
Files:

- apps/user-office-backend/src/datasources/postgres/CallDataSource.ts

### DIFF
--- a/apps/user-office-backend/src/datasources/postgres/CallDataSource.ts
+++ b/apps/user-office-backend/src/datasources/postgres/CallDataSource.ts
@@ -179,12 +179,18 @@ export default class PostgresCallDataSource implements CallDataSource {
   async update(args: UpdateCallInput): Promise<Call> {
     const call = await database.transaction(async (trx) => {
       try {
+        const currentDate = new Date();
         /*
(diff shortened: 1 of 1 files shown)


Junjiequan (approved):
I'm still learning the business logic behind UO application, but the code looks good to me.

voduguwa (approved):

PR #128, opened 2023-01-04
chore: Upgrade apollo packages to prepare for gateway refactor

martin-trajanovski:
## Description

This PR is updating all the apollo packages to latest and moves the backend one step closer to upgrading gateway to version 2.

## Motivation and Context

We were using quite old apollo packages and it is good to be up to date and improve functionality around it.

## How Has This Been Tested

- unit tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2770

## Changes

Changes are mainly fixes after updating apollo packages


apps/user-office-backend/src/middlewares/rejectionLogger.ts
```diff
@@ -1,26 +1,26 @@
...
-    exception
-      ? logger.logException(error, exception, context)
-      : logger.logError(error, context ?? {});
-  }
-
-  return result;
-};
-
-export default rejectionLogger;
+// import { logger } from '@user-office-software/duo-logger';
```
jekabs-karklins, line 1: I think we should delete this

apps/user-office-backend/src/middlewares/rejectionSanitizer.ts
```diff
@@ -1,28 +1,26 @@
...
-  if (isRejection(rejection)) {
-    delete rejection.context;
-    delete rejection.exception;
-  }
-
-  return result;
-};
-
-export default rejectionSanitizer;
+// import { isRejection } from '../models/Rejection';
```
jekabs-karklins, line 1: I think we should delete this

apps/user-office-backend/src/resolvers/CustomScalars.ts
```diff
@@ -1,42 +1,42 @@
...
-    if (ast.kind === Kind.BOOLEAN) {
-      return ast.value;
-    }
-
-    return undefined;
-  },
-});
+import { GraphQLScalarType, Kind } from 'graphql';
+
+export type AnswerType = number | string | Date | boolean | number[] | unknown;
```
jekabs-karklins, line 3: Why is unknown needed here?

jekabs-karklins (approved):

mutambaraf (approved):
Besides removing commented code looks fine.

apps/user-office-backend/src/middlewares/graphql.ts
```diff
@@ -91,61 +154,31 @@ const apolloServer = async (app: Express) => {
...
-      const context: ResolverContext = {
-        ...baseContext,
-        user,
-        clients: {
-          scheduler: initGraphQLClient(req.headers.authorization),
-        },
-      };
-
-      return context;
+    formatError: (formattedError, error) => {
```
mutambaraf, line 157: Deep scan also flagged this declaration as unused.

apps/user-office-backend/src/resolvers/CustomScalars.ts
martin-trajanovski: It is needed after update of graphql. Most probably they changed the types and here is the result https://github.com/graphql/graphql-js/blob/main/src/type/definition.ts#L545. It is a bit disappointing because now the AnswerType is considered as unknown and the other types doesn't make much sense anymore.

PR #129, opened 2023-01-06
fix: use auth server instead of email authorizer

jekabs-karklins:

### DIFF
--- a/apps/user-office-frontend-e2e/cypress/support/user.ts
+++ b/apps/user-office-frontend-e2e/cypress/support/user.ts
@@ -1,7 +1,6 @@
 import {
   CreateUserMutation,
   CreateUserMutationVariables,
-  ExternalTokenLoginMutation,
   FeatureId,
   GetFeaturesQuery,
   Role,
@@ -117,6 +116,15 @@ const getUserIdFromIdOrCredentials = (
   }
 };
 
+const getCredentialsFromUserId = (testUserId: TestUserId) => {
+  const user = initialDBData.users[testUserId];
+
+  return {
+    email: user.email,
+    password: user.password,
+  };
+};
+
 const selectRole = async (token: string, selectedRoleId: number) => {
   const api = getE2EApi();
   const response = await api.selectRole({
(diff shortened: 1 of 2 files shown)


apps/user-office-frontend/src/components/user/UpdateUserInformation.tsx
```diff
@@ -164,40 +159,6 @@ export default function UpdateUserInformation(
     );
   };
 
-  const isUserOfficer = currentRole === UserRole.USER_OFFICER;
```
martin-trajanovski, line 167: So all this thing with the placeholder and email verification is not needed anymore. What about the invitation codes? Aren't those inviting user as a placeholder or it has another way of handling?

martin-trajanovski (approved):
Just a small comment otherwise it looks good.

apps/user-office-frontend-e2e/cypress/support/user.ts
```diff
@@ -1,234 +1,259 @@
...
+  const response = await api.selectRole({
+    token,
+    selectedRoleId,
+  });
+
+  return response.selectRole;
+};
+
+const getOauthExternalToken = async (testUserId: TestUserId) => {
+  const DEV_AUTH_SERVER_URL = 'http://localhost:5000';
```
TCMeldrum, line 129: Could you make this environment variable? Since we have our local setup in the WSL localhost won't point to the auth server so will need to give our own URL to get the test working on local

TCMeldrum (changes requested):
Just a small change to get this working on our local setup. Other than that looks good will run it by the team tomorrow quickly to confirm we are happy with it.

apps/user-office-frontend-e2e/cypress/support/user.ts
martin-trajanovski: Will do it @TCMeldrum. Thanks for the good point.

martin-trajanovski: @TCMeldrum I added this `DEV_AUTH_SERVER_URL` in the cypress.json file as env variable. Let me know if this makes sense.

TCMeldrum: Looks good thanks!

TCMeldrum (commented):
Happy with this do you mind holding off merging until I sort some local config for us?

martin-trajanovski:
> Happy with this do you mind holding off merging until I sort some local config for us?

No problem at all.

TCMeldrum (approved):
Cool local config sorted sorry for the delay

PR #132, opened 2023-01-10
fix: make settings update more resilient  

jekabs-karklins:
## Description

This PR aims to fix the issue where the server will throw an exception if trying to update a setting that does not exist. The more resilient approach is to create the setting in case it does not exist.

## Motivation and Context

This PR makes the backend more resilient

## Fixes

https://jira.esss.lu.se/browse/SWAP-2954

### DIFF
--- a/apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
+++ b/apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
@@ -378,6 +378,14 @@ export default class PostgresAdminDataSource implements AdminDataSource {
       );
   }
 
+  async createSetting(setting: Settings): Promise<Settings> {
+    return database
+      .insert(setting)
+      .into('settings')
+      .returning('*')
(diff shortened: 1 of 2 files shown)


mutambaraf (approved):

apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
```diff
@@ -494,6 +508,15 @@ export default class PostgresAdminDataSource implements AdminDataSource {
   ): Promise<Settings> {
     const { settingsId, description, settingsValue } = updatedSettingsInput;
 
+    const setting = await this.getSetting(settingsId);
```
martin-trajanovski, line 511: Cant we just use upsert instead of update without getting and creating?

martin-trajanovski: In the end your solution might be good enough. Forget about previous comment.

martin-trajanovski (approved):

PR #133, opened 2023-01-10
refactor: Improve graphql error handling and logging by extending GraphQLError

martin-trajanovski:
## Description

**Proper GraphQL error handling and logging.**

- Improve `rejection` by extending the `GraphQLError`
- Remove unnecessary response wrappers and types.
- Improve error logging and handling
- Frontend now doesn't need to handle errors separately because they are handled by default at one place and all end up in the `catch block` instead of then with `rejection` inside instead of successful response.

## Motivation and Context

1. If we use rejection as we did previously `graphql-request` treats that as a success and doesn't end up in a catch statement if there is an error. It just continues with then with empty data inside and no message at all.
2. Another thing that this PR improves is the overview and error reporting in a sense that we log and send all the errors to Apollo studio now which is quite important.

## How Has This Been Tested

- manual tests
- e2e tests
- unit tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2981

## Changes

Main change is in the `Rejection` class that now extends `GraphQLError` and implements error handling in the proper `GraphQL` way. All the other changes are mostly removing unnecessary (old way) rejection connected code.


apps/user-office-frontend/src/components/template/QuestionsPage.tsx
```diff
@@ -1,116 +1,116 @@
...
+      });
+    }
+  };
+
+  const deleteQuestion = async (questionId: string | number) => {
+    await api({
+      toastSuccessMessage: 'Question deleted',
+    }).deleteQuestion({ questionId: questionId as string });
+
+    return true;
```
ACLay, line 82: Could this cause problems for any calling code? If there's an error I think that code would need to catch and handle the error instead of just receiving a return value of false.

ACLay (changes requested):
Only did a look over with whitespace changes hidden, but in general it looks like a nice simplification removing all the WrapResponse calls and classes.

martin-trajanovski:
> Only did a look over with whitespace changes hidden, but in general it looks like a nice simplification removing all the WrapResponse calls and classes.

Thanks for your great review @ACLay! I went trough the comments and pushed some fixes where it was needed. You can double check this once again.

bolmsten (approved):

RasmiaKulan (approved):

apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
```diff
@@ -91,18 +88,13 @@ export default class PostgresAdminDataSource implements AdminDataSource {
       .returning('*');
 
     if (!institutionRecord) {
-      throw new Error('Could not create call');
+      return null;
```
mutambaraf, line 91: We now return now here unless we no longer need throwing an error looks fine

apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
```diff
@@ -67,19 +69,14 @@ export default class PostgresAdminDataSource implements AdminDataSource {
       .returning('*');
 
     if (!institutionRecord) {
-      throw new Error(`Could not update institution with id:${institution.id}`);
+      return null;
```
mutambaraf, line 72: We also do it here.

mutambaraf (approved):
Looks fine with me .

apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
martin-trajanovski: @mutambaraf thanks for noticing this one. I think the change here is that I moved the error handling in the mutation/logic layer instead of database layer. It is here: https://github.com/UserOfficeProject/user-office-core/pull/133/files/71e7ca9c5da4db73ed7aa0dc9a857d8d6b31a534?diff=split&w=1#diff-61a948f290a5811fde75287846cdda53de34db50b47afc1e29a001ed642c85e9R107

Because PR is quite big with lot of changes you might need to wait a bit until this link opens at the right line.

apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
martin-trajanovski: It is the same for this one I just transferred the logic for error handling in the controller so to say.

jekabs-karklins (approved):

Junjiequan (approved):

PR #160, opened 2023-01-18
refactor: Remove cypress-file-upload package because cy.selectFile was introduced and do some cleanup of duplicate tests

martin-trajanovski:
## Description

This PR aims to improve tests and remove duplicates.

## Motivation and Context

1. In the template tests we had duplicate tests that were left after some splitting in multiple files.
2. Remove `cypress-file-upload` package cause it is not needed after cypress introduced `cy.selectFile` command on their own.

### DIFF
--- a/apps/user-office-frontend/src/components/questionary/questionaryComponents/FileUpload/QuestionaryComponentFileUpload.tsx
+++ b/apps/user-office-frontend/src/components/questionary/questionaryComponents/FileUpload/QuestionaryComponentFileUpload.tsx
@@ -49,7 +49,6 @@ export function QuestionaryComponentFileUpload(
       </FormLabel>
       <FileUploadComponent
         id={answer.question.id}
-        question={answer.question.question}
         maxFiles={config.max_files}
         pdfPageLimit={config.pdf_page_limit}
         fileType={config.file_type ? config.file_type.join(',') : ''}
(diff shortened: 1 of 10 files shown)


apps/user-office-frontend/src/components/questionary/questionaryComponents/FileUpload/QuestionaryComponentFileUpload.tsx
```diff
@@ -1,70 +1,71 @@
...
+        {config.small_label && (
+          <>
+            <br />
+            <small>{config.small_label}</small>
+          </>
+        )}
+      </FormLabel>
+      <FileUploadComponent
+        id={answer.question.id}
+        question={answer.question.question}
```
jekabs-karklins, line 52: I think it is better to use answer.question.id here, because id is a unique identifier

jekabs-karklins (approved):

apps/user-office-frontend/src/components/questionary/questionaryComponents/FileUpload/QuestionaryComponentFileUpload.tsx
martin-trajanovski: I am reverting these changes and will just select the element by question name itself. Something like:
`cy.contains(fileQuestion).parent().find('input[type="file"]')`. Thanks for the good comment.

PR #171, opened 2023-02-15
fix: add instrument to RabbitMQ message

jekabs-karklins:
## Description

This PR adds instrument ID and shortcode to the Rabbit MQ message

## Motivation and Context

Dependant systems require this information to be there

## How Has This Been Tested

Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-3050

## Changes
Message brokenr

## Depends on

NA

### DIFF
--- a/apps/user-office-backend/src/eventHandlers/messageBroker.ts
+++ b/apps/user-office-backend/src/eventHandlers/messageBroker.ts
@@ -15,6 +15,7 @@ import { ApplicationEvent } from '../events/applicationEvents';
 import { Event } from '../events/event.enum';
 import { EventHandler } from '../events/eventBus';
 import { AllocationTimeUnits } from '../models/Call';
+import { Instrument } from '../models/Instrument';
 import { Proposal, ProposalEndStatus } from '../models/Proposal';
 import { ScheduledEventCore } from '../models/ScheduledEventCore';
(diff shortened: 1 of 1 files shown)


nitrosx (approved):

apps/user-office-backend/src/eventHandlers/messageBroker.ts
```diff
@@ -90,9 +95,21 @@ const getProposalMessageData = async (proposal: Proposal) => {
...
+  const maybeInstrument = await instrumentDataSource.getInstrumentByProposalPk(
+    proposal.primaryKey
+  );
+
+  const instrument = maybeInstrument
+    ? {
+        id: maybeInstrument.id,
+        shortCode: maybeInstrument.shortCode,
+      }
+    : undefined;
```
mutambaraf, line 107: If we don't have an instrument do we need it to be `undefined` or we can skip it from the message Data. But eventually It will all be about its usage on the receiving consumer.

jekabs-karklins: I tried to stick to the convention in this file. I could see that `newStatus` if not set would be undefined.

mutambaraf: Okay can see the point.

mutambaraf (approved):

PR #182, opened 2023-02-20
fix: fix deployment scripts

jekabs-karklins:
## Description

This PR updates the remaining scripts used also for deployment so that the right node version is used. This makes sure that the merged code into development gets deployed on the test servers.


mutambaraf (approved):
Looks fine to me.

PR #187, opened 2023-02-22
feat: Idle timeout warning

TCMeldrum:
https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/791

## Description

Adds a timeout warning for when the user is near to being logged out. It creates a new context `IdleContext` that refreshes the timer on every call to the backend the context is behind a new feature flag `STFC_IDLE_TIMER`. If this flag is not set it returns a context that will only return the default values. When the warning message is dismissed it checks if the external token is valid and if not redirects the user to the sign-in page. 

Our requirement that activity is refreshed by calls to our backend is different from most idle timers as they define active to be DOM events. For this reason, we are not using any library like [react-idle-timer](https://github.com/SupremeTechnopriest/react-idle-timer)


apps/user-office-backend/db_patches/0130_AddTimedLogoutWarningFeature.sql
```diff
@@ -0,0 +1,12 @@
+DO
+$$
+BEGIN
+    IF register_patch('AddTimedLogoutWarningfeature.sql', 'Thomas Cottee Meldrum', 'Create STFC idle warning feature', '2023-02-14') THEN
+
+    INSERT INTO features(feature_id, description) VALUES ('STFC_IDLE_TIMER', 'STFC idle warning popup');
+    INSERT INTO settings(settings_id, description) VALUES ('IDLE_TIMEOUT', 'Timour for Idle timer');
```
RasmiaKulan, line 7: I may be wrong but is it meant to be `Timour` there?

apps/user-office-frontend/src/components/timeout/TimeOutPrompt.tsx
```diff
@@ -0,0 +1,57 @@
+import {
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+  // useTheme,
```
RasmiaKulan, line 8: Is it alright to remove `useTheme` as its not being used?

apps/user-office-backend/db_patches/0130_AddTimedLogoutWarningFeature.sql
TCMeldrum: Yeah typo thanks for spotting that

ACLay (commented):
I'm assuming milliseconds from the number of zeros, but could we add a note somewhere of the units for the timeout setting?

ACLay (approved):

RasmiaKulan (approved):

apps/user-office-frontend/src/components/timeout/TimeOutPrompt.tsx
```diff
@@ -0,0 +1,56 @@
...
+    display: 'grid',
+    placeItems: 'center',
+  },
+}));
+
+function IdleTimeoutPrompt(props: IdleTimeoutPromptProps): JSX.Element {
+  const classes = useStyles();
+
+  return (
+    <>
```
martin-trajanovski, line 32: I think you don't need this React fragment here or empty tag because the Dialog element wraps everything.

martin-trajanovski (approved):
Just a small comment but it looks good to me otherwise.

apps/user-office-frontend/src/components/timeout/TimeOutPrompt.tsx
TCMeldrum: Yep missed that thanks!

PR #188, opened 2023-02-23
Fix: reference number gap because the SUBMITTED button on result page is clicked multiple times after a proposal submission #740

bashanlam:
Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/740

## Description
After a proposal submission, the SUBMITTED button is still enable on the result page. If the button is clicked, the proposal is re-submitted with the next reference number in the sequence.

## Motivation and Context
To prevent a proposal re-submission, both backend and frontend are fixed. 
1. Backend rejects the proposal re-submission after a proposal submission. 
2. Frontend disable the SUBMITTED button on the result page after a proposal submission. 

## How Has This Been Tested
Manually tested. A backend and frontend re-submission checking are added. Each checking is tested individually.
1. Backend fix is tested first because frontend fix will disable the SUBMITTED button after a proposal submission that forbid the backend fix testing. 
2. Frontend fix is deployed and tested after backend fix is tested.

## Changes
1. Amend backend file ProposalMutations.ts to reject re-submission.
2. Amend frontend file ProposalSummary.tsx file to disable the SUBMITTED button after a proposal submission.

- [x ] I have added tests to cover my changes.

### DIFF
--- a/apps/user-office-backend/src/mutations/ProposalMutations.ts
+++ b/apps/user-office-backend/src/mutations/ProposalMutations.ts
@@ -549,18 +549,12 @@ export default class ProposalMutations {
         sourceProposal.questionaryId
       );
 
-      // if user clones the proposal then it is his/her,
-      // but if userofficer, then it will belong to original proposer
-      const proposerId = this.userAuth.isUserOfficer(agent)
-        ? sourceProposal.proposerId
-        : agent!.id;
-
       // TODO: Check if we need to also clone the technical review when cloning the proposal.
       clonedProposal = await this.proposalDataSource.update({
         primaryKey: clonedProposal.primaryKey,
         title: `Copy of ${clonedProposal.title}`,
         abstract: clonedProposal.abstract,
-        proposerId: proposerId,
+        proposerId: sourceProposal.proposerId,
         statusId: 1,
         created: new Date(),
         updated: new Date(),
--- a/apps/user-office-backend/src/config/ess/configureESSEnvironment.ts
+++ b/apps/user-office-backend/src/config/ess/configureESSEnvironment.ts
@@ -5,6 +5,7 @@ import { FeatureId } from '../../models/Feature';
 import { SettingsId } from '../../models/Settings';
 import { setTimezone, setDateTimeFormats } from '../setTimezoneAndFormat';
 import { Tokens } from '../Tokens';
+import { updateOIDCSettings } from '../updateOIDCSettings';
 
 async function setEssColourTheme() {
   const db = container.resolve<AdminDataSource>(Tokens.AdminDataSource);
@@ -99,4 +100,5 @@ export async function configureESSDevelopmentEnvironment() {
   await enableDefaultEssFeatures();
   await setTimezone();
   await setDateTimeFormats();
+  await updateOIDCSettings();
 }
--- a/apps/user-office-backend/src/config/updateOIDCSettings.ts
+++ b/apps/user-office-backend/src/config/updateOIDCSettings.ts
@@ -1,0 +1,26 @@
+import { OpenIdClient } from '@user-office-software/openid';
+import { container } from 'tsyringe';
+
+import { AdminDataSource } from '../datasources/AdminDataSource';
+import { SettingsId } from '../models/Settings';
+import { Tokens } from './Tokens';
+
+export async function updateOIDCSettings() {
+  if (OpenIdClient.hasConfig() === false) {
+    return;
+  }
+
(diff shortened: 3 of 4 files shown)


apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
```diff
@@ -50,6 +50,9 @@ function ProposalReview({ confirm }: ProposalSummaryProps) {
 
   const proposal = state.proposal;
 
+  // Declare state of submission. This is first submission. (3-Feb-23)
```
mutambaraf, line 53: Looks fine but l don't think we need this comment the code is self explanatory.

apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
```diff
@@ -172,6 +175,7 @@ function ProposalReview({ confirm }: ProposalSummaryProps) {
                     itemWithQuestionary: submitProposal,
                   });
                 } finally {
+                  setFirstSubmit(false); //Set state after first submission. (3-Feb-23)
```
mutambaraf, line 178: This can be removed as well.

apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
```diff
@@ -181,7 +185,7 @@ function ProposalReview({ confirm }: ProposalSummaryProps) {
               }
             )();
           }}
-          disabled={submitDisabled}
+          disabled={submitDisabled || !firstSubmit} //Disable button after first submission. (3-Feb-23)
```
mutambaraf, line 188: This comment can be removed as well.

apps/user-office-backend/src/mutations/ProposalMutations.ts
```diff
@@ -226,6 +226,12 @@ export default class ProposalMutations {
         proposalPk,
       });
     }
+    if (proposal.submitted) {
+      return rejection('Can not submit the proposal again', {
```
mutambaraf, line 230: Text can be improved if possible. "Proposal already submitted "

mutambaraf (changes requested):

apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
bashanlam: Updated. Please check, thanks.

apps/user-office-backend/src/mutations/ProposalMutations.ts
bashanlam: Updated. Please check, thanks.

apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
bashanlam: Updated. Please check, thanks.

apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
bashanlam: Updated. Please check, thanks.

mutambaraf (approved):

mutambaraf (approved):
Don't see any changes so will approve.

apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
```diff
@@ -181,7 +184,7 @@ function ProposalReview({ confirm }: ProposalSummaryProps) {
               }
             )();
           }}
-          disabled={submitDisabled}
+          disabled={submitDisabled || !firstSubmit}
```
martin-trajanovski, line 187: Can't we just use the `submitDisabled` variable instead of introducing new one called `firstSubmit`? For example add `setSubmitDisabled` in the `useState` declaration of it and set it to true at the same place where you `setFirstSubmit`

martin-trajanovski (changes requested):

apps/user-office-frontend/src/components/proposal/ProposalSummary.tsx
martin-trajanovski: Also would be good maybe to have some e2e test that covers this for the future.

martin-trajanovski: @bashanlam I think you can completly remove this `|| !firstSubmit` part from the disabled condition and leave just `submitDisabled` as it was.

martin-trajanovski (approved):
It looks good to me now

PR #192, opened 2023-02-24
feat: added jsonpath to dynamic multi choice questionary component

Junjiequan:
## Description

Included jsonpath library to the dynamic multiple choice questionary component.
Jsonpath can be used to specify and get desired data from response of the external api.
If the external api service is down, the component will render empty list. 
In the future, there will be new solution to address api service down issues and also a popup instruction modal to show how to use jsonpath correctly.

## Motivation and Context

As a user office I want to specify jsonPath for Dynamic Multiple Question to transform the response

## How Has This Been Tested

- Manual
- E2E

## Changes

Here is a simple use case example.
Let's say an external api response with following values:

```
[
  {
    "item": "EXTERNAL_API_RESPONSE_1"
  },
  {
    "item": "EXTERNAL_API_RESPONSE_2"
  },
  {
    "item": "EXTERNAL_API_RESPONSE_3"
  },
  {
    "item": "EXTERNAL_API_RESPONSE_4"
  }
]
```
You can type `$.*.item` into jsonpath and you will get:
```
[
  "EXTERNAL_API_RESPONSE_1",
  "EXTERNAL_API_RESPONSE_2",
  "EXTERNAL_API_RESPONSE_3",
  "EXTERNAL_API_RESPONSE_4"
]
```


bolmsten (changes requested):
We will not merge this until the fetch is done in the backend as we do not want to add packages in frontend

apps/user-office-backend/src/models/questionTypes/QuestionRegistry.ts
```diff
@@ -137,3 +143,20 @@ export function getDefaultAnswerValue(
...
+  const definition = getQuestionDefinition(record.data_type as DataType);
+
+  if (!definition.externalApiCall) {
+    logger.logError('Tried to make non-existing definition callback', {
+      dataType: record.data_type as DataType,
+    });
+    throw new GraphQLError('Tried to execute non-existing definition callback');
+  }
+
+  const options = definition.externalApiCall(record as DynamicRecord);
```
martin-trajanovski, line 157: I see this `externalApiCall` is an async function. Shouldn't we `await` it or I misunderstand something here.

apps/user-office-backend/src/models/questionTypes/QuestionRegistry.ts
```diff
@@ -137,3 +143,20 @@ export function getDefaultAnswerValue(
...
+
+  return options;
+}
+
+// export function
```
martin-trajanovski, line 162: This comment might need to be removed

apps/user-office-backend/src/models/questionTypes/DynamicMultipleChoice.ts
```diff
@@ -50,4 +67,68 @@ export const dynamicMultipleChoiceDefinition: Question = {
...
+        recordWithApiCall = {
+          ...record,
+          config: {
+            ...record.config,
+            options: resp.data,
+          },
+        };
+      } else {
+        try {
+          //TODO: if jsonPathFilteredData is empty, then we know
```
martin-trajanovski, line 92: Is there still something todo here or this is just a NOTE comment.

martin-trajanovski (changes requested):

apps/user-office-backend/src/models/questionTypes/DynamicMultipleChoice.ts
Junjiequan: I was thinking we might possibly want to add error text to frontend to notify users that jsonpath syntax is possibly wrong. But I think This TODO can be removed.

mutambaraf:
We can also fix the deep scan issues.

apps/user-office-frontend-e2e/cypress/integration/templates_basic.ts
```diff
@@ -1378,40 +1413,98 @@ context('Template tests', () => {
...
-      cy.visit('/');
-
-      cy.contains('New Proposal').click();
-      cy.get('[data-cy=call-list]').find('li:first-child').click();
+      cy.get('[data-cy=dropdown-ul]').children().should('not.contain.value');
+    });
 
-      cy.finishedLoading();
+    it('Should be able to use JSONPATH library to extract specific data from API response', () => {
+      cy.intercept(
```
mutambaraf, line 1464: Since we now get values from the backend is it possible to get the options from there instead.

apps/user-office-backend/src/datasources/postgres/records.ts
```diff
@@ -791,11 +795,16 @@ export const createFileMetadata = (record: FileRecord) => {
...
 };
 
-export const createQuestionTemplateRelationObject = (
+export const createQuestionTemplateRelationObject = async (
   record: QuestionRecord &
     QuestionTemplateRelRecord & { dependency_natural_key: string },
   dependencies: FieldDependency[]
 ) => {
+  let recordWithApiCall;
+  if (record.config.hasOwnProperty('url')) {
```
mutambaraf, line 804: I don't know if we have covered a situation where  'url' is not available .

apps/user-office-backend/src/models/questionTypes/QuestionRegistry.ts
```diff
@@ -137,3 +143,18 @@ export function getDefaultAnswerValue(
...
+  const definition = getQuestionDefinition(record.data_type as DataType);
+
+  if (!definition.externalApiCall) {
+    logger.logError('Tried to make non-existing definition callback', {
+      dataType: record.data_type as DataType,
+    });
+    throw new GraphQLError('Tried to execute non-existing definition callback');
+  }
+
+  const options = await definition.externalApiCall(record as DynamicRecord);
```
mutambaraf, line 157: Do we need this declaration if we are not it  beyond this point unless it is for other reasons.

mutambaraf (commented):
Looks good to me except those few issues.

apps/user-office-backend/src/models/questionTypes/QuestionRegistry.ts
Junjiequan: This declaration can be removed, I included it just for better personal understanding of what to expect from the value returned

mutambaraf: Okay thanks.

apps/user-office-backend/src/datasources/postgres/records.ts
Junjiequan: Yes, I included potential errors that I could think of. 
First check if it is a valid url address with regex on frontend
Then we check if api call fails or jsonpath is wrong syntax on backend. And return empty string on either of errors

apps/user-office-frontend-e2e/cypress/integration/templates_basic.ts
Junjiequan: I created mock server to fake the behavior in the new commit.

martin-trajanovski (approved):

martin-trajanovski (approved):
Seems good to me but it will be nice to see a demo of this.

mutambaraf (approved):

martin-trajanovski:
> We will not merge this until the fetch is done in the backend as we do not want to add packages in frontend

@bolmsten if this is fixed you can approve and @Junjiequan can merge this PR I guess.

bolmsten (approved):

PR #200, opened 2023-02-28
fix: Settings and Features could be added before db initialized

TCMeldrum:
UserOfficeProject/user-office-project-issue-tracker/issues/835

## Description

Added check to see if database patches have been done before updating the db on startup. 


apps/user-office-backend/src/datasources/postgres/AdminDataSource.ts
```diff
@@ -44,6 +44,8 @@ const seedsPath = path.join(dbPatchesFolderPath, 'db_seeds');
 
 @injectable()
 export default class PostgresAdminDataSource implements AdminDataSource {
+  protected autoUpgradedDBReady = true;
```
ACLay, line 47: It looks like we've got two initialisers for this property, both here and in the constructor. Does it need to be set twice? Could we just set it false here and remove the other?

TCMeldrum: I had this here originally as if it was using `PostgresAdminDataSource` not `PostgresAdminDataSourceWithAutoUpgrade` this would never be set to `true` but I think it doesn't matter. I'll remove it.

ACLay (approved):

RasmiaKulan (approved):

bolmsten (approved):

PR #205, opened 2023-03-03
Fix: 834 Cloned proposals have no validation

TCMeldrum:
Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/834
## Description

Added a flag to stop cloning a proposal to mark a topic as complete so now a user must review and save all the steps in the front. This was done as if template constraints change the cloned data and could violate them but still be submitted. I have done this through a flag rather than as default behavior as the clone method is used in various places where I am unsure if this validation would be needed. There was also a bug were internal users could submit uncompleted proposals which we (STFC) should probably hotfix right away.

A previous version that I have saved locally wouldn't clone any answers that violated any of the questionaries constraints but I felt keeping the old answers made more sense for when the user is reviewing the proposal. If others disagree I can add it back. 


simonfernandes (approved):
Looks like a really neat solution

martin-trajanovski (approved):
Looks good to me as well

PR #211, opened 2023-03-14
feat: use env variables for queue names and exchanges and bind the queue to the exchange automatically

martin-trajanovski:
## Description

This PR adds the functionality to bind a queue to exchange automatically instead of doing it manually in the rabbitmq

## Motivation and Context

Previously you needed to manually bind the right queue to the right exchange.

## How Has This Been Tested

- manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-3131

## Changes

Changes are mostly related to automating the process of binding queues to exchange and awaiting all async functions.

## Depends on

- Message broker PR: https://github.com/UserOfficeProject/user-office-lib/pull/167
- Scheduler PR: https://github.com/UserOfficeProject/user-office-scheduler-backend/pull/235


apps/user-office-backend/src/eventHandlers/messageBroker.ts
```diff
@@ -45,17 +48,17 @@ type ProposalMessageData = {
...
 };
 
-const createRabbitMQMessageBroker = () => {
+const createRabbitMQMessageBroker = async () => {
   const messageBroker = new RabbitMQMessageBroker();
-  messageBroker.setup({
+  await messageBroker.setup({
     hostname: process.env.RABBITMQ_HOSTNAME,
     username: process.env.RABBITMQ_USERNAME,
     password: process.env.RABBITMQ_PASSWORD,
```
Junjiequan, line 64: Shouldn't we use try catch for this block of code? I assume it tries to connect to something, but what if it fails? Does it throw error or?

apps/user-office-backend/src/eventHandlers/messageBroker.ts
```diff
@@ -287,23 +282,49 @@ export function createPostToRabbitMQHandler() {
...
 }
 
-export function createListenToRabbitMQHandler() {
-  const rabbitMQ = getRabbitMQMessageBroker();
+export async function createListenToRabbitMQHandler() {
+  const EVENT_SCHEDULING_QUEUE_NAME = process.env
+    .EVENT_SCHEDULING_QUEUE_NAME as Queue;
+  const SCHEDULER_EXCHANGE_NAME = process.env.SCHEDULER_EXCHANGE_NAME;
+
+  // TODO: Maybe we need to check if scheduler is enabled then throw these errors. Check this with STFC.
```
ACLay, line 313: Yeah, could this only be done if the scheduler feature flag is set?

ACLay (changes requested):

apps/user-office-backend/src/eventHandlers/messageBroker.ts
martin-trajanovski: I will try to do that

martin-trajanovski: @ACLay do you want to completely disable the `createListenToRabbitMQHandler` and `createPostToRabbitMQHandler` if the scheduler is disabled or just avoid throwing the error messages? Even if we avoid throwing the errors it will still try to listen on a queue that has empty value because that `env` variable is not set.

Junjiequan (approved):

apps/user-office-backend/src/eventHandlers/messageBroker.ts
ACLay: @TCMeldrum did a bit more checking on this. Looks like we can just avoid using this listener altogether by updating the STFC [dependencyConfig file](https://github.com/UserOfficeProject/user-office-core/blob/235c0b687b3bbbee6044708bcd01d191d980f28d/apps/user-office-backend/src/config/dependencyConfigSTFC.ts#L79) so it uses a dummy listener like in the [test depencyConfig](https://github.com/UserOfficeProject/user-office-core/blob/235c0b687b3bbbee6044708bcd01d191d980f28d/apps/user-office-backend/src/config/dependencyConfigTest.ts#L75). If that is changed it should prevent this causing errors on our side, and this TODO comment can be removed.

martin-trajanovski: @ACLay thanks for letting me know. I will try to update that and you can review if it is good enough.

martin-trajanovski: @ACLay just to be clear that I will only skip the listener but not the `createPostToRabbitMQHandler`.

ACLay (approved):

PR #215, opened 2023-03-22
Remove pushes to dockerhub

TCMeldrum:
Ref https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/842
## Description

Remove pushing to docker hub from the build push actions 

Blocked by https://github.com/UserOfficeProject/mock-uows/pull/5

Workflow tested here: the tagging and pushing passed: https://github.com/UserOfficeProject/user-office-core/actions/runs/4491542655


.github/workflows/test-build.yml
```diff
@@ -39,7 +39,6 @@ jobs:
           }
 
           FACTORY_TAGS=$(check_feature ${{ github.head_ref }} \
-            https://registry.hub.docker.com/v2/repositories/dmsc/duo-factory/tags \
```
martin-trajanovski, line 42: So removing this will end up having only one tag inside FACTORY_TAGS variable. And down there where we check if FACTORY_TAGS == "2" this can not be true at any point of time. Maybe we should remove that or if we still want to support checking the registry for image tag with the branch name we should try to find some equivalent solution for ghcr.io

martin-trajanovski (approved):
There is one comment that might need resolution but otherwise looks good.

.github/workflows/test-build.yml
TCMeldrum: Yeah, could not find an easy way for checking the tags on the GitHub registry looks like checking the manifests hasn't worked. Any ideas on how to check the registry?

TCMeldrum: Found a way if you want to have a look

PR #221, opened 2023-03-30
814 add localisation for different terminology

TCMeldrum:
Closes: UserOfficeProject/user-office-project-issue-tracker#814

## Description

Added way for STFC to change the instrument keyword in the system. This is done in 2 ways:

- Added localisation to the project through the i18next package, this can be expanded in the future if there are any other differences in terminology. This involved trying to find every instance of the word instrument in the software and changing it. 
- Added a way to edit the role names on startup similar to settings 

I have (hopefully) found every instance of "Instrument" but I could be wrong.  Further, I am not sure if a screen reader would require more changes of labels/aria-labels 
## Motivation and Context

One of our facilities took issue with the "Instrument" being used to describe their facilities so we needed a way to change this for us but not the ESS.

## Depends on

~~https://git.devs.facilities.rl.ac.uk/isisbusapps/BisAppSettings/pulls/146~~ https://git.devs.facilities.rl.ac.uk/isisbusapps/BisAppSettings/pulls/1
https://github.com/isisbusapps/docker-orchestration/pull/418


TCMeldrum:
> The localisation is applied to all front-end after this change?

The localisation works by looking in a json file and picking out the localised wording using the `t()` function that is imported in `index.ts`. So it can be called where ever in the frontend using the `useTranslation` hook it will not be automatically applied only in places I have identified.

apps/user-office-frontend/src/components/call/AssignedInstrumentsTable.tsx
```diff
@@ -126,7 +128,8 @@ const AssignedInstrumentsTable: React.FC<AssignedInstrumentsTableProps> = ({
 
   const removeAssignedInstrument = async (instrumentId: number) => {
     await api({
-      toastSuccessMessage: 'Assigned instrument removed successfully!',
+      toastSuccessMessage:
+        'Assigned ' + t('instrument.single') + '  removed successfully!',
```
simonfernandes, line 132: Double space here

apps/user-office-frontend/src/components/instrument/CreateUpdateInstrument.tsx
```diff
@@ -74,7 +77,8 @@ const CreateUpdateInstrument: React.FC<CreateUpdateInstrumentProps> = ({
         } else {
           try {
             const { createInstrument } = await api({
-              toastSuccessMessage: 'Instrument created successfully!',
+              toastSuccessMessage:
+                t('instrument.single') + 'created successfully!',
```
simonfernandes, line 81: Missing space

simonfernandes (commented):
Looks good mainly, there's a few spacing issues that I've commented on.

Is there an easy way to pass in the capitalisation of terms? E.g. I think Ideally a message like "Scientist assigned to Instrument" should have instrument in lower case.

TCMeldrum:
> Is there an easy way to pass in the capitalisation of terms? E.g. I think Ideally a message like "Scientist assigned to Instrument" should have instrument in lower case.
We either add a capitalized version in `translation.json` or call `toLowerCase()` where nessasary. I guess adding it to `translation.json` would be more readable?

simonfernandes:
It looks like this version of i18next supports [formatting options](https://www.i18next.com/translation-function/formatting), so that seems like the nicer option to me. Looks like it might then go in `i18n.ts` rather than `translation.json`? If it is tricky then I don't think the way it currently is is too big of a deal.

apps/user-office-frontend/src/components/App.tsx
```diff
@@ -107,36 +107,53 @@ class App extends React.Component {
...
-                      </Router>
-                    </IdleContextPicker>
-                  </DownloadContextProvider>
-                </UserContextProvider>
-              </FeatureContextProvider>
-            </Theme>
-          </SettingsContextProvider>
-        </SnackbarProvider>
-      </StyledEngineProvider>
+      <Suspense
```
martin-trajanovski, line 110: I am just wondering if this is really needed because I think we already have some loader included in the main page or maybe I am wrong. Or at least we have some loading text when we load settings and features. If this works better then we can remove that one over there.

apps/user-office-frontend/src/i18n.ts
```diff
@@ -0,0 +1,27 @@
+import i18n from 'i18next';
+import Backend from 'i18next-http-backend';
+import { initReactI18next } from 'react-i18next';
+
+i18n
+  .use(Backend)
+  .use(initReactI18next)
+  .init({
+    debug: true,
```
martin-trajanovski, line 9: I am also wondering what this debug: true really does and if it should be hardcoded like this.

martin-trajanovski (commented):
It looks good in general but I have few small comments.

apps/user-office-frontend/src/components/App.tsx
TCMeldrum: I couldn't get i18n working without this, I get errors around a suspended while rendering, with no fall back UI specified. Looks like the loading text around setting and features is around waiting for API calls to the backend to be resolved so not a loader as such. Not sure if we have a loader else where can't seem to find one and I guess it would be too low down the react tree for i18n to use.

apps/user-office-frontend/src/i18n.ts
TCMeldrum: Missed that from test thanks!

apps/user-office-frontend/src/components/App.tsx
martin-trajanovski: OK then I understand we can keep it like that for now. We might need to try to remove the other loaders if they are not needed but that is a task for another day maybe. Thanks for the explanation.

martin-trajanovski (approved):

apps/user-office-frontend/src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -870,7 +878,7 @@ const ProposalTableOfficer: React.FC<ProposalTableOfficerProps> = ({
           },
           {
             icon: ScienceIconComponent,
-            tooltip: 'Assign/Remove instrument',
+            tooltip: `Assign/Remove ${t('instrument')}`,
```
simonfernandes, line 881: Looks like this one should be lowercase

TCMeldrum:
Sorry about the delay in merging our internal git sever went down. @simonfernandes @bashanlam Would you mind reviewing it over on gitea: https://git.devs.facilities.rl.ac.uk/isisbusapps/BisAppSettings/pulls/1

PR #227, opened 2023-04-13
feat: make user selection modal more responsive and keep the actions visible in any width/height

martin-trajanovski:
## Description

Make the footer with Dialog actions visible all the time in all screen sizes. This makes it easier for any user either using a big or smaller screen to add co-proposers to a proposal or assign instrument scientists to instrument.

## Motivation and Context

The update button and the footer of the modal wasn't visible in the view and you need to scroll a lot sometimes to get to it.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-3198

## Changes

https://user-images.githubusercontent.com/6333063/231774345-69f21af2-bf10-4819-a7cd-56b477f410f0.mp4


apps/user-office-frontend/src/components/user/PeopleTable.tsx
```diff
@@ -292,12 +305,12 @@ const PeopleTable: React.FC<PeopleTableProps> = (props) => {
       // first clear the current page because if any row was unselected
       // the (un)select all option will select every rows
       // which would result in duplicates
-      setSelectedParticipants((selectedParticipants) =>
+      setSelectedParticipants?.((selectedParticipants) =>
```
Junjiequan, line 308: I wonder would it cause setSelectedParticipants(undefined=>undefined) or just skips it - refer to question 1

apps/user-office-frontend/src/components/user/PeopleTable.tsx
```diff
@@ -177,29 +198,21 @@ const PeopleTable: React.FC<PeopleTableProps> = (props) => {
     FeatureId.EMAIL_SEARCH
   )?.isEnabled;
 
-  const theme = useTheme();
-  const isLargeScreen = useMediaQuery(theme.breakpoints.up('md'));
   const api = useDataApi();
-  const { isLoading } = props;
   const { usersData, loadingUsersData } = useUsersData(query);
-  const [loading, setLoading] = useState(props.isLoading ?? false);
+  const [loading, setLoading] = useState(isLoading ?? false);
```
Junjiequan, line 203: nice cleanup again 👍

apps/user-office-frontend/src/components/user/PeopleTable.tsx
```diff
@@ -337,37 +350,33 @@ const PeopleTable: React.FC<PeopleTableProps> = (props) => {
     orderDirection: 'desc' | 'asc'
   ) => {
     if (
-      columns[orderedColumnId] &&
+      localColumns[orderedColumnId] &&
```
Junjiequan, line 353: Would be great if you can really breifly explain the reson behind why changed the naming to localColumns

Junjiequan (commented):
Nice clean up and I think the position fixed button feature definitly enhances user experience. 
Just had one question - the question 1 down below and also I pointed out where it might be causing the problem

apps/user-office-frontend/src/components/user/PeopleTable.tsx
martin-trajanovski: After removing the `props` from `props.columns` and we had `columns` defined in the same file there was a collision in the naming. I had to change the local ones with `localColumns` and keep the property columns with just `columns` name

apps/user-office-frontend/src/components/proposal/ParticipantModal.tsx
```diff
@@ -30,61 +34,92 @@ function ParticipantModal(props: {
...
-          ? proposalPeopleTable
-          : peopleTable}
+      </DialogTitle>
+      <DialogContent>
+        {participant && !isUserOfficer ? proposalPeopleTable : peopleTable}
       </DialogContent>
+      {selection && (
+        <DialogActions>
+          <div className={classes.selectedUsersText}>
+            {selectedParticipants.length} user(s) selected
```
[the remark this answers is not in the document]
martin-trajanovski: I think it shouldn't be possible because we set it initially to an empty array with this line:
`const [selectedParticipants, setSelectedParticipants] = useState<BasicUserDetails[]>([]);`
And after when we populate it we just push into that array or if we remove all the elements we set it to an empty array again.

apps/user-office-frontend/src/components/user/PeopleTable.tsx
martin-trajanovski: I think it shouldn't happen because if `setSelectedParticipants` is not undefined the `selectedParticipants` is also an array.

Junjiequan (approved):
Looks good to me

PR #257, opened 2023-06-19
When a user logs in set the highest current role possible

bolmsten:
As a user officer or instrument scientist I want the software to default me not to the user role but the highest possible role I have. 

## Motivation and Context

Eliminate tedious task of changing role for the user officer

## How Has This Been Tested

E2E

## Fixes

SWAP-3379

## Changes

Changes in frontend and backend in the user mutations, user provider

## Depends on

NA


martin-trajanovski (approved):

simonfernandes (approved):

apps/user-office-backend/src/mutations/UserMutations.ts
```diff
@@ -478,6 +478,13 @@ export default class UserMutations {
         user.id,
         roles[0]
       );
+
+      // Set the current role to the highest possible, user officer, instrument scientist, user
+      const currentRole =
+        roles.find((role) => role.shortCode === Roles.USER_OFFICER) ||
+        roles.find((role) => role.shortCode === Roles.INSTRUMENT_SCIENTIST) ||
```
mutambaraf, line 485: Is it not possible to have one iteration of find here.

simonfernandes:
Hi Fredrik, I am still looking into getting these tests passing btw. They wouldn't run at all in our local environments to start with, and they're tricky to figure out, but hopefully will get it resolved soon

bolmsten:
> Hi Fredrik, I am still looking into getting these tests passing btw. They wouldn't run at all in our local environments to start with, and they're tricky to figure out, but hopefully will get it resolved soon

Thanks, I have been looking at it myself but have not succeeded with the mock wsdl setup.  My suspicion is that the mock-uows service is not behaving in the same way as the user-office-development-authorization-server.

simonfernandes:
It looks like one of the problems is that we're not returning the correct mocked responses for users (for example, `cy.login('user1')` would in fact return a user officer), and this PR has revealed the problem. More tests are passing now but there's still some issues that I'm continuing to investigate where I have time, so thanks for your patience!

simonfernandes:
I've created a PR that I hope will fix the issues here once merged: https://github.com/UserOfficeProject/mock-uows/pull/8 🙂

PR #260, opened 2023-06-28
feat: sending SEPs review to the factory api service for generating Proposal PDF

yoganandaness:
The Proposal PDF report does not have SEPs review data. To add it, the data need to be passed to the PDF api service, that is taking care of generating Proposal PDF. 

## Motivation and Context

Show SEPs review in the Proposal PDF. 

## How Has This Been Tested

Manually

## Changes

Addition of new Request parameter ie., SEPs reviews


yoganandaness:
https://github.com/UserOfficeProject/user-office-core/assets/132274772/a33ebd76-d36b-45a8-bbb3-dc0f2ee10a4e

yoganandaness:
https://github.com/UserOfficeProject/user-office-core/assets/132274772/03385745-5336-42ab-bee8-aee21dc581e5

apps/user-office-backend/src/factory/pdf/proposal.ts
```diff
@@ -355,6 +357,14 @@ export const collectProposalPDFData = async (
...
   }
 
+  // Get Reviews
+  const reviews = await baseContext.queries.review.reviewsForProposal(
+    user,
+    proposal.primaryKey
+  );
+
+  if (reviews) out.sepReviews = reviews;
+
```
mutambaraf, line 367: Looking good but if reviews is returned as an empty array it will pass the if check.

yoganandaness: @mutambaraf Yeah, that is intentional. If the `reviews` is null, it will fail the if check. But if the `reviews` is [], it passes and assigns the [] to the `out`, which is fine.

bolmsten (approved):

mutambaraf (approved):

PR #236, opened 2023-05-15
Amend code to make it easier for user to change PI/Co-I on a proposal

RasmiaKulan:
Closes: [788](https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/788)
## Description

Made changes to make it easier for users to switch between PI's and Co-I's on a proposal
## Motivation and Context
Users can now switch between the Principal Investigator's and Co-Investigator's in the Co-Proposers table with a click of a button.
tempsnip

## Fixes
[788](https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/788)

## Changes

Files Changed:
                       user-office-frontend/src/components/proposal/ParticipantModal.tsx
                       user-office-frontend/src/components/proposal/ProposalParticipant.tsx
                       user-office-frontend/src/components/proposal/ProposalParticipants.tsx
                       user-office-frontend/src/components/proposalBooking/CreateUpdateVisit.tsx
                       user-office-frontend/src/components/questionary/questionaryComponents/ProposalBasis/QuestionaryComponentProposalBasis.tsx
                       apps/user-office-frontend/src/components/user/PeopleTable.tsx
## Tests included/Docs Updated?

- [ ] I have added tests to cover my changes.
- [ ] All relevant doc has been updated

PR #261, opened 2023-06-28
Fix setting a new STFC PI who hasn't been prev collaborated with

simonfernandes:
## Description

When choosing a new PI who isn't in the users recently collaborated list, the field appeared blank as if it hadn't worked, which was misleading. It now displays the name correctly. The bug was caused by not being able to fetch the user details due to them not being associated with the proposal. 

Part of UserOfficeProject/user-office-project-issue-tracker#856
## Motivation and Context

This issue was blocking this PR to allow quick switching between Co-I and PI: https://github.com/UserOfficeProject/user-office-core/pull/236

## How Has This Been Tested

Manual testing by adding a PI and seeing them appear, and then submitting the proposal.

## Fixes

Fixes UserOfficeProject/user-office-project-issue-tracker#856

## Changes

STFC specific files


apps/user-office-backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -44,6 +44,10 @@ export class StfcUserAuthorization extends UserAuthorization {
     Tokens.InstrumentDataSource
   );
 
+  protected userDataSource: StfcUserDataSource = container.resolve(
+    Tokens.UserDataSource
+  ) as StfcUserDataSource;
```
simonfernandes, line 49: I don't think it's the best practice to use type assertion here, but I opted to do this over adding the `isSearchableUser` function to the `UserDataSource` interface. I think in this case we can always reasonably expect that the STFC auth would use the STFC data source, but let me know if the other way is better practice.

bolmsten (approved):

apps/user-office-backend/src/auth/StfcUserAuthorization.ts
mutambaraf: Looks good to me

apps/user-office-backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -303,4 +307,16 @@ export class StfcUserAuthorization extends UserAuthorization {
...
+  async canReadUser(agent: UserWithRole | null, id: number): Promise<boolean> {
+    const readableUsers = await this.listReadableUsers(agent, [id]);
+
+    if (readableUsers.includes(id)) {
+      return true;
+    }
+
+    const isSearchableUser = await this.userDataSource.isSearchableUser(id);
+
+    return isSearchableUser;
```
mutambaraf, line 320: We can return readableUsers.includes(id) || await this.userDataSource.isSearchableUser(id) if possible.

apps/user-office-backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -518,7 +518,12 @@ export class StfcUserDataSource implements UserDataSource {
   async getRoleByShortCode(roleShortCode: Roles): Promise<Role> {
     throw new Error('Method not implemented.');
   }
+
   mergeUsers(fromUserId: number, intoUserId: number): Promise<void> {
     throw new Error('Method not implemented.');
   }
+
+  async isSearchableUser(userId: number): Promise<boolean> {
+    return !!(await this.getStfcBasicPersonByUserNumber(String(userId), true));
```
mutambaraf, line 527: We can do userid.toString() here if possible.

mutambaraf (commented):
Looking good just minor suggestions happy to approve.

PR #276, opened 2023-07-12
Hotfix: fix users not being to update proposals

### LINKED ISSUE #275: Users cannot update proposals
Users will always get a message about duplicate co-proposers, even when it's not the case

simonfernandes:
Fixes #275 

The check of `users.includes(currentUser.id)` will always return true as `users` contains co-proposers and `getProposalUsers()` returns the PI as well as co-proposers


ACLay (commented):
Surely we still need something to prevent users being added twice?

Rather than remove the check can we replace it with something that actually checks for duplicates, rather than if any user on the updated list of proposers was on the previous list

mutambaraf:
We already have checks in the frontend but we will need to split the ci and pi queries if we are to do the check.

martin-trajanovski (approved):
It seems fine by me if you need it in ASAP.

PR #279, opened 2023-07-13
866 Fix instrument sci list connect to instrument

TCMeldrum:
Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/866

## Description

When looking at instrument scientists connected to an instrument in stfc mode the user details are not being pulled from the UOWS. Add a datasource to get the details since the only users who can call this query are stfc staff.  

## Motivation and Context

Allows users officers to manage the instrument scientist list when we start using SEP and Technical reviews

## How Has This Been Tested

Scientists now appear in the list 

## Fixes

https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/866

## Changes

Add data source and query UOWS once the user numbers have been found


Scott-James-Hurley:
Would it be worth adding some tests for the list of instrument scientist attached to an instrument to make sure it works correctly in the future?

TCMeldrum:
I think they are already covered, the thing was we were using users that are in the db by default so the bug wasn't showing up in them. Now tho with the new datasource if it ever breaks we should know about it.

Scott-James-Hurley (approved):

RasmiaKulan (approved):

PR #280, opened 2023-07-14
Add limit on loaded proposals on instrument scientist table

mutambaraf:
## Description
When using the instrument scientist role  the Proposals page fails to load all proposals because we are loading too many proposals.
image

## Motivation and Context
We would like to limit the amount of proposals loaded and allow the user to select is the need more proposals.

## How Has This Been Tested
Manual

## Fixes
closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/872

## Changes
Files changed:
- apps/user-office-frontend/src/components/proposal/ProposalTableInstrumentScientist.tsx
- apps/user-office-frontend/src/hooks/proposal/useProposalsCoreData.ts
- apps/user-office-backend/src/datasources/postgres/ProposalDataSource.ts
- apps/user-office-backend/src/datasources/stfc/StfcProposalDataSource.ts


Scott-James-Hurley:
This looks good. How would one test this on local? I don't really feel like creating thousands of proposals.

mutambaraf:
> This looks good. How would one test this on local? I don't really feel like creating thousands of proposals.

I used lower numbers like 2,3, 10 to see if it is limiting the proposals for frontend. To test the backend l used graphql playground with dev and prod the query is in the issue.

Scott-James-Hurley (approved):

apps/user-office-frontend/src/components/proposal/ProposalTableInstrumentScientist.tsx
```diff
@@ -188,18 +243,26 @@ const ProposalTableInstrumentScientist = ({
     questionFilter: questionaryFilterFromUrlQuery(urlQueryParams),
     reviewer: getFilterReviewer(urlQueryParams.reviewer),
   });
+  const [queryParameters, setQueryParameters] = useState<QueryParameters>({
+    first: 2000,
+  });
+  const [maximumToload, setMaximumToload] = useState(2000);
```
martin-trajanovski, line 249: I am not quite sure why we need two states that keep the same value. As I see the `queryParameters` has property `first` which keeps the same value as `maximumToLoad`. I was thinking if we can reuse the `queryParameters.first` everywhere instead. Maybe I am wrong here and this is not possible but you can at least give it a try.

martin-trajanovski (dismissed):

apps/user-office-backend/src/datasources/stfc/StfcProposalDataSource.ts
```diff
@@ -57,7 +57,9 @@ export default class StfcProposalDataSource extends PostgresProposalDataSource {
         if (filter?.text) {
           query
             .where('title', 'ilike', `%${filter.text}%`)
-            .orWhere('abstract', 'ilike', `%${filter.text}%`);
+            .orWhere('proposal_id', 'ilike', `%${filter.text}%`)
+            .orWhere('proposal_status_name', 'ilike', `%${filter.text}%`)
+            .orWhere('instrument_name', 'ilike', `%${filter.text}%`);
```
simonfernandes, line 62: I just thought I would mention here, but it's up to you whether you want to do this in this PR or not - Emma said in the last sprint that it would be useful to additionally filter proposals by PI name and email address.

I've opened https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/913, feel free to complete it as part of this or no worries if you think it's out of scope and best left. 🙂

apps/user-office-backend/src/datasources/stfc/StfcProposalDataSource.ts
```diff
@@ -57,7 +57,9 @@ export default class StfcProposalDataSource extends PostgresProposalDataSource {
         if (filter?.text) {
           query
             .where('title', 'ilike', `%${filter.text}%`)
-            .orWhere('abstract', 'ilike', `%${filter.text}%`);
+            .orWhere('proposal_id', 'ilike', `%${filter.text}%`)
```
simonfernandes, line 60: Maybe we can save the odd request by checking that filter.text is not an empty string

simonfernandes (approved):

apps/user-office-backend/src/datasources/stfc/StfcProposalDataSource.ts
mutambaraf: Okay will include the changes.

mutambaraf: It may require a lot of changes will pick up https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/913 separately.

apps/user-office-backend/src/datasources/stfc/StfcProposalDataSource.ts
mutambaraf: I did a little check and if empty it will not pass the if statement check we must be covered.

simonfernandes: Ok sure. Don't pick up the other one though as it's very very low priority :)

apps/user-office-backend/src/datasources/stfc/StfcProposalDataSource.ts
simonfernandes: Ok sure. Don't pick up the other one though as it's very very low priority :)

jekabs-karklins (approved):
LGTM, Nice 👍

PR #287, opened 2023-07-17
Swap 3392 code editor

yoganandaness:
## Description

Introducing Code mirror editor for the PDF template editor

## Motivation and Context

The present editor is basic text area and it is difficult to write HTML code on it. Hence introducing specialized editor. 

## How Has This Been Tested

Both Manual and Cypress

### DIFF
--- a/apps/user-office-frontend-e2e/cypress/e2e/pdfTemplates.cy.ts
+++ b/apps/user-office-frontend-e2e/cypress/e2e/pdfTemplates.cy.ts
@@ -37,7 +37,7 @@ context('PDF template tests', () => {
 
       cy.contains(templateName);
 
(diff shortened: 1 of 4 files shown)


yoganandaness:
https://github.com/UserOfficeProject/user-office-core/assets/132274772/0e289c00-abfd-4ddf-9515-6a8fbc2219bb

Junjiequan (approved):
Looks good to me.
The only thing is that Fredrik is trying to avoid adding new package on FE, I think you can also ask his thoughts on this.

yoganandaness:
> 

Thanks, Yeah, i remember that. But this has been approved by Fredrik.

PR #295, opened 2023-08-07
Update UOWS WSDL

simonfernandes:
## Description

Updates the UOWS WSDL from the latest generator. I also had to make some changes to the logout in the auth file as it turns out we had mistakenly made some direct changes to the interface previously.

The WSDL is taken from https://devapis.facilities.rl.ac.uk/ws/UserOfficeWebService?wsdl
## Motivation and Context

New UOWS version

## How Has This Been Tested

I tested the logout particularly and no errors. Also had a click around and created a proposal while adding a co-i.

## Fixes

Resolves [#903](https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/903)

### DIFF
--- a/apps/user-office-backend/src/auth/StfcUserAuthorization.ts
+++ b/apps/user-office-backend/src/auth/StfcUserAuthorization.ts
@@ -242,22 +242,22 @@ export class StfcUserAuthorization extends UserAuthorization {
           return Promise.resolve('User already logged out');
         }
 
-        return await client
-          .logout(token)
-          .catch((ex: string): Rejection | string => {
-            if (ex === 'The token given is invalid') {
-              logger.logInfo(
-                'Logout failed because the cached token has since expired',
-                { token }
-              );
+        await client.logout(token).catch((ex: string): Rejection | string => {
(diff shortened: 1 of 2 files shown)


apps/user-office-backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -242,22 +242,22 @@ export class StfcUserAuthorization extends UserAuthorization {
...
+
+            return 'User already logged out';
+          }
+
+          logger.logWarn('Failed to log out user', { token });
+
+          return rejection('Failed to log out user', { token });
+        });
+
+        return 'Successfully logged out user';
```
mutambaraf, line 260: Is it possible to do it like this
```
        return await client.logout(token).then(()=>{
          return 'Successfully logged out user';
        }).catch((ex: string): Rejection | string => {
          if (ex === 'The token given is invalid') {
            logger.logInfo(
              'Logout failed because the cached token has since expired',
              { token }
            );

            return 'User already logged out';
          }

          logger.logWarn('Failed to log out user', { token });

          return rejection('Failed to log out user', { token });
        });
```

PR #296, opened 2023-08-09
feat: make proposal workflow completely async and prepare it for the proposal status actions

martin-trajanovski:
## Description

This PR has two main purposes:
1. IT is an improvement of the proposal workflow to make it completely asynchronous so it doesn't block the main thread when it is a lot of work for the workflow and do it's work behind the scenes.
2. Prepare the workflow for the status actions that needed some modifications to optimize the work with a group of proposals.

## Motivation and Context

Improvements of the proposals workflow and preparation for the next bigger feature that is proposal status actions.

## How Has This Been Tested

- manual testing
- e2e tests that cover the features and changes

## Fixes

https://jira.esss.lu.se/browse/SWAP-3405

## Changes

Mainly in the workflow engine logic.


apps/user-office-backend/src/datasources/ProposalDataSource.ts
```diff
@@ -64,8 +64,8 @@ export interface ProposalDataSource {
   deleteProposal(primaryKey: number): Promise<Proposal>;
   markEventAsDoneOnProposal(
```
enthusiastio, line 65: Maybe we can call this markEventAsDoneOnProposals

apps/user-office-backend/src/eventHandlers/customHandler.ts
```diff
@@ -215,12 +215,10 @@ export default function createCustomHandler() {
...
           if (allProposalsOnCall && allProposalsOnCall.proposalViews.length) {
-            await Promise.all(
-              allProposalsOnCall.proposalViews.map(
-                async (proposalOnCall) =>
-                  await handleWorkflowEngineChange(event, proposalOnCall)
-              )
+            const proposalPks = allProposalsOnCall.proposalViews.map(
+              (proposal) => proposal.primaryKey
             );
+            handleWorkflowEngineChange(proposalDataSource, event, proposalPks);
```
enthusiastio, line 221: Why are we sending the reference of datasource to the handler, can it not resolve this dependency itself or is there some circular dependency?

apps/user-office-backend/src/eventHandlers/proposalWorkflow.ts
```diff
@@ -18,17 +16,38 @@ enum ProposalInformationKeys {
...
   const updatedProposals = await markProposalEventAsDoneAndCallWorkflowEngine(
     event.type,
-    proposal
+    isArray ? proposalPks : [proposalPks]
   );
 
   const eventBus = resolveApplicationEventBus();
 
-  if (updatedProposals) {
+  // NOTE: If proposal status is updated manually then we need to fire another event.
```
enthusiastio, line 32: I created a ticket for it https://jira.esss.lu.se/browse/SWAP-3455

apps/user-office-backend/src/workflowEngine/index.ts
```diff
@@ -60,101 +60,133 @@ export type WorkflowEngineProposalType = {
...
-          currentWorkflowConnection.nextProposalStatusId
-        );
-      }
-    })
-  );
+  // NOTE: Filter the undefined or null items in the array.
+  return proposalsWithChangedStatuses.filter((p) => p);
 };
 
 export const markProposalEventAsDoneAndCallWorkflowEngine = async (
```
enthusiastio, line 170: This could also be plural

enthusiastio (approved):
Looks good, just some small comments

apps/user-office-backend/src/eventHandlers/customHandler.ts
martin-trajanovski: @enthusiastio I wasn't sure if this is a good idea and that's why I decided to send it from outside. Because this handler is called quite frequently I was thinking instead of resolving that on every single call I can just send it from outside and resolve only once when the handler is created. Not sure if this makes any difference or not but I guess that was the main reason.

martin-trajanovski: I think it doesn't make any difference and I will try to resolve it in the workflow engine change handler instead. Thanks for the good comments.

Scott-James-Hurley (approved):

PR #298, opened 2023-08-11
feat: define proposal status actions, interfaces and add status action engine handlers

martin-trajanovski:
## Description

This PR aims to define the proposal status actions which are going to be part of the workflow. The main purpose of the actions is to execute some handler when proposal gets into a certain status. For example: when proposal gets into `FEASIBILITY_REVIEW` we would like to send emails to the instrument scientists involved and notify them that they have something to do.
  - Created the database tables needed to support the proposal status actions
  - Added the `statusActionEngine` which is the main functionality around the status actions. After proposals get changed their statuses in the workflow we call the status engine to check if there is action set to that specific status.
  - Also added status action reset after proposal gets changed status manually. We reset the events that are executed on the proposal but we also need to reset the actions now after manual status change

## Motivation and Context

There was no way to define some actions that the system should execute if proposal gets into a certain status. Now this will open a lot of possibilities to the ones creating the workflows to define what should happen when proposal gets into specific statuses throughout the workflow.

## How Has This Been Tested

- manual tests

## Fixes

- https://jira.esss.lu.se/browse/SWAP-3416
- https://jira.esss.lu.se/browse/SWAP-3417

## Changes

Changes are mainly in the workflow and adding new things around the proposal status actions.

## Depends on

https://github.com/UserOfficeProject/user-office-core/pull/296


enthusiastio:
Question: @martin-trajanovski How do you create proposal status action? Is it done from UI?

enthusiastio:
Suggesition: it might be part of another PR but I noticed that when you go to proposal workflow and click any Proposal status that is in the workflow, and in the event selection modal leave all events empty (i,e. select none) then the there is an error. Maybe UI should disable the button Add Status Changing events if no events are selected

apps/user-office-backend/src/datasources/postgres/ProposalSettingsDataSource.ts
```diff
@@ -540,4 +544,76 @@ export default class PostgresProposalSettingsDataSource
...
+
+    const proposalStatusActions = proposalActionStatusRecords.map(
+      (proposalActionStatusRecord) =>
+        this.createProposalStatusActionObject(proposalActionStatusRecord)
+    );
+
+    return proposalStatusActions;
+  }
+
+  async updateStatusAction(
```
enthusiastio, line 589: Suggestion: This feels a little counterintuitive that the parameter accepts all object properties, including `name`, but updates only `executed` and `config`. Maybe you could use the Pick modifier or alternatively update all fields

apps/user-office-backend/src/datasources/SEPDataSource.ts
```diff
@@ -62,6 +63,10 @@ export interface SEPDataSource {
     proposalPk: number
   ): Promise<boolean>;
   getSEPProposals(sepId: number, callId: number | null): Promise<SEPProposal[]>;
+  getSEPUsersByProposalPkAndCallId(
```
enthusiastio, line 66: Nitpick: I think this could have been part of another PR where you add specific actions after adding an engine. But it's fine :+1:

apps/user-office-backend/src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -764,39 +766,42 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
...
-        proposalEventsToReset.map(
-          (event) =>
-            `${event.status_changing_event.toLocaleLowerCase()} = false`
-        )
-      ).join(', ');
+      const dataToUpdate: Record<string, boolean> = {};
+      proposalEventsToReset.forEach((event) => {
+        if (
+          !dataToUpdate.hasOwnProperty(
+            event.status_changing_event.toLocaleLowerCase()
```
enthusiastio, line 793: Suggestion: I think it would be nice to add some comments here, in which cases dataToUpdate would not have the property, or factor out the contents in condition into constant with descriptive name

enthusiastio (approved):
As I understand this PR is part of the feature, but so far looks good. I have checked the full PR and left a small comment there as well.

martin-trajanovski:
> Suggesition: it might be part of another PR but I noticed that when you go to proposal workflow and click any Proposal status that is in the workflow, and in the event selection modal leave all events empty (i,e. select none) then the there is an error. Maybe UI should disable the button Add Status Changing events if no events are selected

@enthusiastio I created a ticket for this. Once all the changes for status actions gets merged we can start testing this better and fixing bugs.

apps/user-office-backend/src/datasources/postgres/ProposalSettingsDataSource.ts
```diff
@@ -540,4 +544,76 @@ export default class PostgresProposalSettingsDataSource
         );
       });
   }
+
+  // TODO: This might need to be moved to it's own ProposalStatusActionsDataSource.ts file
```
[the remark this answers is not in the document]
martin-trajanovski: @Scott-James-Hurley I just created an issue on our side that will do some refactoring after the changes for status actions gets merged. Thanks for the reminder

apps/user-office-backend/src/datasources/postgres/ProposalSettingsDataSource.ts
martin-trajanovski: I think the other properties are not updated but they are still used in the `where` clause and in the return result.

Scott-James-Hurley (approved):

PR #301, opened 2023-08-16
Add check to prevent duplicate co-proposers

simonfernandes:
## Description

Adds a check that prevents a user from being added twice on a proposal.

## Motivation and Context

This is more of a preventative measure as the frontend (and seemingly some database constraints) should mean this isn't possible, but in the past we've had strange issues of co-proposers being duplicated affecting many proposals. There have been some large changes in this area since then so it may be that the bug has been fixed, but we want to add some logging and and a check in before our next big round.

## How Has This Been Tested

Tested that a proposal can still be created and updated.

## Fixes

Part of UserOfficeProject/user-office-project-issue-tracker#873


jekabs-karklins (approved):
LGTM

RasmiaKulan (approved):

PR #303, opened 2023-08-22
Fix stfc get users from fetching some users twice

TCMeldrum:
Small fix to stop the STFC data source from requesting users it already has found in the cache.


RasmiaKulan (approved):

simonfernandes (approved):
Nice catch 🕵️‍♂️

apps/user-office-backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -417,7 +417,7 @@ export class StfcUserDataSource implements UserDataSource {
 
     return {
       users: userDetails.sort((a, b) => a.id - b.id),
-      totalCount: finalTotalCount,
+      totalCount: userDetails.length,
```
ACLay, line 420: It doesn't look like `finalTotalCount` is used now, can it be removed?

ACLay (changes requested):

ACLay (approved):

PR #315, opened 2023-08-30
871 Add STFC grading guide

RasmiaKulan:
Closes [#871](https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/871)
## Description
Added a optional custom grade guide feature which allows userofficers to assign grading guides specific to different facility access panels. 

## Motivation and Context
The proposal system implemented the grading guide which was gerneralised to all FAPs. This feature allows to have specific grading guides for different FAPs.

## How Has This Been Tested
Manually and e2e tests.


RasmiaKulan:
Corrected the branch name from https://github.com/UserOfficeProject/user-office-core/pull/307

apps/user-office-frontend/src/components/SEP/General/SEPGeneralInfo.tsx
```diff
@@ -1,170 +1,214 @@
...
+  const sep = { ...data };
+  const classes = useStyles();
+  const { api, isExecutingCall } = useDataApiWithFeedback();
+  const hasAccessRights = useCheckAccess([UserRole.USER_OFFICER]);
+  const { t } = useTranslation();
+
+  const [customGradeGuideChecked, setCustomGradeGuideChecked] = useState(
+    sep.customGradeGuide
+  );
+  const handlecustomGradeGuideChange = (
```
enthusiastio, line 45: There is a typo with the camel casing

apps/user-office-frontend/src/components/review/ProposalTableReviewer.tsx
```diff
@@ -411,6 +411,13 @@ const ProposalTableReviewer = ({ confirm }: { confirm: WithConfirmType }) => {
         <ProposalReviewContent
           reviewId={urlQueryParams.reviewModal}
           tabNames={reviewerProposalReviewTabs}
+          sepId={
```
enthusiastio, line 414: Question: I understand that this is not part of this PR, but what are your thoughts on why sepId can be both `null` and `undefined`? Is there a conceptual difference between the two or do you think we should strive to use just one in cases like these?

enthusiastio (approved):
Overall this looks good to me

apps/user-office-backend/src/datasources/postgres/records.ts
```diff
@@ -404,6 +404,8 @@ export interface SEPRecord {
   readonly code: string;
   readonly description: string;
   readonly number_ratings_required: number;
+  readonly grade_guide: string;
+  readonly custom_grade_guide: boolean;
```
simonfernandes, line 408: It looks good but does `customGradeGuide` need to be `boolean | null`?

simonfernandes (approved):

apps/user-office-backend/src/datasources/postgres/records.ts
RasmiaKulan: It gives a type error with that change where `null` is not assignable to `boolean`.
After some digging around I found that because we have `strictNullChecks` enabled in the config `null` and `undefined` will have their own distinct types and an error can be expected if theyre used where concrete value is expected and as the `custom_grade_guide` here is a checkbox it can only be true or false and can't be left unchecked.
Hope that makes more sense 🙂

simonfernandes: Is that error because the SEP model would also need updating to `boolean | null`? My line of thinking was that it doesn't look like we are modifying any of the existing SEPs to set the `customGradeGuide` to false, so those rows will have a null value, and I wasn't sure how reading those database rows will be handled if we expect a boolean.

apps/user-office-frontend/src/components/review/ProposalTableReviewer.tsx
RasmiaKulan: As we dont have `strictNullChecks` enabled in the frontend config, `null` and `undefined` are ignored by the language. It is always optimal to use one for consistency until or unless necessary. There are instances such as [**this**](https://github.com/UserOfficeProject/user-office-core/pull/315/files#diff-925a6f70af4c7667d840a9341626e39db2f75ba5ab7920945528541156afe81aR137) where type assertion has to be used because for example in this case `sepId` couldnt be left `undefined`.

simonfernandes (approved):

PR #319, opened 2023-09-04
feat: query email templates from sparkpost api and cleanup user create mutation

martin-trajanovski:
## Description

This PR aims to integrate ESS email sending service with the status actions. We add the possibility to query the email templates from SparkPost and select from those when we need to send emails in the status actions. It also improves the email sending process where we will be able to test with real data in staging environment .

## Motivation and Context

Email status action was not integrated with SparkPost at all and only dummy email templates were shown.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-3449

## Changes

https://github.com/UserOfficeProject/user-office-core/assets/6333063/dcfaa548-01ed-4ebd-90d9-05f4c348d751

## Depends on

https://gitlab.esss.lu.se/swap/duo-kubernetes/-/merge_requests/54


enthusiastio:
I know this is still in draft but I think we should not toggle the accordion when the checkbox is selected because it can get into the wrong state if the arrow-down is pressed before
Peek 2023-09-12 21-15

janosbabik (commented):
It looks good. However, one thing I'm not sure about, especially regarding the UserMutation.create function, is whether this change allows it to be called without any restrictions in the production environment.

martin-trajanovski:
@janosbabik and @Scott-James-Hurley thanks for the good review comments. I fixed all of them and it is ready to be reviewed again now.

Scott-James-Hurley (approved):

PR #320, opened 2023-09-04
788 make it easier to change pi

bolmsten:
Needed to resubmit this PR as the branch name was breaking the flow


apps/user-office-frontend/src/hooks/user/useUserData.ts
```diff
@@ -96,7 +96,7 @@ export function useBasicUserData(userId?: number) {
   const [userData, setUserData] =
     useState<GetBasicUserDetailsQuery['basicUserDetails']>(null);
   useEffect(() => {
-    if (userId) {
+    if (userId != null && userId != undefined) {
```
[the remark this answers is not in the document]
mutambaraf: The value which is either null or undefined are always Falsy so if check must be good enough.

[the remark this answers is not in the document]
martin-trajanovski: Maybe we can just say something like `typeof userId === 'number'` as it should always be number and not `undefined` or `object`.

apps/user-office-frontend-e2e/cypress/e2e/peopleTable.cy.ts
```diff
@@ -526,4 +526,32 @@ context('PageTable component tests', () => {
...
+    it('Should allow user to swap between PIs and CoIs', () => {
+      cy.login('user1');
+      cy.visit('/');
+
+      cy.contains('New Proposal').click();
+      cy.get('[data-cy=call-list]').find('li:first-child').click();
+
+      cy.get('[data-cy=add-participant-button]').click();
+
+      cy.get('[role="presentation"] [role="dialog"]').as('modal');
```
martin-trajanovski, line 540: This line might not be needed because I don't see `modal` usage anywhere bellow in this test case.

apps/user-office-frontend/src/components/proposal/ProposalParticipant.tsx
```diff
@@ -73,12 +68,15 @@ export default function ProposalParticipant(props: {
         />
 
         <Tooltip title="Edit Principal Investigator">
-          <IconButton
-            onClick={() => setIsPickerOpen(true)}
-            className={classes.addButton}
-          >
-            <EditIcon data-cy="edit-proposer-button" fontSize="small" />
-          </IconButton>
+          <span>
```
martin-trajanovski, line 71: I think this `<span>` element is not needed here

martin-trajanovski (changes requested):

apps/user-office-frontend/src/components/proposal/ProposalParticipant.tsx
simonfernandes: It has to be left in unfortunately like Fredrik was saying earlier:

> Tooltip.js:280 MUI: You are providing a disabled `button` child to the Tooltip component.
A disabled element does not fire events.
Tooltip needs to listen to the child element's events to display the title.
>
> Add a simple wrapper element, such as a `span`.

mutambaraf (approved):

simonfernandes:
I think every time the factory default PDF is updated, it breaks the e2e test because the pdf fixture needs updating to the new file size. I don't know how to grab the generated PDF from the cypress test to commit it though 😕

martin-trajanovski:
> I think every time the factory default PDF is updated, it breaks the e2e test because the pdf fixture needs updating to the new file size. I don't know how to grab the generated PDF from the cypress test to commit it though 😕

I will update it this time. Otherwise, I just run the test `Should be able to verify pdf download and compare with a fixture` and that generates a pdf file in the `downloads` folder and I open that one to verify it is similar with the fixture and then replace the fixture.

simonfernandes:
Thanks @martin-trajanovski, I'll remember that for next time :)

simonfernandes:
Please could I get your review on these latest commits? 🙂 @mutambaraf @martin-trajanovski

martin-trajanovski (approved):
Looks good to me

PR #322, opened 2023-09-08
fix: run patches in transaction, and make sure to upsert an appsetting GRADE_PRECISION

enthusiastio:
## Description
This PR aims to add extra safety by wrapping the patch in knex transaction. In case the patch fails the rollback is automatically performed by knex.

This PR also addresses the small issue with db_patch, where db_patch tries to insert app setting. However the appsetting is already added by the code in configureESSEnvironment.ts and configureSTFCEnvironment.ts and since those run before the patches are applied, then there is a conflict when the app setting is inserted. A quick on conflict clause addresses this issue

## Motivation and Context

Make patches more resiliant

## Fixes

SWAP-3456

## Changes

Patches


bolmsten (approved):

Junjiequan (approved):
Looks good!

PR #325, opened 2023-09-13
fix: PostgresAdminDataSource.applyPatches function runs DB patches in one transaction

janosbabik:
## Description

PostgresAdminDataSource.applyPatches() function runs DB patches in one transaction

## Motivation and Context

Currently db patches are applied in separated transactions, which means that if one fails, the rest will still be applied.
This change will fix that by wrapping all patches in a single transaction.

## How Has This Been Tested

- Unit testing
- Integration testing
- Manually

## Fixes

https://jira.esss.lu.se/browse/SWAP-3499


yoganandaness (approved):

martin-trajanovski (approved):
This looks like a nice improvement to me

Scott-James-Hurley (approved):

PR #335, opened 2023-09-19
Fix: Error occurs when removing max or min value from a question #837

bashanlam:
Closes # https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/837 

## Description
When removing max or min values from a Rich Text Input or Text Input the system produces an error.  

## Motivation and Context
Resolve the error when removing max or min values from a Rich Text Input or Text Input. 

## How Has This Been Tested
Manually tested user-friendly error message. 

## Changes
Update frontend pages to resolve the error.

### DIFF
--- a/apps/user-office-frontend/src/components/questionary/questionaryComponents/RichTextInput/QuestionRichTextInputForm.tsx
+++ b/apps/user-office-frontend/src/components/questionary/questionaryComponents/RichTextInput/QuestionRichTextInputForm.tsx
@@ -1,10 +1,11 @@
 import { Field } from 'formik';
 import { CheckboxWithLabel, TextField } from 'formik-mui';
-import React from 'react';
+import React, { ChangeEvent } from 'react';
 import * as Yup from 'yup';
 
 import TitledContainer from 'components/common/TitledContainer';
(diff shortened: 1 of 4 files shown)


mutambaraf:
Should we not add e2e test for this since it was a bug not easy to find.

simonfernandes (commented):
I think there's a problem with this because `formikProps.setFieldValue('config.max', value || null)` will also consider the string '0' as falsy.

But I wonder if a cleaner way might be to use Yup for this so that validation is kept consistent and in the same place. Something like this maybe:

```typescript
  config: Yup.object({
    min: Yup.number()
      .transform((value, originalValue) => (originalValue === '' ? null : value))
      .nullable()
```

bashanlam:
After testing, it is found that Yup.number().transform is not working in this scenario because the field clear up will not trigger the function.

enthusiastio:
Maybe this PR could also widen a little bit of scope and address preventing negative values for min/max as well. Otherwise, this could also be solved in another task.
image

enthusiastio (approved):

simonfernandes (approved):
It looks good assuming the onChange value can never be null and `value.length` can't cause a null pointer

bashanlam:
Replaced

PR #336, opened 2023-09-21
Swap 3466 support sub templates

yoganandaness:
## Description

New PDF template field called Sample Declaration have been added alongside Body, Header and Footer. The logic is, if the Sample declaration design html is provided, it will be generated as an attachment with its own bookmark, else, it will not be generated.

## Motivation and Context

To support Sample Declaration

## How Has This Been Tested

Manually and e2e test

## Changes

New field called 'template_sample_declaration' in 'pdf_templates' table. 


martin-trajanovski (approved):
This looks good to me.

PR #343, opened 2023-09-25
Swap 3500 investigate add auto assignment option of sep for

yoganandaness:
## Description

Auto Assign SEP to proposal, when assigning Instrument

## Motivation and Context

At the moment, the Proposal is independently assigned to the SEP and Instrument. But there is a strong connection between SEP and Instrument in the call level, which needs to be established. Once this is done, when a Proposal is connected with the Instrument, we can get the corresponding SEP and assign the same to the Proposal automatically eliminating the need for assigning SEP manually. 

## How Has This Been Tested

Manually


yoganandaness:
https://github.com/UserOfficeProject/user-office-core/assets/132274772/2d3b66cc-037d-4741-a4e8-39ee87b25344

martin-trajanovski (commented):
I left some comments. In general looks great but just need some small changes maybe.

apps/frontend/src/components/call/AssignInstrumentsToCall.tsx
```diff
@@ -20,22 +26,69 @@ type AssignInstrumentsToCallProps = {
   assignedInstruments?: InstrumentWithAvailabilityTime[] | null;
 };
 
+// Top 100 films as rated by IMDb users. http://www.imdb.com/chart/top
```
martin-trajanovski, line 29: I think you can remove this comment as well

martin-trajanovski (approved):
There is a comment that needs to be removed but otherwise it looks good to me now.

PR #344, opened 2023-09-25
fixed the e2e issue by replacing the fixture pdf with the desired one.

yoganandaness:
## Description

Universally in all PR, e2e is failing and the reason is the generated PDF and fixture PDF are not identical(fonts are different). Due to this the e2e is failing blocking all PRs from getting approved. But the confusion still persisting is, how this issue came suddenly, since there were no recent changed in PDF fonts related work. 

## Motivation and Context

Fix PR build issues

## How Has This Been Tested

Manually

Replaced the fixture PDF with the intended PDF and test case is passing. But not really sure, if this is the great idea.


martin-trajanovski:
I am closing this because it is already done with the folder renaming PR.

PR #345, opened 2023-09-25
Swap 3395 test case of automatic assignment of instrument

yoganandaness:
## Description

e2e test cases for Automatic assignment of Instrument on a Proposal

## Motivation and Context

Earlier, a new question component for Instrument picker has been introduced with an intention to bring automatic assignment of Instrument to a Proposal, when the Question is answered. 

## How Has This Been Tested

Cypress


apps/user-office-frontend-e2e/cypress/e2e/proposals.cy.ts
```diff
@@ -1119,4 +1119,137 @@ context('Proposal tests', () => {
...
+
+      cy.login('officer');
+
+      cy.visit('/');
+
+      cy.contains('Proposals').click();
+
+      cy.contains(title).parent().find('[aria-label="View proposal"]').click();
+
+      cy.contains('td', instrument.name).should('exist');
```
martin-trajanovski, line 1252: As I understand this test case is that you view the proposal and then check if instrument name is the one that you selected in the questionary. This is more like checking the proposal questionary answer. I was wondering if we can check the proposal instrument in the proposals table. If it is auto assigned it should be present even in the table. Something like: `cy.contains(title).parent().contains(instrument.name)`.

martin-trajanovski (approved):
Looks good to me now

PR #365, opened 2023-09-28
Fix: Display preferred name if user has one

ellen-wright:
Closes #UserOfficeProject/user-office-project-issue-tracker/issues/766

## Description
Uses the user's preferred name if they have one, otherwise still uses their first name.  

## Motivation and Context
Requested in user feedback. 

## How Has This Been Tested
Manually tested on users with and without a set preferred name. 

## Changes
STFC Data source initially looks for a preferred name if their isn't one the preferredname field is set to the user's first name. 
When viewing as user preferred name is used over first name if there is one. 


mutambaraf (commented):
It would me nice if we can test this feature .

bolmsten:
Add E2E, one with a user that has a preferred name and one without

apps/frontend/src/hooks/user/useUserData.ts
```diff
@@ -113,7 +113,7 @@ export function useBasicUserData(userId?: number) {
 export interface BasicUserData {
   firstname: string;
   lastname: string;
-  preferredname: string;
   organisation: string;
   id: number;
+  preferredname: string | undefined | null;
```
mutambaraf, line 118: You can just make this optional.

ellen-wright:
Blocked by issue with mock-uows

apps/frontend/src/utils/user.ts
```diff
@@ -5,10 +5,17 @@ export const getFullUserName = (
 ): string => (user ? `${user.firstname} ${user.lastname}` : 'None');
```
martin-trajanovski, line 5: I was just wondering if we need to do the same `preferredname` logic here as well. I see that this function `getFullUserName` is quite similar to the `getFullUserNameWithEmail` but it doesn't use the preferred name logic. Maybe I am wrong but just wanted to make it clear.

apps/frontend/src/components/proposal/ProposalParticipant.tsx
```diff
@@ -55,7 +55,9 @@ export default function ProposalParticipant(props: {
           label="Principal Investigator"
           value={
             props.principalInvestigator
-              ? `${props.principalInvestigator.firstname} ${props.principalInvestigator.lastname}; ${props.principalInvestigator.organisation}`
+              ? `${props.principalInvestigator.preferredname}`
```
martin-trajanovski, line 58: Maybe it is good idea to extract this in a function to be more clear what is it about. For example `getFullUserNameWithOrganisation` something similar as `getFullUserNameWithEmail`.

martin-trajanovski (commented):
In general it looks good to me. I just left two small comments that could be checked.

mutambaraf (approved):

martin-trajanovski (approved):
It looks good to me

PR #366, opened 2023-09-28
feat: add other option for recipients by typed email in email status action

martin-trajanovski:
## Description

This PR aims to add option for typing custom recipient email that is not part of any proposal group in the email status action. 

## Motivation and Context

Previously there was no option to type some custom recipient email in the status actions and this extends the functionality

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-3530

## Changes

https://github.com/UserOfficeProject/user-office-core/assets/6333063/6c272e82-c7e2-40c2-9996-47c139123b60


apps/backend/src/resolvers/types/ProposalStatusActionConfig.ts
```diff
@@ -47,6 +52,9 @@ export class EmailStatusActionRecipientsWithTemplate {
 
   @Field(() => EmailStatusActionEmailTemplate)
   emailTemplate: EmailStatusActionEmailTemplate;
+
+  @Field(() => [String], { nullable: true })
+  otherRecipientEmails: string[];
```
janosbabik, line 57: If it is a nullable property, then the otherRecipientEmails should be marked as optional with a "?:" operator, shouldn't it?

apps/frontend/src/components/settings/proposalWorkflow/EmailActionConfig.tsx
```diff
@@ -38,6 +40,101 @@ const EmailActionConfig = ({
...
+    recipientsWithEmailTemplate.findIndex(
+      (item) => item.recipient?.name === recipientName
+    );
+
+  const isInList = (email: string, list?: string[] | null) => {
+    return list?.includes(email);
+  };
+
+  const isEmail = (email: string) => {
+    return /[\w\d\.-]+@[\w\d\.-]+\.[\w\d\.-]+/.test(email);
```
janosbabik, line 60: If I'm correct, it's not necessary when the input type is an email because it already has basic validation. Adding an additional regex like this might reject valid email addresses.

janosbabik (commented):
Overall, it looks good to me. However, when I tried to add an 'Other email address' without pressing the Enter key and, instead, simply by clicking the submit button, I encountered an 'Input validation error' because the other field was 'empty.' Firstly, it was not clear to me which field was in error.
Screenshot 2023-10-19 at 14 13 57

apps/frontend/src/components/settings/proposalWorkflow/EmailActionConfig.tsx
martin-trajanovski: I added the Yup validation here instead of custom one now. This is done because on the backend we use Yup validation for email and if I allow user to send data with only html email field validation we end up with backend input validation errors which is not ideal.

martin-trajanovski:
> Overall, it looks good to me. However, when I tried to add an 'Other email address' without pressing the Enter key and, instead, simply by clicking the submit button, I encountered an 'Input validation error' because the other field was 'empty.' Firstly, it was not clear to me which field was in error. Screenshot 2023-10-19 at 14 13 57

@janosbabik this should be improved now. I added error/information message that will point what needs to be done in this scenario.
Screenshot_15

janosbabik (approved):

PR #367, opened 2023-09-29
chore: auto merge dependabot PR workflow if all checks pass

Junjiequan:
## Description

Enabled auto-merge for Dependabot minor and patch updates upon successful checks.

The logic of Dependabot auto-merge workflow are:

- checks if pull request user is dependabot
- approve the pull request made by dependabot, if the bump is either minor or patch level. ( TOKEN with PR approval permission is required )
- enabling auto merge, so the PR is only get merged after all checks passes (auto-merge prerequisite can be configured)

## Motivation and Context

Manually approving packages version bump PR dependabot is burdensome.
If all checks pass it should automatically merge minor & patch versions.

## How Has This Been Tested

Manually

### **1. Token is used for approve dependabot PR is generated by this account with minimal permission**

image

### **2. auto-merge prerequisite configuration can be done in Settings > Branches > Edit or Add rule**

image


.github/workflows/auto-merge-dependabot.yml
```diff
@@ -0,0 +1,48 @@
+name: Dependabot auto-merge
```
martin-trajanovski, line 1: @Junjiequan I was just wondering if we can use some of the ready made actions for this. For example like: https://github.com/marketplace/actions/dependabot-auto-merge

Junjiequan: We could. Initially I tried to use exactly this 3rd party action in scicat backend.. And I was being accused for using intransparent third-party action from one of the reviewers from the other facility : D
I don't know should we use this 3rd party action or just have our own version. 
One benefit of having our own flow is ofc we have more control.

.github/dependabot.yml
```diff
@@ -2,11 +2,14 @@
...
     schedule:
-      interval: "weekly"
+      interval: 'weekly'
     pull-request-branch-name:
       # Separate sections of the branch name with a hyphen because docker hub doesn't want slashes
       # for example, `dependabot-npm_and_yarn-next_js-acorn-6.4.1`
-      separator: "-"
+      separator: '-'
+    versioning-strategy: increase-if-necessary
+    open-pull-requests-limit: 10
```
martin-trajanovski, line 14: I set this explicitly to only 2 dependabot PRs open at a time. https://github.com/UserOfficeProject/user-office-core/pull/354
It was needed because when too many PRs are open at the same time we run out of available job workers and it is hard to run our PR actions.

enthusiastio (approved):
I am not sure what was the discussion regarding enabling the auto-merge, but if this has been agreed to be good then code-wise the solution looks good to me. I think because we have so many good tests, this should work without issues.

PR #372, opened 2023-10-04
Swap 3542 proposal auto sep assignment trigger event

yoganandaness:
## Description

Trigger event, when assigning SEP automatically to a Proposal

## Motivation and Context

In the implementation of the automatic assignment of SEP to a Proposal, when an Instrument is selected, the Action event was not triggered, which affected the Proposal workflow. To address that, the code has been optimized to support event emitting by moving the logic from data source to mutation. 

## How Has This Been Tested

Manually

## Fixes

Trigger events

## Changes

Move logic from datasource to mutation.

### DIFF
--- a/apps/backend/src/models/Call.ts
+++ b/apps/backend/src/models/Call.ts
@@ -33,6 +33,16 @@ export class Call {
   ) {}
 }
 
+export class CallHasInstrument {
+  constructor(
+    public callId: number,
+    public instrumentId: number,
+    public availabilityTime: number,
+    public submitted: boolean,
+    public sepId: number
+  ) {}
+}
+
 export enum AllocationTimeUnits {
   Day = 'day',
   Hour = 'hour',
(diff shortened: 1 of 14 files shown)


apps/backend/src/models/Call.ts
```diff
@@ -33,6 +33,16 @@ export class Call {
   ) {}
 }
 
+export class CallHasInstrument {
```
janosbabik, line 36: It might be worth creating a "CallHasInstrument" model file for this because it represents a different table.

apps/backend/src/models/Call.ts
```diff
@@ -33,6 +33,16 @@ export class Call {
...
+  constructor(
+    public callId: number,
+    public instrumentId: number,
+    public availabilityTime: number,
+    public submitted: boolean,
+    public sepId: number
```
janosbabik, line 42: Is "sepId" nullable? If so, we might need to update the type.

janosbabik (commented):
It seems good to me. Some small notices. Also, it might be worth creating some unit tests for the "assignProposalsToSEPUsingCallInstrument" function.

janosbabik (approved):

PR #373, opened 2023-10-05
fix: The AddDefaultWorkflowToTheDefaultCall database patch has been fixed

janosbabik:
## Description

The hardcoded proposal_workflow_id (1) has been removed and replaced with the value of the newly inserted row's ID.

## Motivation and Context

If the "proposal_workflows" table is not empty and does not contain any rows with an ID of 1, the insert into the proposal_workflow_connections table would violate a foreign key constraint.
With this change we are using the newly created id instead of the hardcoded 1.

## How Has This Been Tested

- Manually

### DIFF
--- a/apps/backend/db_patches/0137_AddDefaultWorkflowToTheDefaultCall.sql
+++ b/apps/backend/db_patches/0137_AddDefaultWorkflowToTheDefaultCall.sql
@@ -1,15 +1,18 @@
 DO
 $$
+DECLARE
+  var_proposal_workflow_id INTEGER;
 BEGIN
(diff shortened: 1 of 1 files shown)


Junjiequan (approved):

ACLay:
Just to be sure, editing a patch like this isn't going to cause problems? I'm just used to liquibase which stores and compares hashes for applied patches, but I assume the reason we generally advise against editing patches here is because if a named patch has already run, any edits made to it later would be missed out.

janosbabik:
Hi @ACLay, the user-office-core’s applyPatches function takes all the database patches and executes them within a single transaction. This means that if any of them fail, all of them will be rolled back. Additionally, these patches are not stored in the database as records of successful executions. 
In this case, this particular patch caused an error and was not recorded in the “applied patches” table. It will be re-executed during the next application startup.
Which means, the application is not able to startup until this patch is fixed.

ACLay (approved):

martin-trajanovski (approved):

PR #376, opened 2023-10-08
Swap 3471 pdf template table shows 0 templates

yoganandaness:
## Description

Number of calls count added to pdf proposals table

## Motivation and Context

show number of calls, that are using the pdf template

## How Has This Been Tested

manually

### DIFF
--- a/apps/frontend/src/components/template/PdfTemplatesTable.tsx
+++ b/apps/frontend/src/components/template/PdfTemplatesTable.tsx
@@ -1,0 +1,133 @@
+import MaterialTable, { Column } from '@material-table/core';
+import Button from '@mui/material/Button';
+import Link from '@mui/material/Link';
+import React, { useCallback, useState } from 'react';
+
+import { ActionButtonContainer } from 'components/common/ActionButtonContainer';
+import InputDialog from 'components/common/InputDialog';
+import { Template, TemplateGroupId } from 'generated/sdk';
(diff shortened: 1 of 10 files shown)


yoganandaness:
https://github.com/UserOfficeProject/user-office-core/assets/132274772/61a808e6-f91e-4193-80c3-e005ef3418a7

apps/frontend/src/components/template/PdfTemplatesTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+          const pdfTemplateRowData = rowData as PdfTemplateRowDataType;
+
+          return pdfTemplateRowData.callCount === 0;
+        }}
+        dataProvider={props.dataProvider}
+      />
+      <CallsModal
+        templateId={selectedTemplateId}
+        onClose={() => setSelectedTemplateId(undefined)}
+      ></CallsModal>
```
martin-trajanovski, line 128: Just a small comment here because I see no children of CallsModal component maybe we can avoid using the closing tag and finish with self closing like /> instead.

martin-trajanovski (approved):
I left a minor comment but otherwise it looks great.

apps/frontend/src/components/template/PdfTemplatesTable.tsx
yoganandaness: @martin-trajanovski Thank you. Corrected.

PR #383, opened 2023-10-16
fix: External auth redirect_uri inconsistency 

enthusiastio:
## Description
This PR aims to address the issue where integration with OAuth is not working if the URL has any query parameters before the redirect to authorization happens.

The problem was that when the oAuth session was initialized the redirectUri was specified without query params, but when the code was validated the redirectUri was sent with query params, and in case there were any, the redirectUris were not matching and the server was rejecting the authorization.

## Motivation and Context
Make sure the authorization succeeds

## How Has This Been Tested

Manually against dev-auth-server and PING


mutambaraf (approved):
Looks good

PR #399, opened 2023-10-23
Swap 3464 pdf template preview

yoganandaness:
## Description

Added PDF Preview feature for the PDF editor

## Motivation and Context

It is difficult to test the PDF report template, when editing the PDF. Hence a live Preview would be an useful option. 

## How Has This Been Tested

Manual


apps/frontend/src/components/template/PdfTemplateEditorViewer.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }, [pdfTemplate, token]);
+
+  React.useEffect(() => {
+    fetchGeneratedPdfPreviewData().then((pdfBlob) => {
+      setGeneratedPdfPreviewBlob(pdfBlob);
+    });
+  }, [fetchGeneratedPdfPreviewData]);
+
+  return (
+    <Worker workerUrl="https://unpkg.com/pdfjs-dist@3.4.120/build/pdf.worker.min.js">
```
martin-trajanovski, line 40: I see that in the `package.json` you added this library as dependency `pdfjs-dist` but I don't understand why we need to load it from an external source.

EdwardHaynes (approved):
Other than Martin's comment and the failed test I'm happy to approve.

martin-trajanovski (approved):

PR #401, opened 2023-10-24
Fixed the bug of proposal SEP assignment, when Instrument picker is a…

yoganandaness:
## Description

When an Instrument Questionare is answered, an Instrument will be assigned to the Proposal. But the corresponding SEP was not assigned to the Proposal. However when trying to assign the Instrument to the Proposal manually, the SEP assignment is happening. 

## Motivation and Context

To enable automatic and seamless assignment of the Instrument and SEP to the Proposal. 

## How Has This Been Tested

Manually

## Fixes

Automatic assignment of SEP to the Proposal, when an Instrument is assigned to the Proposal.


martin-trajanovski (approved):
Seems good to me

PR #403, opened 2023-10-24
fix: Restricting SEP assignment to Proposal, only if one has not been assigned already

yoganandaness:
## Description

Perform automatic assignment of SEP to the Proposal, only if one has not been assigned already. If one has not been assigned already, just skip the assignment. 

## Motivation and Context

When a Proposal has been assigned a SEP manually or automatically, it is not a wise idea to replace the SEP during the automatic process, when an Instrument has been assigned to the Proposal. 

This avoids a scenario, where a Proposal has been cautiously assigned to a SEP by the User officer and he/she doesn't want an override. 

## How Has This Been Tested

Manually


simonfernandes (approved):

martin-trajanovski (commented):
This looks good but I left a small comment that you might want to check. And also would be nice to name the PR by the convention if it is a `fix` or a `feat`(feature) and later to `Squash and merge` to get the right release notes in the end when we have this in the staging or prod.

apps/backend/src/datasources/postgres/SEPDataSource.ts
```diff
@@ -510,6 +510,20 @@ export default class PostgresSEPDataSource implements SEPDataSource {
       });
   }
 
+  async getSEPsByProposalPks(proposalPks: number[]): Promise<SEPProposal[]> {
+    const sepProposal: SEPProposalRecord[] = await database
+      .select(['sp.*'])
+      .from('SEP_Proposals as sp')
+      .join('proposals as p', {
```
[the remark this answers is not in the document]
yoganandaness: @martin-trajanovski Thanks, that was helpful. It is done. Can you check it again.

martin-trajanovski (approved):

apps/backend/src/mutations/SEPMutations.ts
```diff
@@ -274,7 +272,6 @@ export default class SEPMutations {
     return true;
   }
 
-  @Authorized([Roles.USER_OFFICER])
```
martin-trajanovski, line 277: I am concerned a bit about this removal of the authorization check. Can't we keep that somehow because I know that it can be used from outside and like this anyone can assign proposal to a SEP I guess. Correct me if I am wrong.

martin-trajanovski (changes requested):

apps/backend/src/mutations/SEPMutations.ts
martin-trajanovski: @yoganandaness I am thinking if the one can send something like this:

`

await sepMutation.assignProposalsToSEPUsingCallInstrument(
        {
          id: 0,
          currentRole: {
            id: 1,
            shortCode: Roles.USER_OFFICER,
            title: 'UserOfficer',
          },
        } as UserWithRole,
        {
          instrumentId: instrumentId,
          proposalPks: [proposal.primaryKey],
        }
      );

`

Instead of null for the `agent` we send some data just to make `Authorized` decorator happy internally.

martin-trajanovski (approved):
Now it looks good to me

PR #404, opened 2023-10-25
Tech review data added to pdf generation immaterial of the user role

yoganandaness:
## Description

Fixed the issue of technical review information not showing up in the proposal pdf. 

## Motivation and Context

In the Proposal PDF, the template can be configured to display the technical review. This was working for sometime. Now it is stopped working. While debugging, it was figured out that the tech review data is not being passed for few user roles. Since we have a user level condition in the PDF template, we can provide the tech review data from user office no matter what the user role is and the pdf template can now control who can see and who can't see. 

## How Has This Been Tested

Manually

### DIFF
--- a/apps/backend/src/factory/pdf/proposal.ts
+++ b/apps/backend/src/factory/pdf/proposal.ts
@@ -334,18 +334,16 @@ export const collectProposalPDFData = async (
     out.attachments.push(...genericTemplateAttachments);
   }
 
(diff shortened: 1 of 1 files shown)


martin-trajanovski (approved):
Looks good to me as it just removes some check which I guess it is not relevant anymore if we want to print the review for a regular user.

PR #406, opened 2023-10-27
update: Remove runtime-generated options from DYNAMIC_MULTIPLE_CHOICE question type

janosbabik:
## Description

The "options" field in the DYNAMIC_MULTIPLE_CHOICE question type is now fetchable rather than generated every time when the field is resolved.

## Motivation and Context

Previously, the "options" field was resolved in many cases, even when the specific "FieldConfig" was not present in the query. This was because the resolver function would read and calculate all the data, resulting in an excessive number of external API calls. With this change, this behavior is now controlled, and the frontend can fetch the dynamic options only when necessary.

## How Has This Been Tested

- Manually
- Unit test

## Fixes

https://jira.esss.lu.se/browse/SWAP-3591


bashanlam (approved):

martin-trajanovski (approved):
This looks good to me.

PR #407, opened 2023-10-30
Fixed the e2e failing case by replacing the fixture file with the lat…

yoganandaness:
## Description

Usually, when there is an upgrade in the Chromium in the factory, the generated PDF file varies in some factors including size. This causes the e2e to fail in the user office. To fix this, the latest generated file is replaced with the fixture to make the test case pass. 

However, on a long term, we need to figure out a way to catch this issue whenever the factory codebase changes.


TCMeldrum (approved):
Maybe that why we were using a fixed version https://github.com/UserOfficeProject/user-office-factory/pull/330

janosbabik (approved):

martin-trajanovski:
@yoganandaness I think to be able to generate the same pdf fixture as in the tests you need to run the factory in a container locally as well and you will get the same result.

PR #414, opened 2023-11-07
Redirecting Proposal Create Page to Proposal Edit Page passively with…

yoganandaness:
## Description

During the creation of a Proposal, when refreshing the page, the Proposal data gets reset and needs to start from the beginning. However, behind the scenes, a Proposal is created in DRAFT mode and it can be accessed through My Proposal Page using the URL `/Proposal/Edit`. So the idea behind this fix is to change the URL from `ProposalCreate` to `ProposalEdit` and continue on with the Proposal form. The challenge here during the process is, that the existing React Components and state should not be disturbed as disturbing it may cause re-rendering leading to performance issues. 

So `window.history.replaceState` has been used during Proposal Creation. This changed the URL without affecting any running code. 

## Motivation and Context

To improve User experience in Proposal Creation. 

## How Has This Been Tested

Manually

## Fixes

Losing the state of the Proposal during Proposal Creation. 

## Changes

Change of URL


yoganandaness:
https://github.com/UserOfficeProject/user-office-core/assets/132274772/0ca9d539-6d77-410f-9162-b331fed36581

martin-trajanovski (commented):
This looks good to me as a simple change. I was just wondering if we can add or extend one of the existing e2e tests as well to cover this change. For example would be nice to show that in the middle of proposal creation you can refresh the page and it will stay on the same page or something similar to that. Or maybe just testing the URL after hitting 'Save and Continue'

yoganandaness:
> just testing the URL after hitting 'Save

PR #433, opened 2023-11-17
refactor(SEPs): rename SEP(Scientific evaluation panel) to FAP(Facility access panel)

martin-trajanovski:
## Description

This PR aims to rename SEPs (Scientific evaluation panels) to FAPs (Facility access panels).
1. First of all everything related to SEPs in the database is renamed and migrated to FAPs.
2. All variables, types and classes are also renamed. In variables instead of capital case for `SEP` we use camel case `Fap` because that follows our best practices much better.

## Motivation and Context

After discussions with other facilities this seems to be more standard name for these kind of panels.

## How Has This Been Tested

- e2e tests
- manual tests
- unit tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-3635

## Changes

Changes are mostly renaming and improvements.


janosbabik (approved):
Due to the volume of changes, I can't review this effectively. However, I tested it with my local db & backed up staging db, and it works—I haven't found any issues. Nice job Martin! :)

enthusiastio (approved):
We should probably merge this in soon to avoid conflict resolution

TCMeldrum (approved):
We should still remove the i8ln function calls but that can be done later if you want to get this in.

apps/frontend/src/components/DashBoard.tsx
```diff
@@ -353,20 +353,20 @@ const Dashboard = () => {
...
             component={ChangeRole}
           />
-          {isSEPEnabled && (
+          {isFapEnabled && (
             <TitledRoute
               setHeader={setHeader}
-              title={t('SEP')}
-              path="/SEPPage/:id"
-              component={SEPPage}
+              title={t('Fap')}
```
TCMeldrum, line 359: There is the function I was referring to sorry, they take the corresponding entry in the translation.json file and insert the corriponding string. So this `t()` can be removed and just to leave `Fap`. I think since it cannot find a entry in translation.json it just returns `Fap` in this case.

apps/frontend/src/components/DashBoard.tsx
```diff
@@ -353,20 +353,20 @@ const Dashboard = () => {
...
             />
           )}
-          {isSEPEnabled && (
+          {isFapEnabled && (
             <TitledRoute
               setHeader={setHeader}
-              title={i18n.format(t('SEP'), 'plural')}
-              path="/SEPs"
-              component={SEPsPage}
+              title={i18n.format(t('Fap'), 'plural')}
```
TCMeldrum, line 367: This is another example

apps/frontend/src/components/DashBoard.tsx
martin-trajanovski: I get it now. Thanks for the clarification.

martin-trajanovski: @TCMeldrum it should be updated now and you can doublecheck.

TCMeldrum (approved):
Thanks! Looks good to me

PR #434, opened 2023-11-20
Fix sorting issue for proposal table

deepaksftc:
## Description

Fixes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/881

## Motivation and Context

The user office proposal table is not able to sort by the following column principal investigator name ,PI email, submitted and notified. When we select any of the above columns, the table show blank data.

## How Has This Been Tested

Manual

## Fixes

https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/881

## Changes

Fields 'notified' and 'submitted' is added to the known fields list of proposal view.
Sorting for email and name column is disabled as per discussion.

## Screenshots

image
image
image
image

### DIFF
--- a/apps/backend/src/datasources/postgres/ProposalDataSource.ts
+++ b/apps/backend/src/datasources/postgres/ProposalDataSource.ts
@@ -52,6 +52,8 @@ const fieldMap: { [key: string]: string } = {
   statusName: 'proposal_table_view.proposal_status_id',
   proposalId: 'proposal_table_view.proposal_id',
   title: 'title',
+  submitted: 'proposal_table_view.submitted',
+  notified: 'proposal_table_view.notified',
 };
 
 export async function calculateReferenceNumber(
--- a/apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
+++ b/apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
@@ -118,6 +118,7 @@ let columns: Column<ProposalViewData>[] = [
   {
     title: 'Principal Investigator',
     field: 'principalInvestigator',
+    sorting: false,
     emptyValue: '-',
     render: (proposalView) => {
       if (
@@ -133,6 +134,7 @@ let columns: Column<ProposalViewData>[] = [
   {
     title: 'PI Email',
     field: 'principalInvestigator.email',
+    sorting: false,
     emptyValue: '-',
   },
   {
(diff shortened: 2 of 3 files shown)


apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -132,8 +132,15 @@ let columns: Column<ProposalViewData>[] = [
...
   {
     title: 'PI Email',
-    field: 'principalInvestigator.email',
+    field: 'principalInvestigator',
     emptyValue: '-',
+    render: (proposalView) => {
+      if (proposalView.principalInvestigator?.email) {
+        return `${proposalView.principalInvestigator.email}`;
+      }
+
```
bolmsten, line 141: We could use something likes this:  

Return proposalView.principalInvestigator.email ?? '';

apps/backend/src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -409,11 +411,16 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
         }
 
         if (sortField && sortDirection) {
-          if (!fieldMap.hasOwnProperty(sortField)) {
-            throw new GraphQLError(`Bad sort field given: ${sortField}`);
+          if (
```
enthusiastio, line 414: I believe there is a slight problem with this approach. If these fields are excluded then sorting on them will not be done. If you have five 10 or fewer proposals in the system then on the table the sorting will appear to work because sorting happens on frontend, but if you try sorting when you have more proposals only the visible 10 on the first page will be sorted.

I suggest disabling the sorting functionality on the `email` and `last name` in the frontend. Alternatively, we need more elaborate work in the backend to add support for being able to sort these fields

deepaksftc: Disabling the sorting functionality on email and name in frontend.

enthusiastio (approved):
LGTM now.

I realized that if you ever want to add the sorting, you could do this by recreating the view 
 and add columns to the view by joining the corresponding table
`principalInvestigator.email`
`principalInvestigator`

(see example https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/backend/db_patches/0136_AddWorkflowIdToProposalView.sql)

Then these values would be in the view and could be sorted by.

mutambaraf (approved):

PR #437, opened 2023-11-22
fix: Remove 'password' column Swap-3655 (BREAKING CHANGE)

enthusiastio:
## Description

This PR removes the `password` column from the `users` table because user management should not be part of the user office responsibilities. ESS has moved user profile information into PING and removing passwords is one of the few steps to remove obsolete fields from the User Office Software

## Motivation and Context

Move user identity and profile out of User Office Software

## How Has This Been Tested

- Manually
- E2E

## Changes
Here are the screenshots of how the user interface has been affected due to this change
### Removed change password for own account section
image
image

------
### Removed change password for other account section
image
image
------


martin-trajanovski (approved):
Seems like a nice cleanup.

PR #442, opened 2023-11-28
Run frontend component NGINX unprivileged

ErmakovDmitriy:
## Description

This PR changes configuration of NGINX in frontend component to allow running the web server in unprivileged container.

## Motivation and Context

This PR is necessary to allow the container image to run in RedHat Openshift environment.

The Openshift by default does not allow container images running in the privileged mode,
so default NGINX image cannot bind to TCP/80 port.

To support Openshift and similar Kubernetes environments, NGINX can be reconfigured 
to listen on a port number >1024 and to use `/tmp` (or a similar filesystem) to store its temporary data.

A similar configuration is used in https://hub.docker.com/r/nginxinc/nginx-unprivileged container image.

## Changes

The frontend's component container image listens on port TCP/8080 and uses `/tmp/` path for NGINX's temporary files.


ACLay (changes requested):
Rather than managing our own copies of the nginx config files for running in an unprivileged mode, can we not just switch the dockerfile to use the unprivileged image you linked in the PR as a base image?

ErmakovDmitriy:
I replaced the image with the NGINXINC/nginx-unprivileged.
Could you, please, take a look again?

apps/frontend/.dockerignore
```diff
@@ -8,6 +8,7 @@
 !tsconfig.json
 !iis.conf
 !nginx.conf
```
ACLay, line 10: If nginx.conf is being removed/renamed, can it also be removed from the dockerignore

apps/frontend/nginx.conf
```diff
@@ -1,20 +0,0 @@
...
-  listen 80;
-
-  gzip on;
-  gzip_disable "msie6";
-
-  gzip_vary on;
-  gzip_proxied no-cache no-store private expired auth;
-  gzip_buffers 16 8k;
-  gzip_http_version 1.1;
-  gzip_types text/plain application/javascript text/css application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript application/vnd.ms-fontobject application/x-font-ttf font/opentype;
```
ACLay, line 11: The new config file doesn't have the same gzip config. This was added just over a month ago to improve performance (#450). Does the unprivileged image enable this by default, or should this also be copied over too?

ACLay (changes requested):

apps/frontend/.dockerignore
ErmakovDmitriy: Thank you. Removed.

apps/frontend/nginx.conf
ErmakovDmitriy: Thank you for your suggestion.
I checked the nginx-unprivileged image and it does not have the "gzip" settings, so I added them back.

ACLay (approved):

ACLay:
@bolmsten just realised the automatic checks are being skipped on this. Does Dmitriy need to be added to https://github.com/orgs/UserOfficeProject/teams/swap?

janosbabik (approved):

PR #443, opened 2023-11-28
add github action

enthusiastio:


enthusiastio:
@ellen-wright Please take a look at this when you have a minute. This PR will aid developers with PR descriptions. The description will only be triggered if left completely empty.

ellen-wright:
The workflow looks good to me and will be happy to approve but what's happening with the two failing tests?

PR #444, opened 2023-11-29
refactor(request package): replace request with node native fetch and create simple sparkpost client because node-sparkpost used request

martin-trajanovski:
## Description

This PR aims to:
1. remove `request` package from the user office code codebase because it is deprecated for quite some time. The replacement is the native `fetch` that comes by default with node version >=18.
2. remove `sparkpost` node package because it was using `request` as well and it wasn't maintained for a while. Sparkpost themselves suggest using their API directly as an official interaction. Therefore small sparkpost client was created in our own codebase which is using node native fetch.
3. replace `axios` package used in the backend with node fetch api

## Motivation and Context

Request package was deprecated and sparkpost was also using it.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-3658

## Changes

Changes are mostly replacing `request` deprecated package with node native fetch.


apps/backend/src/eventHandlers/MailService/EmailSettings.ts
```diff
@@ -1,4 +1,156 @@
...
+  token: string;
+  /** SparkPost Enterprise API only. Required if channel is “gcm” or “apns” */
+  app_id: string;
+}
+
+interface Address {
+  /** Valid email address */
+  email: string;
+  /** User-friendly name for the email address */
+  name?: string | undefined;
```
yoganandaness, line 31: Is `?` and `undefined` redundant?? Is it necessary to have both?

yoganandaness (changes requested):

apps/backend/src/eventHandlers/MailService/EmailSettings.ts
yoganandaness: @martin-trajanovski

Bhaswati1148 (approved):

apps/backend/src/eventHandlers/MailService/EmailSettings.ts
martin-trajanovski: You were right @yoganandaness. This is redundant as I copied the types from `sparkpost-node` package directly. Now it is cleaned up with the latest commit.

Junjiequan (approved):
Looks good to me

yoganandaness (approved):

PR #457, opened 2023-12-12
Fix stfc dev env

TCMeldrum:
## Description

Vite was trying to open a browser in our docker container and crashing, so I've added a `--no-open` to stop this. 

## Motivation and Context

Frontend crashes on start in stfc development env

## How Has This Been Tested

Manually

### DIFF
--- a/apps/frontend/package.json
+++ b/apps/frontend/package.json
@@ -66,7 +66,7 @@
   "proxy": "http://localhost:4000",
   "scripts": {
     "dev": "vite",
-    "dev:docker": "npm install && vite",
+    "dev:docker": "npm install && vite --no-open",
     "build": "vite build",
     "preview": "vite preview",
     "lint": "tsc --noEmit && eslint . --ext .js,.jsx,.ts,.tsx --quiet",


Bhaswati1148 (approved):

apps/frontend/package.json
```diff
@@ -66,7 +66,7 @@
   "proxy": "http://localhost:4000",
   "scripts": {
     "dev": "vite",
-    "dev:docker": "npm install && vite",
+    "dev:docker": "npm install && vite --no-open",
```
mutambaraf, line 69: This potentially override config in the `vite.config `can we do it there if possible. It effectively closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/975 also.

TCMeldrum: Yeah it overrides `vite.config`, I assumed that ESS wanted that to stay as is so I only overrode it when launching in our dev enviornment. Happy to change it over there instead

mutambaraf: Okay happy for it to go in as is.

mutambaraf (approved):

PR #460, opened 2023-12-14
fix: opening assigned FAP on a proposal after closing proposal view modal and improve menu drawer re-opening on tablet/mobile view when switched from desktop

martin-trajanovski:
## Description
This pull request fixes the issue related to opening assigned FAP on a proposal after closing the proposal view modal and improves the menu drawer re-opening on tablet/mobile view when switched from desktop view.

## Motivation and Context
The changes are required to enhance user experience and eliminate bugs related to the UI on different devices. It resolves the problem of the drawer not being able to re-open when switching from desktop to mobile view and the assigned FAP not opening on a proposal after closing the proposal view modal.

## Changes
The main changes introduced in this PR are:
1. Added a setTimeout function in the useEffect hook in the Dashboard component to fix the bug where the drawer cannot be re-opened when switching from desktop to mobile view.
2. Updated the variable name from 'Faps' to 'faps' in various components and hooks to maintain consistency and readability.
3. Added 'fapId' to the 'fromProposalToProposalView' function in helperFunctions to fetch the 'fapId' from the proposal.

https://github.com/UserOfficeProject/user-office-core/assets/6333063/bded3db0-af46-4504-823e-c1d46746f2e0

## How Has This Been Tested?

- manual tests
- e2e tests

## Fixes Jira Issue


RasmiaKulan (approved):

apps/frontend/src/components/DashBoard.tsx
```diff
@@ -262,6 +255,17 @@ const Dashboard = () => {
     setOpen(false);
   };
 
+  useEffect(() => {
+    if (isTabletOrMobile) {
```
mutambaraf, line 259: Just to understanding when we switch to mobile the drawer is automatically closed.

martin-trajanovski: Yes you are right. That is the idea to hide the drawer by default in mobile views and make more space for the pages that need to be shown. The bug here was that you were not able to re-open it if you do the switch from desktop view to mobile which was a bit tricky.

mutambaraf: Also users normally login with either a mobile/table or desktop. But with the assumption that some users may switch between the two looks fine to fix.

mutambaraf (approved):

Junjiequan (commented):
The changes looks fine to me. 
Just trying to understand in what circumstances would we need this feature. Does the status of opened modal or drawer stays open when user switch their physical laptop to physical mobile phone?

yoganandaness (approved):

martin-trajanovski:
> The changes looks fine to me. Just trying to understand in what circumstances would we need this feature. Does the status of opened modal or drawer stays open when user switch their physical laptop to physical mobile phone?

One of the main reasons behind this was improving the experience for developers because if you are on a smaller screen (laptop for example) and open the browser debugging console it usually shrinks down to a mobile/tablet view and the menu drawer cannot be opened again for navigation which is a bit annoying.
The other reason or scenario could be if somebody uses tablet for example and opens the app in landscape view then rotates to a portrait view. Then we end up in the same situation where the menu is not accessible.

Thanks for your question @Junjiequan it's always nice to have some more explanation about why this is needed.

Junjiequan:
> > The changes looks fine to me. Just trying to understand in what circumstances would we need this feature. Does the status of opened modal or drawer stays open when user switch their physical laptop to physical mobile phone?
> 
> One of the main reasons behind this was improving the experience for developers because if you are on a smaller screen (laptop for example) and open the browser debugging console it usually shrinks down to a mobile/tablet view and the menu drawer cannot be opened again for navigation which is a bit annoying. The other reason or scenario could be if somebody uses tablet for example and opens the app in landscape view then rotates to a portrait view. Then we end up in the same situation where the menu is not accessible.
> 
> Thanks for your question @Junjiequan it's always nice to have some more explanation about why this is needed.

I didn't think of the scenerio where user enabled auto-rotate on tablet and accidently changed view, that's a very good point 👍   Thanks for the detailed answer !

Junjiequan (approved):

PR #469, opened 2024-01-02
Graphql dataloader integration in proposal backend

Bhaswati1148:
## Description
The system now makes N+1 round trips to database for fetching principal investigator details while resolving proposal query,
Let’s say it finds ‘n’ proposals in database. For each proposal found, it fetches the details of principal investigator, which will be ‘n’ calls to database or UOWS end point. This means in total there are ‘n+1’ database calls occurring.
This is much less efficient than having two database calls, one for proposals and one for principal investigators.

The efficient way to solve this would be using [graphql/dataloader](https://github.com/graphql/dataloader) for batching requests to database.

## Motivation and Context

Graphql data loader will batch the principal investigator queries for all proposals through a batch loading function. It will also cache the results for a given principal investigator as to avoid redundant calls. Details about caching mechanism can be found here https://github.com/graphql/dataloader#caching-per-request

## How Has This Been Tested

Unit test cases ran successfully.

## Fixes

Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/914

## Changes
Files Changed
- apps/backend/package.json
- apps/backend/src/buildContext.ts
- apps/backend/src/context/index.ts
- apps/backend/src/datasources/UserDataSource.ts
- apps/backend/src/datasources/mockups/UserDataSource.ts
- apps/backend/src/datasources/postgres/UserDataSource.ts
- apps/backend/src/datasources/stfc/StfcUserDataSource.ts
- apps/backend/src/loaders/UsersLoader.ts
- apps/backend/src/resolvers/types/ProposalView.ts


ACLay (approved):
This is looking good, nice job.

apps/backend/src/resolvers/types/ProposalView.ts
```diff
@@ -112,9 +112,10 @@ export class ProposalResolver {
     @Root() proposal: ProposalView,
     @Ctx() context: ResolverContext
   ): Promise<User | null> {
-    return await context.queries.user.get(
-      context.user,
+    const user = await context.loaders.user.batchLoader.load(
       proposal.principalInvestigatorId
     );
+
+    return user ? user : null;
```
Junjiequan, line 119: I have one question regarding this change. 
I see that `batchLoader.load` gets user list with `getUsersByUserNumbers`, what would happen if we from ess side try to call this query? Does it still work as the same for us?

Junjiequan (commented):
I've left one question

apps/backend/src/resolvers/types/ProposalView.ts
Bhaswati1148: `getUsersByUserNumbers` is implemented only for STFC, for ESS it needs to be implemented. I just added the method signature in `PostgresUserDataSource`, without any implementation.

Junjiequan: Thank you for the response :)
I'll go ahead and implement the batchLoader on our end and ensure all tests are thoroughly conducted. Once everything checks out, we can then proceed to merge this commit, ensuring success integration for both teams :)

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -347,7 +347,18 @@ export class StfcUserDataSource implements UserDataSource {
...
   }
+  async getUsersByUserNumbers(ids: number[]) {
+    const stfcBasicPersonDetails = await this.getStfcBasicPeopleByUserNumbers(
+      ids.map((id) => id.toString())
+    );
+    const stfcUsers = stfcBasicPersonDetails
+      ? stfcBasicPersonDetails.map((stfcbasicPerson) =>
+          toEssUser(stfcbasicPerson)
+        )
+      : null;
```
Junjiequan, line 358: We might want to remove the ternary operator which returns null in false case. The definition for stfcBasicPersonDetails type is stfcBasicPersonDetails[] here which can only return array type and will never return falsy value because empty array is always true.

apps/backend/src/datasources/UserDataSource.ts
```diff
@@ -20,6 +20,7 @@ export interface UserDataSource {
   // Read
   me(id: number): Promise<User | null>;
   getUser(id: number): Promise<User | null>;
+  getUsersByUserNumbers(id: readonly number[]): Promise<User[] | null>;
```
Junjiequan, line 23: For the returned type, it should be Promise<User[]>, as we will not receive null value

Junjiequan (changes requested):
I have commented a few minor type issues, after fixing that I think this PR is ready to be merged

apps/backend/src/loaders/UsersLoader.ts
```diff
@@ -0,0 +1,26 @@
...
+      `Inside batch loading function fetching the user details for user id(s)  ${keys}`,
+      {}
+    );
+
+    const usersList = await this.userDataSource.getUsersByUserNumbers(keys);
+    const result = keys.map((id) => {
+      return usersList?.find((user) => user.id === id);
+    });
+
+    return Promise.resolve(result);
```
enthusiastio, line 24: I think you don't need Promise.resolve(result). I think you can simply return "result"

Junjiequan (approved):

PR #480, opened 2024-01-22
Selecting beamline manager

Bhaswati1148:
## Description
Loading the beam line manager list in create/update instrument screen is slow and un-ordered

## Motivation and Context

The beam line manager list is currently populated by the api response of `getUsers`, which is fetching the entire set of user's records (approx. more than 11.5k) from [postgress database](https://github.com/UserOfficeProject/user-office-core/blob/97f28f45620dd8b959abb4c561a360e7eb411d33/apps/backend/src/datasources/stfc/StfcUserDataSource.ts#L386)
Although user role is [passed](https://github.com/UserOfficeProject/user-office-core/blob/97f28f45620dd8b959abb4c561a360e7eb411d33/apps/frontend/src/components/instrument/CreateUpdateInstrument.tsx#L41) but its never been used while querying the database.

The user numbers fetched from postgress is then used in [getStfcBasicPeopleByUserNumbers](https://github.com/UserOfficeProject/user-office-core/blob/97f28f45620dd8b959abb4c561a360e7eb411d33/apps/backend/src/datasources/stfc/StfcUserDataSource.ts#L397)
Effectively two database calls with large set of users seems to be a probable cause for slow loading of the beam line manager drop down list.

## Fixes
Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/974

## Changes
It was decided that `getBasicPeopleDetailsFromSurname` can be used to populate the beam line manager list. This was mainly because in `scheduler`, a similar functionality is used in one of the screens. So to keep things consistent across sites, it was decided to use it here too.
Below is the list of tasks identified to achieve the desired feature
- [x] Adding a new input field for surname/last name 
- [x] Adding a search button to search using the input from the surname field.
- [x] Populating the beam line manager drop down list from the search's response
- [x] Error message when no users found.
- [x] When updating an instrument, the current beam line manager's name associated with the record should be shown.

Files changed

- apps/backend/src/datasources/stfc/StfcUserDataSource.ts
- apps/e2e/cypress/e2e/instruments.cy.ts
- apps/frontend/src/components/instrument/CreateUpdateInstrument.tsx
- apps/frontend/src/hooks/user/useUsersDataByFilter.ts


TCMeldrum (commented):
Is it not possible to use the `useUsersData` hook rather than making a new one? That also passes the filter to the backend

Bhaswati1148:
> Is it not possible to use the `useUsersData` hook rather than making a new one? That also passes the filter to the backend

The new hook makes two consecutive api calls to populate the beam line manager list when updating an instrument record.
`useUsersData` is being referred at other files, changing that would break many other places, so I kept it separate.

TCMeldrum (approved):
OK looks good to me then

ellen-wright (approved):

apps/frontend/src/hooks/user/useUsersDataByFilter.ts
```diff
@@ -0,0 +1,33 @@
...
+        .then((data) => {
+          setSurName(data.user?.lastname);
+          api()
+            .getUsers({ filter: surName })
+            .then((data) => {
+              setUsersData(data.users || { totalCount: 0, users: [] });
+            });
+        });
+    }
+  }, [api, surName, instrument]);
```
Junjiequan, line 30: I'm a little confused with the reason beheind to inject surName as dependency to retrigger useEffect hook
Maybe`({ filter: data.user?.lastname}) ` and remove surName stats should be enough?

Bhaswati1148: Sorry, for the confusion. Initially added `surName` variable for some reason but later didn't use it, however forgot to remove from here.

Junjiequan (approved):

PR #483, opened 2024-01-26
979 multiple secretary pre fap

TCMeldrum:
Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/979

## Description

Added the ability to have multiple secretaries per FAP. Also fixed the FAP meeting tests not running (sorry about that one)

## How Has This Been Tested

Manually and E2E tests
## Changes

- Added new table fap_secretaries and migrate old FAPS to new table
- Added ablity to see and add multiple Secretaries in frontend

### DIFF
--- a/apps/backend/src/datasources/postgres/FapDataSource.ts
+++ b/apps/backend/src/datasources/postgres/FapDataSource.ts
@@ -42,6 +42,7 @@ import {
   createReviewObject,
   UserRecord,
   createBasicUserObject,
+  FapSecretariesRecord,
 } from './records';
 
 export default class PostgresFapDataSource implements FapDataSource {
@@ -123,7 +124,8 @@ export default class PostgresFapDataSource implements FapDataSource {
       .first()
       .then((fap: FapRecord) => {
         return fap ? createFapObject(fap) : null;
-      });
+      })
+      .then((fap) => (fap ? this.getSecretaries(fap) : null));
   }
 
   async getUserFapsByRoleAndFapId(
@@ -134,6 +136,7 @@ export default class PostgresFapDataSource implements FapDataSource {
     const fapRecords = await database<FapRecord[]>('faps')
       .select<FapRecord[]>('faps.*')
       .leftJoin('fap_reviewers', 'fap_reviewers.fap_id', '=', 'faps.fap_id')
+      .leftJoin('fap_secretaries', 'fap_secretaries.fap_id', '=', 'faps.fap_id')
       .where((qb) => {
         if (fapId) {
           qb.where('faps.fap_id', fapId);
@@ -141,7 +144,7 @@ export default class PostgresFapDataSource implements FapDataSource {
         if (role.shortCode === Roles.FAP_CHAIR) {
           qb.where('fap_chair_user_id', userId);
         } else if (role.shortCode === Roles.FAP_SECRETARY) {
-          qb.where('fap_secretary_user_id', userId);
+          qb.where('fap_secretaries.user_id', userId);
         } else {
           qb.where('fap_reviewers.user_id', userId);
         }
@@ -161,6 +164,9 @@ export default class PostgresFapDataSource implements FapDataSource {
       .distinct('s.fap_id')
       .then((faps: FapRecord[]) => {
         return faps.map(createFapObject);
+      })
+      .then((faps) => {
+        return Promise.all(faps.map(this.getSecretaries));
       });
   }
 
@@ -170,7 +176,12 @@ export default class PostgresFapDataSource implements FapDataSource {
     if (role.shortCode === Roles.FAP_CHAIR) {
       qb.where('fap_chair_user_id', userId);
     } else if (role.shortCode === Roles.FAP_SECRETARY) {
-      qb.where('fap_secretary_user_id', userId);
+      qb.join(
+        'fap_secretaries',
+        'fap_secretaries.fap_id',
+        '=',
+        'faps.fap_id'
+      ).where('fap_secretaries.user_id', userId);
     } else if (role.shortCode === Roles.FAP_REVIEWER) {
       qb.join(
         'fap_reviewers',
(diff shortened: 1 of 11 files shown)


apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -225,12 +238,16 @@ export default class PostgresFapDataSource implements FapDataSource {
...
+            totalCount: allFaps[0] ? allFaps[0].full_count : 0,
+            faps: faps,
+          };
+        });
 
-        return {
-          totalCount: allFaps[0] ? allFaps[0].full_count : 0,
-          faps,
-        };
+        return faps;
```
ACLay, line 250: I don't think there's a scoping issue, but this block feels a bit oversaturated with properties or variables that share the name `faps`

I think some of them could be eliminated to be a bit simpler, maybe start `return Promise.all(...` to elimintate one?

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
```diff
@@ -464,10 +468,12 @@ const FapProposalsAndAssignmentsTable = ({
...
-          fapSecretaryProposalCount:
-            assignedReviewer.fapMemberUserId === data.fapSecretary?.id
-              ? data.fapSecretaryProposalCount! - 1
-              : data.fapSecretaryProposalCount,
+          fapSecretaryProposalCount: data.fapSecretaryProposalCount.map(
+            (value, index) =>
+              assignedReviewer.fapMemberUserId === data.fapSecretary[index].id
+                ? value - 1
+                : value
+          ),
```
ACLay, line 476: I don't understand this code. When I try to find the origins of `data.fapSecretaryProposalCount` I ended up here: https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/backend/src/datasources/postgres/FapDataSource.ts#L319 which seems like it should just return the number, but then we do some mapping and id checks and conditional subtractions. May be lack of familiarity with the framework, but I'm feeling clueless, can you explain it to me?

TCMeldrum: I think this is required as the Secretary and Chairs details, including Proposal Count, are returned as part of the `getFap` call that is done when opening the page to edit the Panel. We don't refetch the FAP when we go between the tabs, but we do refetch the reviewers when we load the `Members` page which updates the review counts for all the reviewers but not the Secretaries and chairs. So this keeps track of the reviews for the Secretaries and Chairs count tho I agree it is a be overly complex and unclear

apps/backend/src/resolvers/types/Fap.ts
```diff
@@ -65,26 +65,30 @@ export class FapResolvers {
...
 
-  @FieldResolver(() => Int, { nullable: true })
+  @FieldResolver(() => [Int])
   async fapSecretaryProposalCount(
     @Root() fap: Fap,
     @Ctx() context: ResolverContext
   ) {
-    if (!fap.fapSecretaryUserId) {
-      return null;
+    if (!fap.fapSecretaryUserIds) {
```
ACLay, line 84: With it just returning a list of integers, is this method useful? Would it be better to return a list of objects like `{userId,count}` so the numbers can have context attached?

apps/backend/src/resolvers/types/Fap.ts
```diff
@@ -65,26 +65,30 @@ export class FapResolvers {
     );
   }
 
-  @FieldResolver(() => BasicUserDetails, { nullable: true })
+  @FieldResolver(() => [BasicUserDetails])
   async fapSecretary(@Root() fap: Fap, @Ctx() context: ResolverContext) {
```
ACLay, line 69: Should the name of this resolver be made plural?

ACLay (commented):
If the types of the secreatary fields are being changed for multiple secretaries, does this need tagging as a breaking change?

ACLay:
I kicked off the failing E2E test check a couple times when reviewing and it failed each time, I guess something broke in the latest merge?

TCMeldrum:
> I kicked off the failing E2E test check a couple times when reviewing and it failed each time, I guess something broke in the latest merge?

I'll have a look

TCMeldrum:
Turns out i had not pushed this change up but had it on local: https://github.com/UserOfficeProject/mock-uows/pull/13

Scott-James-Hurley (commented):
Was part-way through looking at this, and can finish it you want, but I reckon one review on our side is sufficient.

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -230,7 +230,7 @@ export default class FapMutations {
     }
 
     return this.dataSource
-      .removeMemberFromFap(args, isMemberToRemoveChairOrSecretaryOfFap)
+      .removeMemberFromFap(args, args.roleId === UserRole.FAP_CHAIR)
```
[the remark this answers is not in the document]
TCMeldrum: This varible is for the function in the datasource to know where to delete the member from the FAP or the Fap_reveiers table. Since sec is no longer on the FAP table we now only need to check for the Chiar role. I think ill just move this check into the datasource as it makes more sense there

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -229,15 +229,13 @@ export default class FapMutations {
...
-      .catch((error) => {
-        return rejection(
-          'Could not remove member from scientific evaluation panel',
-          { agent },
-          error
-        );
-      });
+    return this.dataSource.removeMemberFromFap(args).catch((error) => {
+      return rejection(
+        'Could not remove member from scientific evaluation panel',
```
ACLay, line 234: Seems odd that there's still a reference to the SEPs name, but I've opened https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/1008 for that

ACLay (approved):

TCMeldrum:
Didn't read all your comment lol, I went ahead and fix them here

bolmsten (approved):

PR #488, opened 2024-01-30
Fix call_ended event firing

simonfernandes:
## Description

Closes UserOfficeProject/user-office-project-issue-tracker#937

This fixes logic for updating a call's `call_ended` and `call_ended_internal` flags. These flags are only used for determining whether to fire the `CALL_ENDED` and `CALL_ENDED_INTERNAL` events respectively. We find calls where call ended = false, set the flag to true, then fire the event. The previous logic would mean that the values were always left false, so the events would keep firing at each check and we'd notice them in the STFC logs.

The new logic says that:
- The flag can be explicitly set to true or false
- When a call is updated with a future call end date, the call ended flag is changed to false.
  - If the call was closed (flag was true), this ensures the event will still fire now that it's open.
  - If the call was open, nothing changes.
- When a call is updated with a past call end date, the call ended flag is left as its previous value
  - If the call was closed (flag was true), do nothing as it's an old call being updated.
  - If the call was open (flag was false), do nothing. At the next `CALL_ENDED` check, the event needs to detect this call, update the boolean to true, and fire the event.

## Motivation and Context

Our calls at STFC don't close and therefore the call_ended event fires every hour

## How Has This Been Tested

- Manual testing
- Integration tests added

## Changes

Fix logic for updating call_ended and call_ended_internal flags.

### DIFF
--- a/apps/backend/src/datasources/postgres/CallDataSource.test.ts
+++ b/apps/backend/src/datasources/postgres/CallDataSource.test.ts
@@ -9,9 +9,12 @@ import { Proposal } from '../../models/Proposal';
 import { dummyCallFactory } from '../mockups/CallDataSource';
 import CallDataSource from './CallDataSource';
 import database from './database';
-import { createCallObject, createProposalObject } from './records';
+import { CallRecord, createCallObject, createProposalObject } from './records';
 
 const callDataSource = new CallDataSource();
+
+const tomorrow = new Date(new Date().setDate(new Date().getDate() + 1));
+const yesterday = new Date(new Date().setDate(new Date().getDate() - 1));
 
 async function getProposalsInCall(callId: number): Promise<Proposal[]> {
   const proposals = await database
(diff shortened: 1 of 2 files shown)


apps/backend/src/datasources/postgres/CallDataSource.test.ts
```diff
@@ -9,10 +9,13 @@ import { Proposal } from '../../models/Proposal';
...
 import CallDataSource from './CallDataSource';
 import database from './database';
-import { createCallObject, createProposalObject } from './records';
+import { CallRecord, createCallObject, createProposalObject } from './records';
 
 const callDataSource = new CallDataSource();
 
+const tomorrow = new Date(new Date().setDate(new Date().getDate() + 1));
+const yesterday = new Date(new Date().setDate(new Date().getDate() - 1));
+
```
enthusiastio, line 18: While it is a bit longer, this version may be easier to understand. No strong preference from my side.
```suggestion
const today = new Date();
const tomorrow = new Date(today);
tomorrow.setDate(tomorrow.getDate() + 1);

const yesterday = new Date(today);
yesterday.setDate(yesterday.getDate() - 1);


```

apps/backend/src/datasources/postgres/CallDataSource.test.ts
```diff
@@ -22,42 +25,40 @@ async function getProposalsInCall(callId: number): Promise<Proposal[]> {
...
-        });
-      }
-    })
+async function createCall(args: {
+  format?: string;
+  callEnded?: boolean;
+  callEndedInternal?: boolean;
+  endCall?: Date;
+  endCallInternal?: Date;
+}): Promise<Call> {
```
enthusiastio, line 34: A suggestion, maybe we can use default value if no argument is passed? Up to you.
```suggestion
async function createCall(args: {
  format?: string;
  callEnded?: boolean;
  callEndedInternal?: boolean;
  endCall?: Date;
  endCallInternal?: Date;
} = {}): Promise<Call> {
```

enthusiastio (approved):
In essence this looks good

apps/backend/src/datasources/postgres/CallDataSource.ts
```diff
@@ -310,14 +310,24 @@ export default class PostgresCallDataSource implements CallDataSource {
...
+                ? args.callEnded
+                : args.endCall && args.endCall.getTime() > currentDate.getTime()
+                ? false
+                : preUpdateCall.call_ended,
+              call_ended_internal: args.callEndedInternal
+                ? args.callEndedInternal
+                : args.endCallInternal &&
+                  args.endCallInternal.getTime() > currentDate.getTime()
+                ? false
+                : preUpdateCall.call_ended_internal,
```
mutambaraf, line 330: This logic looks like the same can we potential extract it out into sometime simple like 
if (args.callEndedInternal) {
    return preUpdateCall.call_ended_internal,
} else {

}

mutambaraf (approved):
Looks okay to me despite the comment.

apps/backend/src/datasources/postgres/CallDataSource.test.ts
simonfernandes: I didn't know about this, thanks! It is nicer

apps/backend/src/datasources/postgres/CallDataSource.ts
simonfernandes: I did originally have it extracted to a function, but in the end I thought it was a bit overkill and the ternary was more understandable, even if there is some duplication. I can change it if you think it's better.

```typescript
        /*
         Work out whether the call_ended and call_ended_internal flags need updating.
        */
        const determineCallEndedFlag = (
          newFlagValue: boolean | undefined,
          previousFlagValue: boolean | undefined,
          endCall: Date | undefined
        ) => {
          // Use the new value if explicitly passed in.
          if (newFlagValue) {
            return newFlagValue;
          }

          /*
           If the call end date has been changed to the future, set
           the flag to false.
          */
          if (endCall && endCall.getTime() > currentDate.getTime()) {
              return false;
          }

          /*
           Where the date has been changed to the past, leave the flag unchanged from
           its previous value. If it's false, the call end event will fire for this 
           call and update the flags, and if true it indicates an old call being updated).
          */
          return previousFlagValue;
        };
```

mutambaraf: This looks good easy to read than the chained ternary operators .

mutambaraf (approved):

PR #490, opened 2024-02-01
MaxIV, Technical review impromevement - notification for Safety team

bxo11:
## Description

During the technical review, the MaxIV Safety team needs to be informed about a problem with a sample, and there should be the possibility to assign a safety manager and send them an email template. We can use internal reviews to assign people but without notifications.

I created some basic functionality, but it is very specific and probably not everyone needs this feature. Therefore, I need your opinion and any feedback on how to improve it to be more generic.

What do you think about adding a new role, Safety Manager? Maybe it's still better to stick with internal reviewers instead.

I was considering adding new email recipient groups to the proposal workflow status action -> email action, like "safety managers", "assigned internal reviewers", or something similar that could be more generic instead of creating another case in the email handler file that will send an email.

## Changes
- New button and modal in Technical Review -> Internal Reviews
- New role: Safety Manager
- New event: PROPOSAL_SAFETY_NOTIFIED.


bolmsten:
I agree with you that a email is needed. I think we do not need to send it explicitly, instead it could be sent automatically when they are added. And the template used could be set in the feature settings

PR #491, opened 2024-02-01
981 display summary of scored proposals

RasmiaKulan:
## Description
Closes [981](https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/981)

Added a `Reviews` column in the `FAP Proposals and Assignments` table which displays the number of proposals reviewed out of total number of proposals to review.
## Motivation and Context

Initially, the system required to click the `Show Reviews` drop down to check if proposals were graded or not. Having this feature makes it easier for FAP members to quickly check if there are any outstading proposals that require grading especially for calls like ISIS which require two reviewers per proposal.

image

## How Has This Been Tested
Manually

### DIFF
--- a/apps/e2e/cypress/e2e/FAPs.cy.ts
+++ b/apps/e2e/cypress/e2e/FAPs.cy.ts
@@ -608,6 +608,8 @@ context('Fap reviews tests', () => {
 
       cy.finishedLoading();
 
+      cy.contains('0 / 1').should('be.visible');
+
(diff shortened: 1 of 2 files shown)


martin-trajanovski (approved):

Scott-James-Hurley (commented):
I think it would be good to have some testing around this. Would that be possible?

martin-trajanovski (approved):

Scott-James-Hurley (approved):

PR #492, opened 2024-02-01
Refactor: Update meeting export for stfc mode

TCMeldrum:
Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/988

## Description

For the current round, Emma has requested these rows for the meeting export. This is subject to change when the stfc start using the meeting component. 

Columns requested: 
Proposal Reference Number
Principal Investigator
PI Country
Instrument Name
Requested Time
Proposal Title
Reviewer 1 score
Reviewer 1 review comment
Reviewer 2 score
Reviewer 2 review comment
Average score

## How Has This Been Tested

Manually

## Changes

More data is passed to the populateRow and fapDataRow functions and the instrument's foreach has been made a asynchronous map since we need data to be fetched from our user db for the PIs contrey. 


Scott-James-Hurley (approved):
Outside the scope of this PR: should we get the factory stuff under test?

apps/backend/src/factory/xlsx/stfc/StfcFapDataRow.ts
```diff
@@ -1,20 +1,46 @@
...
+  reviews: Review[] | null
 ) {
+  const daysRequested = proposalAnswers
+    ?.find((step) =>
+      step.fields.find(
+        (answer) => answer.question.naturalKey === 'days_requested'
+      )
+    )
+    ?.fields.find((answer) => answer.question.naturalKey === 'days_requested')
+    ?.value.value;
```
ACLay, line 34: This feels like there's a bit of duplicate filtering going on to find the questionary with the days_requested answer in it, and then the days_requested answer. I think it could simplify to something like:
```
proposalAnswers
  .flatMap(step => step.fields)
  .find(answer => answer.question.naturalKey === 'days_requested')
```

It also seems odd that only certain steps in the call chain are null conditional. I'd have thought if we need to have the first value protected, then the second one would also have a possibility of being called on null, and would need the same protection.

RasmiaKulan (approved):

ACLay (changes requested):

apps/backend/src/factory/xlsx/stfc/StfcFapDataRow.ts
TCMeldrum: When I was writing this I thought the double `find` is slightly more efficient than the `flatMap` as I does not need to construct a new array? This way is much more readable and I guess the cost différance cant be that much.

ACLay (approved):

apps/backend/src/factory/xlsx/fap.ts
```diff
@@ -180,46 +184,69 @@ export const collectFaplXLSXData = async (
...
+      const proposalReviews = proposalsReviews[indx];
+      const proposalPrincipalInvestigators =
+        proposalsPrincipalInvestigators[indx];
+      const technicalReviews = proposalsTechnicalReviews[indx];
+      const fapProposals = instrumentsFapProposals[indx];
+      const fapMeetingDecisions = proposalsFapMeetingDecisions[indx];
+      const proposalsAnswers = instrumentProposalsAnswers[indx];
+
+      const rows = await Promise.all(
+        proposals.map(async (proposal, pIndx) => {
```
martin-trajanovski, line 216: Not sure why do we need an async function here. I don't see anything asynchronous inside this function.

martin-trajanovski (commented):
There is a one comment left that needs a bit more attention otherwise it looks good.

apps/backend/src/factory/xlsx/fap.ts
TCMeldrum: I made it async as the STFC `fapDataRow` is asynchronous: https://github.com/UserOfficeProject/user-office-core/pull/492/commits/9f0e8477c9836201c3c00b625e6a1a5b57114bba#diff-0f8ddd45ec0c99053d9e83df998c997723ace6e88b91a18f19f20e1e28d4be3dR36

martin-trajanovski: I see now. Thank you for clarification.

martin-trajanovski (approved):

PR #496, opened 2024-02-02
Add a view to simplify proposal answer access in SQL queries

ACLay:
## Description

Adds a new view for accessing a proposals answers.

Sample output of the SELECT statement on our dev db with some sorting applied:
image

For common cases; plain text inputs, radio buttons & number inputs where we only want the number, the `first_value` column pulls out the answer in an easy access text format. For anything more complex, `json_value` returns a json object (slightly extracted from) the answer table for processing.

Skipped patch number 146 as I saw that in use in #483

## Motivation and Context

We've got a tool full of SQL queries to allow STFC staff to access select data from our various systems. Often this needs to pull in answers from the proposal system, and if you're starting with the proposal's ID then it's several joins to get from that to the proposals answers and the key of the question they're for. With this view it'd only need a single join. Applied over all our queries it'll add up to a lot of lines saved and simpler to read queries.

## How Has This Been Tested

I've run the view's `SELECT` statement against our databases, and applied the patch to my local environment without issue.

## Fixes

Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/994


TCMeldrum (approved):

simonfernandes (commented):
This looks really useful! I think it's missing the generic template title (`gt.title`) though, is that intentional?

ACLay:
Turns out I had no idea how generic templates work, so missing that wasn't intentional, and the joins weren't quite right.

ACLay:
I've added the generic template title, as well as their natural key. The joins I'd been using were reporting non-template answers as part of each template, so I've replaced that with a UNION over 2 selects.

TCMeldrum (approved):

simonfernandes (approved):

ACLay:
I've renamed the json column to avoid clashing with the name of a built in function in our trino SQL environment. That can be solved by adding quotes around the column name, but since this is a new view for convenience with no existing users to affect, lets just rename the column to be as easy to use as possible.

simonfernandes (approved):

TCMeldrum (approved):

janosbabik (approved):

PR #512, opened 2024-02-13
fix: for throw error with proposal template question character limit

YufanKambang:
Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/924 

## Description

I have added a max character limit to the Yup schema.

## Motivation and Context

An error was found to be thrown to the user when they exceeded a character limit of 256, in the proposal template question.
Wanting to stop this error being given to the user as it also prevented them from saving their question description.

## How Has This Been Tested

Tested the code changes on office.local.
Entered an entry with a character count larger than the limit, on the limit and less than the limit.

## Changes

The changes introduced, is adding an extra constraint to the Yup schema builder, specifically the .max()


Scott-James-Hurley:
Would it be an idea to add an E2E test for this?

YufanKambang:
what is an E2E test?

Scott-James-Hurley:
> what is an E2E test?

End-to-End test. It tests the software's functionality from start to finish (so the backend, frontend, factory, etc.). The E2E tests can be found [here](https://github.com/UserOfficeProject/user-office-core/tree/develop/apps/e2e/cypress/e2e). They take the form of a bot using the website in a prescribed manner and verifying it's behaving as expected. They're really useful for testing the UI and making sure all the pieces of your software interact correctly. Would be happy to talk about it more if you'd like.

YufanKambang:
yes please Scott, i dont know how to add the E2E test

martin-trajanovski (approved):

martin-trajanovski (approved):
It looks good to me but as mentioned would be nice to also test this.

mutambaraf:
This `  20:31  error  Replace .required('Question·is·required') with ··········.required('Question·is·required')··········  prettier/prettier` looks like prettier do we need to update this rule because the reference statement looks okay to me.

Scott-James-Hurley:
> This ` 20:31 error Replace .required('Question·is·required') with ··········.required('Question·is·required')·········· prettier/prettier` looks like prettier do we need to update this rule because the reference statement looks okay to me.

Personal preference, I suppose.

Before:
```
validationSchema={Yup.object().shape({
  naturalKey: naturalKeySchema,
  question: Yup.string().required('Question is required').max(256, 'There is a 256 max character limit'),
})}
```
After:
```
validationSchema={Yup.object().shape({
  naturalKey: naturalKeySchema,
  question: Yup.string()
    .required('Question is required')
    .max(256, 'There is a 256 max character limit'),
})}
```

I prefer after.

mutambaraf (approved):

apps/e2e/cypress/e2e/genericTemplates.cy.ts
```diff
@@ -726,6 +726,43 @@ context('GenericTemplates tests', () => {
 
       cy.contains(genericTemplateTitle);
     });
+
+    it('should be a character limit to the template proposal question for office user', () => {
+      cy.createTemplate({
+        name: proposalTemplateName,
+        groupId: TemplateGroupId.PROPOSAL,
+      });
+      const char256 =
```
martin-trajanovski, line 735: Instead of hardcoding this you can maybe use `faker.string.alphanumeric(256)` to generate random 256 characters long string.

YufanKambang: that sounds better than what i have going on, ill add that change

YufanKambang: image

martin-trajanovski: This is what I found on their documentation site https://fakerjs.dev/api/string.html#sample

deepaksftc (approved):

PR #515, opened 2024-02-15
Feat: ROR integration for institution

Junjiequan:
## Description
This PR aims to adapt ROR integration for Institutions. Manual ROR ID updates for already used ROR certified Institution in prod environment is required. 

Changes including :

1. Created DB patch for institutions table. Added `ROR ID` field and Dropped  `VERIFIED` field.
2. In the Institution UI, removed `verified` column and added `ROR ID` column.
3. In the Institution UI, removed `Create` button.
4. When a user login via OAuth, creates a new institution record if user's institution is not found, or updates with ROR ID otherwise ( ROR ID can be null ).

## How Has This Been Tested

- Manually

## Changes

https://github.com/UserOfficeProject/user-office-core/assets/78078898/1d334fcb-fabc-4c5a-b7a6-866931a6bc5b


apps/frontend/src/components/institution/InstitutionTable.tsx
YufanKambang, line 1: Is this a link to an external site?

apps/backend/src/auth/OAuthAuthorization.ts
```diff
@@ -102,17 +103,46 @@ export abstract class OAuthAuthorization extends UserAuthorization {
...
+        );
+
+    if (!institution) {
+      let institutionCountry = userInfo.institution_country
+        ? await this.adminDataSource.getCountryByName(
+            userInfo.institution_country as string
+          )
+        : null;
+
+      if (!institutionCountry || !institutionCountry.countryId) {
```
bolmsten, line 130: Do we need to check both institutionCountry or institutionCountry.countryId? All countries does have an id?

bolmsten (approved):

apps/frontend/src/components/institution/InstitutionTable.tsx
bolmsten: We are using ROR as a way of uniquely identifying a organisation

Junjiequan: > Is this a link to an external site?

Yes, it is external site. For example, the link will be like https://ror.org/01aj84f44

apps/backend/src/auth/OAuthAuthorization.ts
Junjiequan: We don't need to check institutionCountry.countryId here. 
InstitutionCountry value can only be `Country | null` type. I will remove the redundant condition

Junjiequan:
pull request from user-office-development-authorization-server needs to be merged to pass the e2e

apps/backend/src/datasources/UserDataSource.ts
```diff
@@ -75,11 +75,7 @@ export interface UserDataSource {
   ): Promise<User>;
   ensureDummyUserExists(userId: number): Promise<User>;
   ensureDummyUsersExist(userIds: number[]): Promise<User[]>;
-  createInstitution(
-    name: string,
-    verified: boolean,
-    countryId?: number
-  ): Promise<number>;
+  createInstitution(name: string, countryId?: number): Promise<number>;
```
janosbabik, line 78: Is `rorId` not missing here?

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -353,7 +353,7 @@ export interface InstitutionRecord {
   readonly institution_id: number;
   readonly institution: string;
   readonly country_id: number;
-  readonly verified: boolean;
+  readonly ror_id?: string;
```
janosbabik, line 356: Why the `ror_id` is optional? Due to the database patch, the column will be added to the Institution table, isn't it?

apps/backend/src/models/Institution.ts
```diff
@@ -3,6 +3,6 @@ export class Institution {
     public id: number,
     public name: string,
     public country: number,
-    public verified: boolean
+    public rorId?: string
```
janosbabik, line 6: Necessary optional?

janosbabik (changes requested):
Looks good, just some small notes.

apps/backend/src/models/Institution.ts
Junjiequan: optional is necessary here, Institution can be without rorId

janosbabik (approved):

PR #524, opened 2024-02-21
Add option to omit images from proposal pdf

deepaksftc:
## Description

Fixes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/959

## Motivation and Context

Adding an option to file upload question template to specify if an image need to be included or not when generating a proposal pdf. 

## Changes
Omit from pdf checkbox is added for File Upload question type. 
Depending on this Boolean field, the image fetch from db is controlled from backend.

## How Has This Been Tested

Manually.

## Fixes

https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/959


mutambaraf:
If possible we can just exclude some questions when we actually download the question so as to make it easier to download pdf with images and another without. Hope l did understand this much better.

mutambaraf (approved):

bolmsten (approved):

PR #531, opened 2024-02-27
fix: change file upload error message

ellen-wright:
## Description
Closes: UserOfficeProject/issue-tracker/issues/663

## Motivation and Context
Changes error message when no file is uploaded on required question to be more user friendly. 

## How Has This Been Tested
Manually and changes added to E2E test.

## Changes
Added a new file upload validation schema with new error message. 


martin-trajanovski (commented):
It looks good to me but as I can see we keep all the validation schemas for the other questions in the `@user-office-software/duo-validation` package. Maybe it is a good idea to add this one over there as well.

martin-trajanovski:
> > It looks good to me but as I can see we keep all the validation schemas for the other questions in the `@user-office-software/duo-validation` package. Maybe it is a good idea to add this one over there as well.
> 
> @martin-trajanovski We seem to have some validations schemas in the repo should these be migrated as well (although out of scope of this issue maybe). image

@ellen-wright yes you are right we should move those over to the duo-validation and keep everything in one place. I also see that for your specific case the validation was in the duo-validation package and I would prefer to keep it there.

martin-trajanovski (approved):

TCMeldrum (approved):

PR #537, opened 2024-02-28
Fix: STFC Tech Reviewers not loading

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1012

## Description
This PR fixes the issue of STFC Tech Reviewers not loading in the proposal view.

## Motivation and Context
The change was required to solve the problem where the Tech Reviewers for STFC were not being loaded

## Changes
  - Added the `getProposalsFromView` method in `StfcProposalDataSource` to include details of technical reviewers in the proposal views fetching them from the UOWS.


apps/backend/src/datasources/stfc/StfcProposalDataSource.ts
```diff
@@ -134,4 +137,64 @@ export default class StfcProposalDataSource extends PostgresProposalDataSource {
...
+
+  async getProposalsFromView(
+    filter?: ProposalsFilter,
+    first?: number,
+    offset?: number,
+    sortField?: string,
+    sortDirection?: string,
+    searchText?: string
+  ): Promise<{ totalCount: number; proposalViews: ProposalView[] }> {
+    const proposals = await super.getProposalsFromView(
```
martin-trajanovski, line 149: It is good that you call the parent class method here and it won't require a lot of additional changes or conflict resolutions against the multi-instrument proposals approach. 👍 💯

apps/backend/src/datasources/stfc/StfcProposalDataSource.ts
```diff
@@ -134,4 +137,64 @@ export default class StfcProposalDataSource extends PostgresProposalDataSource {
...
+    );
+
+    const technicalReviewers = new Set(
+      proposals.proposalViews
+        .filter((proposal) => !!proposal.technicalReviewAssigneeId)
+        .map((proposal) => proposal.technicalReviewAssigneeId.toString())
+    );
+
+    const technicalReviewersDetails =
+      await stfcUserDataSource.getStfcBasicPeopleByUserNumbers(
```
martin-trajanovski, line 165: I guess this will slow down the proposal view query a bit but maybe not that much. We can have better overview once it gets tested properly with real data.

TCMeldrum: Hopefully there should not be more than 20-40 different reviewers so it shouldn't slow it down too much, but we will see. I don't really see another way of populating that field since we only store user numbers in proposals db.

Scott-James-Hurley (approved):
Sorry it took me so long to look at this.

TCMeldrum:
> Sorry it took me so long to look at this.

No worries!

PR #543, opened 2024-03-05
feat: added USER_SEARCH_FILTER featureId for conditional surname search of beamline managers

Junjiequan:
## Description

On the instrument page, the beamline manager list in the create/edit instrument modal is initially empty and requires a surname to populate it. This doesn't align with ESS user experience needs. To address the issue while considering STFC's limitations on fetching beamline managers, including a new featureId flag to toggle the surname search could reconcile ESS and STFC use cases.


apps/backend/src/config/stfc/configureSTFCEnvironment.ts
```diff
@@ -83,6 +83,7 @@ async function enableDefaultStfcFeatures() {
       FeatureId.STFC_IDLE_TIMER,
       FeatureId.TECHNICAL_REVIEW,
       FeatureId.FAP_REVIEW,
+      FeatureId.USER_SURNAME_SEARCH_FILTER,
```
martin-trajanovski, line 86: Maybe it is better to name the feature just `USER_SEARCH_FILTER`. Now it is searching by surname but it might change in the future I guess and being able to search by email as well for example.

martin-trajanovski (commented):
In general it looks good. I left a small comment on the feature naming and I also think that you should name the PR starting with `feat:` instead of `fix:` as this looks more like a feature addition than fixing a bug.

martin-trajanovski (approved):

PR #550, opened 2024-03-11
feat: mass assign needed reviewers to FAP proposals

Scott-James-Hurley:
Closes: UserOfficeProject/issue-tracker/issues/964

## Description

Adds a button to the `Proposals and Assignments` tab in Fap page that assigns all reviews still needed for the proposals for a given FAP as evenly as possible to the FAPs reviewers. An assumption I made was that there will only be one call in the review phase for a FAP at any given time.

image

## Motivation and Context

This will make easier it to assign reviews when there are a large number of proposals. Tries to ensure reviewers have an even workload.

## How Has This Been Tested

- Added unit tests
- Added E2E tests
- Tested manually

If you're testing this manually, the call containing the proposals you want assigned with have to have and end date in the past and the `call_fap_review_ended` field must be false (The call must be in the review phase).

## Changes

- `apps/backend/src/datasources/postgres/FapDataSource.ts` - adds methods for getting FAP reviews for a FAP and call, assigning a reviewer to multiple reviews, getting the call currently in the review phase for a FAP and getting a map for FAP proposals to the number of reviews they need
- `apps/backend/src/mutations/FapMutations.ts` - adds a method for assigning all remaining reviews needed as evenly as possible (takes into account assignments from other FAPs) to all reviewers on the FAP
- `apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx` - adds a button for mass assigning reviews and logic for updating the view.


apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -381,6 +383,22 @@ export default class PostgresFapDataSource implements FapDataSource {
       });
   }
 
+  async getFapReviewsByCallAndFap(
+    callId: number,
+    fapId: number
+  ): Promise<Review[]> {
+    const fapReviews: ReviewRecord[] = await database
+      .select(['sr.*'])
+      .from('fap_reviews as sr')
```
ACLay, line 392: Why are the short names `sr` and `sp`? They don't seem to line up with the table names.

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -779,6 +797,80 @@ export default class PostgresFapDataSource implements FapDataSource {
     return fapUpdated;
   }
 
+  async assignMemberToFapProposals(
```
ACLay, line 800: This looks very similar to the existing `assignMemberToFapProposal`. I wonder if it might be better to change that method to something like `assignMembersToFapProposals(assignments: {proposalPk, fapId, memberId}[])` and tweak the places that call it instead of near duplicating it?

apps/backend/src/mutations/FapMutations.ts
ACLay, line 1: Stopping here as I just saw the time.

ACLay (changes requested):
You mentioned that the new methods assume a FAP only has a single open call. I'm okay to bake that assumption in, but can we add some docstrings on the relevant methods/mutations(?) to make it easier for anyone using the method in the future to see that that's the case. Or maybe have it fail with an error if there's multiple?

apps/backend/src/datasources/postgres/FapDataSource.ts
Scott-James-Hurley: I don't know where the 's' comes from. I decided it best to follow the naming convention established in the other queries.

apps/backend/src/datasources/postgres/FapDataSource.ts
Scott-James-Hurley: Good point. Would it be an idea to open a separate issue to do that? Having a look, this change would make the PR a fair bit bigger and harder to review I reckon.

ACLay: Sure, that sounds reasonable

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -379,6 +380,128 @@ export default class FapMutations {
...
+  /**
+   * Assigns all remaining reviews needed for a FAP among its members. Attempts to distribute reviews as evenly as possible.
+   * Assumes there is ony one call in the review phase for a FAP at a time.
+   * @param agent User
+   * @param args ID of the FAP to mass assign reviews for
+   * @returns Updated FAP
+   */
+  @Authorized([Roles.USER_OFFICER, Roles.FAP_SECRETARY, Roles.FAP_CHAIR])
+  @EventBus(Event.FAP_MEMBER_ASSIGNED_TO_PROPOSAL)
+  async massAssignReviews(
```
ACLay, line 392: While this method looks like it'll work, I needed to write out a summary to figure out what it's doing, and I feel like it might be longer and more complex than it needs to be. 

At present I think it's something like this:

- Get how many proposals each reviewer is assigned to in the current round
- Get the highest number of reviews a reviewer has
- Count the total number of reviews needed for each reviewer to have that maximum
- Count the total number of reviews needed by the proposals
- Calculate how many reviews to assign to each proposer
- Assign reviews by looping over reviewers:
  - lookup from earlier how many reviews they need to give
  - looping over proposals until you've found that many that they're not already on
  - assign the reviewer to those proposals


I think a simpler process would be:
- Loop over each proposal
  - while it has fewer reviewers than needed
    - find the reviewer with the lowest number of reviews who isn't assigned to this proposal
    - assign them

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -379,6 +380,121 @@ export default class FapMutations {
...
+    for (const reviewer of reviewers) {
+      reviewersAssignedReviewsMap.set(
+        reviewer,
+        await this.dataSource.getFapReviewerProposalCountCurrentRound(
+          reviewer.userId
+        )
+      );
+    }
+
+    for (const review of [...reviewsNeededMap.keys()]) {
```
ACLay, line 421: It looks like `reviewsNeededMap` is a `Map<FapProposal, number>`, but we're putting those proposal objects into a variable called `review`. Could that variable be renamed to line up with the type?

Is this also missing an inner loop for the number of reviews the proposal needs? It looks like this'll only assign 1 reviewer to each proposal that needs reviewers

ACLay (changes requested):

ACLay (approved):

apps/backend/src/resolvers/mutations/AssignMembersToFapMutation.ts
```diff
@@ -125,6 +131,14 @@ export class AssignMembersToFapMutation {
     );
   }
 
+  @Mutation(() => Fap)
+  async massAssignReviews(
```
martin-trajanovski, line 135: Maybe it is better to use `Fap` keyword in the mutation name. Something like `massAssignFapReviews` sounds better for me because it is much more clear when you use it from the frontend, or in the schema playground, that this is Fap related reviews and not some other review.

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -381,6 +382,123 @@ export default class FapMutations {
       });
   }
 
+  /**
+   * Assigns all remaining reviews needed for a FAP among its members. Attempts to distribute reviews as evenly as possible.
```
martin-trajanovski, line 386: This is a good note here as the functionality is a bit complicated it is good to have an overview of what we want to actually achieve here.

martin-trajanovski (commented):
In general it looks good to me and it is a nice functionality to have I think. I left some comments that might require some attention. Another small thing is that it might be beneficial to improve the title of the PR as if it gets `Squash and merge` would be nice to have a good commit message and note in the release about the functionality that was added. For example something like: `feat: mass assign needed reviewers to FAP proposals` or something similar.

apps/backend/src/datasources/postgres/FapDataSource.ts
Scott-James-Hurley: Thanks for clearing up the confusion. I'll pop in an issue to change that, don't think it's relevant to this PR.

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -49,6 +49,8 @@ import {
   InstitutionRecord,
 } from './records';
 
+const DEFAULT_NUM_REVIEWS_REQUIRED = 2;
```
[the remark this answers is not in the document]
Scott-James-Hurley: Probably best to un-bury now, thanks for pointing this out.

apps/backend/src/datasources/mockups/AdminDataSource.ts
```diff
@@ -41,6 +41,7 @@ export class AdminDataSourceMock implements AdminDataSource {
       new Settings(SettingsId.FEEDBACK_MAX_REQUESTS, '', '2'),
       new Settings(SettingsId.FEEDBACK_EXHAUST_DAYS, '', '90'),
       new Settings(SettingsId.FEEDBACK_FREQUENCY_DAYS, '', '14'),
+      new Settings(SettingsId.DEFAULT_NUM_REVIEWS_REQUIRED, '', '2'),
```
martin-trajanovski, line 44: Maybe including `FAP` keyword here is beneficial as well because it is not clear what reviews is this for.

apps/backend/src/models/Settings.ts
```diff
@@ -27,6 +27,7 @@ export enum SettingsId {
   FEEDBACK_MAX_REQUESTS = '2',
   FEEDBACK_FREQUENCY_DAYS = '14',
   FEEDBACK_EXHAUST_DAYS = '90',
+  DEFAULT_NUM_REVIEWS_REQUIRED = '2',
```
martin-trajanovski, line 30: Including `FAP` keyword in this setting name would be nice.

martin-trajanovski (approved):
Just a small rename of the setting but otherwise I think it looks good now

ACLay (approved):

PR #559, opened 2024-03-18
fix: fap proposal count

TCMeldrum:


Scott-James-Hurley (approved):
Does the ESS have a use case for included drafts in the count?

bolmsten (approved):

bolmsten:
> Does the ESS have a use case for included drafts in the count?

No, we do not

PR #563, opened 2024-03-22
feat: add availability of PI & Co-I in status action email templates

simonfernandes:
## Description

In many of our STFC email templates, we state the PI and Co-Proposers in the template, regardless of the recipient. This makes sure that they're made available in the templates.

## Motivation and Context

Part of the process of STFC switching over to proposal status actions for emails (https://github.com/UserOfficeProject/issue-tracker/issues/1011).

## How Has This Been Tested

I've only tested this on the STFC side, by checking that emails send to the right people for each status action recipient.


apps/backend/src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -141,21 +141,30 @@ export class SMTPMailService extends MailService {
...
-          name: 'My First Email',
-          description: 'A test message from STFC',
+          id: 'call-created-email',
+          name: 'Call Created Email',
         },
         {
-          id: 'my-second-email',
-          name: 'My Second Email',
-          description: 'A test message from STFC',
+          id: 'clf-proposal-submitted-pi',
```
martin-trajanovski, line 154: I see that you hardcode the templates here. Maybe this is temporary or you want to keep it like this for now if you cannot fetch them from an endpoint.

martin-trajanovski (approved):
It looks fine to me. Just a small comment left

PR #571, opened 2024-04-02
fix: issue to find embellishment questions in the question drawer #1005

bashanlam:
Closes # https://github.com/UserOfficeProject/issue-tracker/issues/1005

## Description
Resolve issue to find embellishment questions in the questionary drawer.  

## Motivation and Context
Resolve issue to find embellishment questions in the questionary drawer.  

## How Has This Been Tested
Manually tested the questionary drawer function. 

## Changes
Update frontend pages to resolve the error.


apps/frontend/src/components/template/QuestionPicker.tsx
```diff
@@ -141,9 +142,15 @@ export const QuestionPicker = (props: QuestionPickerProps) => {
       return true;
     }
 
-    const textMatch = question.question
-      .toLowerCase()
-      .includes(questionFilter.searchText.toLowerCase());
+    const textMatch =
+      question.dataType == 'EMBELLISHMENT'
```
martin-trajanovski, line 146: Maybe it is better to use `DataType.EMBELLISHMENT` here instead of hardcoding it directly

martin-trajanovski (approved):
This looks fine to me. The only concern I have is why the `Embellishment` question is different than others and we need to have separate logic for searching.

simonfernandes (commented):
It looks good but Emma has said she wants to be able to search by question key in the parent issue and this would be an optional extra:

> We could also search the text of embellishments, though Emma has said this won't be necessary so long as we have question key searching.

bashanlam:
> This looks fine to me. The only concern I have is why the `Embellishment` question is different than others and we need to have separate logic for searching.

The embellishment question form lacks a designated question field unlike other question forms, necessitating distinct logic for searching. Reinstating the question field would allow us to utilize the existing search logic. Is there a specific rationale for opting for the 'Plain description' field over the question field?

martin-trajanovski:
> > This looks fine to me. The only concern I have is why the `Embellishment` question is different than others and we need to have separate logic for searching.
> 
> The embellishment question form lacks a designated question field unlike other question forms, necessitating distinct logic for searching. Reinstating the question field would allow us to utilize the existing search logic. Is there a specific rationale for opting for the 'Plain description' field over the question field?

To be honest I am not sure about this. Just wanted to know the reason why do we use different logic only for this field search. I am fine as it is but I guess if it can be improved we should do that and refactor in the future.

martin-trajanovski (approved):

simonfernandes (approved):
This change looks good but as mentioned before Emma was asking to be able to search by question key and not question text. Have you checked she's happy with this?

PR #576, opened 2024-04-05
refactor: remove oauth_access_token column from users table

janosbabik:
## Description
This PR removes the `oauth_access_token` column from the `users` table.

## Motivation and Context
The `oauth_access_token` was previously used for user authentication. However, it is no longer needed and its presence may pose a potential security risk. Therefore, it has been deemed necessary to remove this column from the `users` table.

## Changes
1. A new database patch has been created to drop the `oauth_access_token` column from the `users` table.
2. All references to `oauthAccessToken` in the codebase have been removed.
3. The `logout` method in `OAuthAuthorization` has been simplified since it no longer needs to handle the `oauthAccessToken`.

## Fixes Jira Issue


apps/backend/src/auth/OAuthAuthorization.ts
```diff
@@ -66,26 +60,7 @@ export class OAuthAuthorization extends UserAuthorization {
...
-
-      await OpenIdClient.logout(user.oauthAccessToken);
-
-      return 'logged out';
-    } catch (error) {
-      return rejection('Error ocurred while logging out', {
-        error: (error as Error)?.message,
-      });
-    }
+    return 'logged out';
```
martin-trajanovski, line 63: Is this `logout` function needed at all or we can remove it as well?

martin-trajanovski (approved):
In general it looks good. I just left a small comment that can be checked

TCMeldrum (approved):

PR #578, opened 2024-04-08
feat: update instrument availability during meeting

TCMeldrum:


apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,97 @@
+// import Alert from '@mui/material/Alert';
```
martin-trajanovski, line 1: This comment should be removed I guess

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,97 @@
...
+    <Container component="main" maxWidth="sm">
+      <Formik
+        initialValues={{}}
+        onSubmit={() => {
+          updateTimeAvailable();
+          props.close(+(newTime as string), props.instrument.id);
+        }}
+      >
+        <Form>
+          <Typography
```
martin-trajanovski, line 75: Maybe instead of using custom `Typography` with aditional styles this can be taken out and put into `DialogTitle` component.

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,97 @@
...
+            variant="h6"
+            component="h1"
+            className={classes.cardHeader}
+          >
+            Update {props.instrument.name} Avalibabity Time
+          </Typography>
+          {AvailabilityTimeEditComponent(newTime, SetNewTime)}
+          <Button
```
martin-trajanovski, line 83: The same goes here. Maybe we can avoid style customization as much as possible. If not I think you can just use `sx` property on the button to do some small adjustments.

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,97 @@
...
+        toastSuccessMessage: 'Availability time updated successfully!',
+      }).setInstrumentAvailabilityTime({
+        callId: props.callId,
+        instrumentId: props.instrument.id,
+        availabilityTime: +newTime,
+      });
+  };
+
+  return (
+    <Container component="main" maxWidth="sm">
```
martin-trajanovski, line 66: Instead of using `Container` component here maybe you can use `DialogContent` inside the `Dialog` or maybe you can even move the whole `Dialog` component here and have all in one place.

martin-trajanovski (changes requested):

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
martin-trajanovski: The commented code should be removed

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <Formik
+      initialValues={{}}
+      onSubmit={() => {
+        updateTimeAvailable();
+        props.close(+(newTime as string), props.instrument.id);
+      }}
+    >
+      <Form style={{ padding: '10px' }}>
+        <DialogTitle variant="h6" component="h1">
```
martin-trajanovski, line 69: This looks good to be in the `DialogTitle` component as it is the title of the `Dialog`. But I think it will look more correct maybe if you can bring the whole `Dialog` component in here and wrap the `Formik` with it. Inside you can have `DialogTitle` on the top and `Formik` can sit into `DialogContent`. Maybe this looks a bit better if possible. Here is a good example: https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/frontend/src/components/fap/Proposals/RankInputModal.tsx

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,90 @@
...
+    <Dialog open={props.open} onClose={props.close}>
+      <Formik
+        initialValues={{}}
+        onSubmit={() => {
+          updateTimeAvailable();
+          props.updateTime(+(newTime as string), props.instrument.id);
+        }}
+      >
+        <Form style={{ padding: '10px' }}>
+          <DialogTitle variant="h6" component="h1">
```
martin-trajanovski, line 72: I think you can move the `DialogTitle` outside of the form right after `Dialog` tag and put the `Formik` inside `DialogContent`

apps/frontend/src/components/fap/MeetingComponents/FapMeetingInstrumentsTable.tsx
```diff
@@ -202,6 +224,16 @@ const FapMeetingInstrumentsTable = ({
           debounceInterval: 400,
         }}
       />
+
+      {updateInstrumentTime && selectedCall && (
+        <FapUpdateInstrumentTime
+          open={!!updateInstrumentTime}
```
martin-trajanovski, line 230: I think in this case you might not need to pass the `open` property from outside if you check `updateInstrumentTime` and `selectedCall` already above. Otherwise I think it might be smarter to remove the check above and add something like: `open={!!updateInstrumentTime && !!selectedCall}`

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,89 @@
...
+        callId: props.callId,
+        instrumentId: props.instrument.id,
+        availabilityTime: +newTime,
+      });
+  };
+
+  return (
+    <Dialog open={true} onClose={props.close}>
+      <DialogTitle variant="h6" component="h1">
+        Update {props.instrument.name} Avalibabity Time
```
Scott-James-Hurley, line 64:
```suggestion
        Update {props.instrument.name} Availability Time
```

apps/frontend/src/components/fap/MeetingComponents/FapUpdateInstrumentTime.tsx
```diff
@@ -0,0 +1,89 @@
...
+      }).setInstrumentAvailabilityTime({
+        callId: props.callId,
+        instrumentId: props.instrument.id,
+        availabilityTime: +newTime,
+      });
+  };
+
+  return (
+    <Dialog open={true} onClose={props.close}>
+      <DialogTitle variant="h6" component="h1">
```
Scott-James-Hurley, line 63: I think it would make more sense for the dialog title to be in a header tag numbered higher than the one used for the title of the FAP (`h2`). Ideally the latter would be `h1` and the former `h2`.

Scott-James-Hurley (changes requested):

Scott-James-Hurley (approved):
Ta very much

martin-trajanovski (approved):

PR #580, opened 2024-04-09
feat: add weeks to time allocations on call

TCMeldrum:


Bhaswati1148 (approved):

martin-trajanovski (commented):
This looks fine to me but there is one other place where you should add seconds calculation for the new `Week` unit. It is this file `apps\backend\src\eventHandlers\messageBroker.ts`. Function is called: `getSecondsPerAllocationTimeUnit`

martin-trajanovski (approved):

PR #582, opened 2024-04-10
fix: genericTemplate deletion error

deepaksftc:
## Description

Fixes https://github.com/UserOfficeProject/issue-tracker/issues/906

## Motivation and Context

After deleting an answer/template from a subtemplate question, the state of deleted templates is not getting updated correctly for some scenarios. Because of which we get an error 'Can not delete genericTemplate because of insufficient permissions' when we try to submit the form after adding another answer or going to previous page and coming back. 
Another issue was that, if we have 2 subtemplate questions, after an answer deletion from any of the subtemplate, when we try to add another answer to one of the sub templates, then whole data from second sub template disappears. 

## How Has This Been Tested

Manually

## Changes

Update to not consider deletedtemplates again while reinitializing the form.
apps/frontend/src/components/proposal/ProposalContainer.tsx

Clearing the list of unsaved deletedtemplates while going back to previous page.
apps/frontend/src/models/questionary/useEventHandlers.ts

### DIFF
--- a/apps/frontend/src/components/proposal/ProposalContainer.tsx
+++ b/apps/frontend/src/components/proposal/ProposalContainer.tsx
@@ -72,7 +72,10 @@ export default function ProposalContainer(props: ProposalContainerProps) {
       case GENERIC_TEMPLATE_EVENT.ITEMS_MODIFIED:
         if (action.newItems) {
           if (state.proposal.genericTemplates) {
-            const questionIds = action.newItems.map((value) => value.id);
+            const modifiedTemplates = action.newItems.filter(
+              (value) => !state.deletedTemplates.includes(value.id)
+            );
+            const questionIds = modifiedTemplates.map((value) => value.id);
             draftState.proposal.genericTemplates = [
               ...state.proposal.genericTemplates.filter(
                 (value) =>
@@ -81,17 +84,17 @@ export default function ProposalContainer(props: ProposalContainerProps) {
                       id === value.id || state.deletedTemplates.includes(id)
                   )
               ),
-              ...action.newItems,
+              ...modifiedTemplates,
             ];
             //if new templates have been created add them to the list of created templates
             if (
-              action.newItems.length > state.proposal.genericTemplates.length
+              modifiedTemplates.length > state.proposal.genericTemplates.length
             ) {
               draftState.createdTemplates = [
                 ...state.createdTemplates,
                 ...questionIds.slice(
                   -(
-                    action.newItems.length -
+                    modifiedTemplates.length -
                     state.proposal.genericTemplates.length
                   )
                 ),
--- a/apps/frontend/src/models/questionary/useEventHandlers.ts
+++ b/apps/frontend/src/models/questionary/useEventHandlers.ts
@@ -63,6 +63,7 @@ export default function useEventHandlers(templateGroupId: TemplateGroupId) {
           if (state.isDirty) {
             if (confirmNavigation()) {
               await handleReset();
+              dispatch({ type: 'CLEAR_DELETE_LIST' });
               dispatch({ type: 'GO_STEP_BACK' });
             } else {
               // do nothing


apps/frontend/src/components/proposal/ProposalContainer.tsx
```diff
@@ -72,6 +72,9 @@ export default function ProposalContainer(props: ProposalContainerProps) {
       case GENERIC_TEMPLATE_EVENT.ITEMS_MODIFIED:
         if (action.newItems) {
           if (state.proposal.genericTemplates) {
+            action.newItems = action.newItems.filter(
+              (value) => !state.deletedTemplates.includes(value.id)
+            );
```
mutambaraf, line 77: Can we assign this to a variable and use that instead.

mutambaraf:
We should also try not to delete a none exiting template in the backend .

deepaksftc:
> We should also try not to delete a none exiting template in the backend .

We already have a db call to check if template exists in db at backend. It logs error if the template does not exist.

apps/frontend/src/components/proposal/ProposalContainer.tsx
```diff
@@ -81,28 +85,28 @@ export default function ProposalContainer(props: ProposalContainerProps) {
...
               ];
             }
           } else {
-            draftState.proposal.genericTemplates = action.newItems;
+            draftState.proposal.genericTemplates = modifiedTemplates;
           }
         }
         if (
-          action.newItems?.length !== state.proposal.genericTemplates?.length
+          modifiedTemplates?.length !== state.proposal.genericTemplates?.length
```
mutambaraf, line 109:
```suggestion
          action.newItems?.length !== state.proposal.genericTemplates?.length
```

apps/frontend/src/components/proposal/ProposalContainer.tsx
```diff
@@ -81,28 +85,28 @@ export default function ProposalContainer(props: ProposalContainerProps) {
...
-                    action.newItems.length -
+                    modifiedTemplates.length -
                     state.proposal.genericTemplates.length
                   )
                 ),
               ];
             }
           } else {
-            draftState.proposal.genericTemplates = action.newItems;
+            draftState.proposal.genericTemplates = modifiedTemplates;
```
mutambaraf, line 105:
```suggestion
            draftState.proposal.genericTemplates = action.newItems;
```

apps/frontend/src/components/proposal/ProposalContainer.tsx
```diff
@@ -70,9 +70,13 @@ export default function ProposalContainer(props: ProposalContainerProps) {
         draftState.isDirty = true;
         break;
       case GENERIC_TEMPLATE_EVENT.ITEMS_MODIFIED:
-        if (action.newItems) {
+        let modifiedTemplates = action.newItems;
+        if (modifiedTemplates) {
```
mutambaraf, line 74:
```suggestion
        if (action.newItems) {
```

apps/frontend/src/components/proposal/ProposalContainer.tsx
```diff
@@ -70,9 +70,13 @@ export default function ProposalContainer(props: ProposalContainerProps) {
...
           if (state.proposal.genericTemplates) {
-            const questionIds = action.newItems.map((value) => value.id);
+            modifiedTemplates = modifiedTemplates.filter(
+              (value) => !state.deletedTemplates.includes(value.id)
+            );
+            const questionIds = modifiedTemplates.map((value) => value.id);
```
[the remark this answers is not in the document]
deepaksftc: @mutambaraf, the purpose here is not only to populate the table but also to set the status as dirty if there are deleted templates present. So for the below scenario, we need to check the list without the deletedtemplates here right. In that case, current code is fine right.
       if (
          modifiedTemplates?.length !== state.proposal.genericTemplates?.length
        ) {
          draftState.isDirty = true;
        }

mutambaraf (approved):

Junjiequan (approved):
looks fine to me

PR #586, opened 2024-04-17
feat: add option to combine emails in status actions

simonfernandes:
## Description

This adds an option to combine emails to a recipient when using the status action emails. The default behaviour was previously that, when multiple proposals reach a certain status at the same time, the proposals were grouped by recipient and they would only receive a single email for many proposals. Now the behaviour is that this will only happen when the "combine emails" option is checked and it will otherwise send an individual email for every proposal.

[Previous discussion on Teams](https://teams.microsoft.com/l/message/19:rbRjSbIAOCCDh_gBJLMud7DxrEAnqAQOyjZFEBSm5rU1@thread.tacv2/1711018151883?tenantId=3f66361c-a87e-4158-8f61-99e82db3cac8&groupId=f29c28ff-1d98-4762-bf7f-3c7cbd55ba37&parentMessageId=1711018151883&teamName=User%20Office%20Project%20STFC%2FESS&channelName=General&createdTime=1711018151883)

## Motivation and Context

We have different requirements between STFC and ESS, and it would be nice to have the option.

## How Has This Been Tested

e2e tests for status of checkbox, manual testing that each recipient receives the correct number of emails depending on the option.

## Fixes

Closes UserOfficeProject/issue-tracker#1040


martin-trajanovski (approved):
This looks good to me.

YufanKambang (approved):

PR #587, opened 2024-04-18
BREAKING CHANGE: assigning multiple FAPs per proposal

martin-trajanovski:
## Description
The PR introduces the functionality to allow multiple Facility Access Panels (FAPs) to be associated with a single proposal.

## Motivation and Context
The need for this change arose due to the real-world scenario where a proposal may need to go trough reviews from multiple Facility access panels. This PR allows the backend to handle such scenarios, enhancing the application's flexibility and usability.

## Changes
1. The main change is in the `ProposalDataSource.ts` file where the new column `fap_proposals_id` has been added and the primary key constraint has been redefined.
2. The PostgreSQL version for all docker-compose files has been upgraded to version 16.
3. The view `proposal_table_view` has been modified to accommodate the multiple FAPs per proposal change.
4. The `technical_review` table now includes a new column `technical_review_assignee_id`.
5. The `proposal_events` table has been updated to include a new column `proposal_faps_removed` and the column `proposal_fap_selected` has been renamed to `proposal_faps_selected`.

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-3734](https://jira.esss.lu.se/browse/SWAP-3734)


Junjiequan:
One bug I've found so far. When you remove an instrument from a proposal that was attached to a FAP.  The Fap still remains  assigned status to the proposal/instrument.

apps/backend/src/models/Fap.ts
```diff
@@ -32,7 +32,8 @@ export class FapProposal {
     public fapId: number,
     public dateAssigned: Date,
     public fapTimeAllocation: number | null,
-    public instrumentId: number,
+    public instrumentId: number | null,
+    public callId: number,
```
ACLay, line 36: Is this making FAP's tied to a specific call? Would we have to create new panels for each round?

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -243,62 +244,108 @@ export default class FapMutations {
   @Authorized([Roles.USER_OFFICER])
   async assignProposalsToFapUsingCallInstrument(
     agent: UserWithRole | null,
-    args: AssignProposalsToFapUsingCallInstrumentArgs
+    args: AssignProposalsToFapsUsingCallInstrumentArgs
   ): Promise<boolean | Rejection> {
     return this.assignProposalsToFapUsingCallInstrumentInternal(agent, args);
   }
 
   async assignProposalsToFapUsingCallInstrumentInternal(
```
ACLay, line 252: Should the Fap and Instrument in this name be made plural?

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -243,62 +244,108 @@ export default class FapMutations {
...
-        });
+      if (callHasInstruments.length) {
+        const fapInstruments = callHasInstruments
+          .filter((callHasInstrument) => callHasInstrument.fapId)
+          .map((callHasInstrument) => ({
+            fapId: callHasInstrument.fapId,
+            instrumentId: callHasInstrument.instrumentId,
+          }));
+        if (fapInstruments.length) {
+          await this.assignProposalsToFapsInternal(agent, {
```
ACLay, line 275: Would it be more efficient to move the `for (const callId of callIds) {` loop to only include this call? It seems unnecessary to be making the same `callHasInstruments` call every time round the loop.

apps/backend/src/mutations/FapMutations.ts
```diff
@@ -243,62 +244,108 @@ export default class FapMutations {
...
+          );
 
-    if (result.proposalPks.length !== args.proposals.length) {
+        const fullProposal = await this.proposalDataSource.get(proposalPk);
+
+        if (!fullProposal) {
+          return rejection(`Proposal not found with id: ${proposalPk}`, {
+            args,
+          });
+        }
```
ACLay, line 325: Can this code be moved out of the innermost loop? It looks like it's only dependent on the variable of the outer one.

ACLay (commented):
Only been through the first ~40% of this. Here's my thoughts so far.

apps/backend/src/models/Fap.ts
martin-trajanovski: This is `FapProposal` model and it means that it only connects `FAP` to a `Proposal` and it uses the proposal's call id for some filtering and other needs. So the answer is no to your second question because `FAP` is not directly connected to a call but `FapProposals`

apps/frontend/src/components/fap/Proposals/AssignProposalsToFaps.tsx
```diff
@@ -0,0 +1,161 @@
...
+  });
+
+  const proposalsUniqueInstruments: ProposalViewInstrument[] = getUniqueArrayBy(
+    proposalInstruments.flat(),
+    'id'
+  );
+
+  const allSelectedProposalsHaveSameInstruments = [
+    ...proposalInstruments,
+  ].every(
```
ACLay, line 50: Why is this spreading an array into a new array? Is there a reason we can't just call `proposalInstruments.every()` directly?

apps/frontend/src/components/fap/Proposals/AssignProposalsToFaps.tsx
```diff
@@ -0,0 +1,161 @@
...
+        .toString()
+  );
+
+  const initialValues: Record<string, number | undefined | null> = {};
+
+  proposalsUniqueInstruments.forEach((instrument) => {
+    initialValues[`selectedFapIds_${instrument.id}`] =
+      proposalFapInstruments?.find((i) => i?.instrumentId === instrument.id)
+        ?.fapId || null;
+  });
```
ACLay, line 64: Why does this map to `number | undefined | null`? It looks like all the `?.` operators and the `find()` call should return an ID or `undefined`, but those undefined's are caught and replaced with `null`. It feels like one of these options should be safely removable.

apps/frontend/src/components/fap/Proposals/AssignProposalsToFaps.tsx
```diff
@@ -0,0 +1,161 @@
...
+                assign instruments first and then you can select FAPs
+                accordingly.
+              </Alert>
+            )}
+
+            {proposalFapInstruments?.[0] && !hasEmptyValue(values) && (
+              <Alert severity="warning" data-cy="remove-fap-alert">
+                Be aware that leaving FAP selection empty will remove assigned
+                FAP from proposal/s and delete all FAP reviews on that
+                assignment.
```
ACLay, line 133: I don't like the sound of this. I don't think a user should be able to mass delete reviews after the FAP's have happened.

ACLay (changes requested):
Finished reading through the rest of the changes.

apps/frontend/src/components/fap/Proposals/AssignProposalsToFaps.tsx
martin-trajanovski: This is because the `sort` function inside the `every` mutates the original array and I don't want to do that. It is better to work with the array copy instead of modifying the original array.

martin-trajanovski: But to be more clear I will define it on the top with:
const proposalInstrumentsCopy = [...proposalInstruments];
and then after that use the copy array directly.

apps/frontend/src/components/fap/Proposals/AssignProposalsToFaps.tsx
martin-trajanovski: Yes I undertand you. I presented this to Carina and Emma with a demo and they were fine with it for now. Maybe this will need some modifications in the future but for the initial phase it works. This action is allowed only for user officers and I have few tickets to improve their awareness that they are working with multi-instrument and FAP proposals.

martin-trajanovski:
> One bug I've found so far. When you remove an instrument from a proposal that was attached to a FAP. The Fap still remains assigned status to the proposal/instrument as well as the record in the fap_proposals table.

@Junjiequan this should be fixed now and e2e test is added to cover the bug. You are free to review it again.

Junjiequan (approved):
Looks fine to me

martin-trajanovski:
@ACLay thanks for the great feedback. I have fixed and addressed most of your comments and you are ready to give it another try.

martin-trajanovski:
@TCMeldrum or @Scott-James-Hurley if you have some time one of you can pull this code and test it locally using STFC configuration.

apps/backend/db_patches/0151_MultipleFapPerProposal.sql
```diff
@@ -0,0 +1,129 @@
...
+
+			ALTER TABLE fap_assignments 
+			ADD COLUMN fap_proposal_id INT REFERENCES fap_proposals (fap_proposal_id) ON UPDATE CASCADE ON DELETE CASCADE;
+			UPDATE fap_assignments
+			SET fap_proposal_id = (SELECT fap_proposal_id FROM fap_proposals WHERE fap_proposals.proposal_pk = fap_assignments.proposal_pk AND fap_proposals.fap_id = fap_assignments.fap_id);
+			ALTER TABLE fap_assignments ALTER COLUMN fap_proposal_id SET NOT NULL;
+
+			ALTER TABLE fap_reviews 
+			ADD COLUMN fap_proposal_id INT REFERENCES fap_proposals (fap_proposal_id) ON UPDATE CASCADE ON DELETE CASCADE;
+			UPDATE fap_reviews
```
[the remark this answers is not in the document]
martin-trajanovski: I see your point @TCMeldrum. This is fixed in the next PR that is coming right after the assigning and it is about handling the proposals in FAPs. https://github.com/UserOfficeProject/user-office-core/pull/596/files#diff-8a7411d969f089044d0001a38200f57b55bd1b8d407c6409cb3a2a0244938f89R29.
To be honest I don't have anything specific but maybe when reviewing the next PR related to this then we can test the whole functionality more in details.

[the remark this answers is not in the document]
TCMeldrum: ok sounds good then

TCMeldrum (approved):

martin-trajanovski:
Hi @ACLay can you please check this one again and if you think it looks good I can merge it today.

apps/backend/src/mutations/InstrumentMutations.ts
```diff
@@ -189,10 +189,15 @@ export default class InstrumentMutations {
               );
             })
           );
+          // NOTE: Remove all FAP proposals that are related with the removing instruments
+          await this.fapDataSource.removeProposalsFromFapsByInstrument(
```
ACLay, line 193: I don't follow the logic here. Why as part of `assignProposalsToInstrumentsInternal`, a method that sounds like it should only be doing assignments, are we also removing assignments?

martin-trajanovski: You are right. I was thinking about this while developing the new feature but will try my best to fix it in some better way.

martin-trajanovski: @ACLay I have done some refactoring where I removed that manually cleanup and added better database connections to be able to automatically remove records if proposal is removed from instrument. For example now `fap_proposals` include connection from `instrument_has_proposals` table and if record is removed from `instrument_has_proposals` (proposal is removed from instrument) then proposal is removed from FAP as well automatically in the database level. The same happens with the `technical_review`.

ACLay (approved):

PR #594, opened 2024-04-24
Hotfix 2024 04 23

TCMeldrum:
Hotfix of some cherry-picked bug fixes and the ability to add multiple chairs

### DIFF
--- a/apps/backend/db_patches/0154_MultipleFAPChair.sql
+++ b/apps/backend/db_patches/0154_MultipleFAPChair.sql
@@ -1,7 +1,7 @@
 DO
 $$
 BEGIN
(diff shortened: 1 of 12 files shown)


mutambaraf (approved):

martin-trajanovski (approved):

ACLay (commented):
I think the code looks fine, but should the FAP chairs stuff be marked as a breaking change, since we're renaming, so effectively removing, the old queries & fields? If we need it out then I guess its unavoidable, and it's probably less of an issue than the instrument -> instruments rename since I doubt anything else is using the FAP fields, but I think it should be labelled that way.

TCMeldrum:
I think your right should have flagged it.

PR #597, opened 2024-04-26
fix: make it clearer if you are editing a question or a template

ellen-wright:
## Description
Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1007

## Motivation and Context
Makes it clearer if you are editing a Question or if you are editing a Template with a label on each question type. 

## How Has This Been Tested
Tested manually 

## Changes
A Question appears like this:
image

A Question on a specific Template appears like this:
image


martin-trajanovski (commented):
I am wondering if this cannot be done as some general solution for all of the questions instead of repeating the same label in each of the questions. Like for example in the `QuestionTemplateRelationEditor` and/or `QuestionEditor` components.

apps/frontend/src/components/questionary/questionaryComponents/Boolean/QuestionBooleanForm.tsx
```diff
@@ -25,6 +26,7 @@ export const QuestionBooleanForm = (props: QuestionFormProps) => {
     >
       {() => (
         <>
+          <TemplateEdit pageType="Question" />
```
mutambaraf, line 29: Can we have the naming be clear it is just a label component. Also do we have to hard code this as question or template is the information not being loaded in the parent component.

apps/frontend/src/components/template/QuestionTemplateLabel.tsx
```diff
@@ -0,0 +1,26 @@
...
+        You are editing the question as it appears on the current template
+      </label>
+    );
+  } else if (props.pageType == 'Question') {
+    return (
+      <label className={classes.label}>You are editing the question</label>
+    );
+  } else {
+    return null;
+  }
```
mutambaraf, line 25: Are we expecting another prop instead of question or template .We can also use `===`  if we are comparing strings here.

mutambaraf (approved):

apps/frontend/src/components/template/QuestionTemplateLabel.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const classes = useStyles();
+  if (props.pageType === 'Template') {
+    return (
+      <label className={classes.label}>
+        You are editing the question as it appears on the current template
+      </label>
+    );
+  }
+
+  return <label className={classes.label}>You are editing the question</label>;
```
martin-trajanovski, line 21: Instead of using `makeStyles` we can try to use the `FormLabel` MUI component with `sx` property where we can pass the custom styles that we need. https://mui.com/material-ui/api/form-label/

apps/frontend/src/components/questionary/questionaryComponents/RichTextInput/QuestionRichTextInputForm.tsx
```diff
@@ -27,6 +28,7 @@ export const QuestionRichTextInputForm = (props: QuestionFormProps) => {
     >
       {(formikProps) => (
         <>
+          <TemplateEditLabel pageType="Question" />
```
martin-trajanovski, line 31: Instead of repeating the same label component in all of the forms I was thinking if it is possible to take it out of the form and make it global. Something like putting it in the https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/frontend/src/components/template/QuestionTemplateRelationEditor.tsx before the form itself and that way will be global for all of the questions in one place. We can try the same for https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/frontend/src/components/template/QuestionEditor.tsx

martin-trajanovski (changes requested):

martin-trajanovski (approved):
This looks much better and cleaner now. Great job!

martin-trajanovski:
Just don't forget to do "Squash and merge" because there are quite much commits to be squashed into one it is better.

PR #598, opened 2024-04-26
feat: changes for adding a new question type TIME_REQUESTED

Bhaswati1148:
Creating a new question data type TIME_REQUESTED, similar to instrument picker. This new question type would be used to capture user's input for requesting time allocation on an instrument while submitting a proposal. The response would be saved in proposals table. 
## Motivation and Context
In STFC, during the FAP meetings this requested time will be used as a basis for their decisions. 

## How Has This Been Tested

Manually

## Fixes

https://github.com/UserOfficeProject/issue-tracker/issues/1044


apps/backend/db_patches/0155_AddTimeRequestedDataType.sql
```diff
@@ -0,0 +1,13 @@
+DO
```
TCMeldrum, line 1: Could you combine the patches into one patch?

TCMeldrum (changes requested):
Could we add some e2e testing to check that the time? Maybe in templatesBasics.cy.ts?

Also, it would be useful to have this information accessible in places like the technical review or FAP meeting proposal view, rather than only having it as one of the questions.

Maybe it would make this pr to big, but as the STFC we would like to have this requested time replace the technical review allocated time wherever it is used, but I think that would require discussion with the ESS.

PR #605, opened 2024-05-01
fix: update proposal workflow to include handling submitted proposals after call ended externally

mutambaraf:
## Description
This PR introduces improved handling for submitted proposals after a call has ended eternally. 

## Motivation and Context
This change is required to ensure that the system behaves as expected when processing submitted proposals after their associated call has ended eternally but still open internally. This was identified as a missing feature during our latest system in the last ISIS 2024_2 where proposals submitted when the call was open internal did not have the correct status and where not updated to the next status in the workflow.

## Changes
1. Introduced a new function `publishProposalStatusChange` to publish the proposal status change. 
2. Added a new function `handleSubmittedProposalsAfterCallEnded` to handle the case where a proposal is submitted after the call ends.
3. Refactored `handleWorkflowEngineChange` to use the new functions and handle the case where a proposal is submitted after the call ends.

## How Has This Been Tested?
Cypress tests.

## Fixes UOP Issue

Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1065


yoganandaness:
@mutambaraf Should we have e2e tests for this workflow? I would prefer to have one.

mutambaraf:
> @mutambaraf Should we have e2e tests for this workflow? I would prefer to have one.

I have added a test for the workflow.

yoganandaness (approved):

Scott-James-Hurley (approved):

PR #606, opened 2024-05-02
feat: add endpoint to retry sending emails for proposal events

mutambaraf:
## Description
This PR introduces a new endpoint to reattempt sending emails for proposal events. 

## Motivation and Context
The change is required to ensure the system can retry sending emails if the initial attempt fails. It enhances the reliability of the system's communication and ensures that the appropriate users are notified of proposal events.

## Changes
1. Added a new event handler that logs a warning when an email send request is not implemented by the email handler.
2. Added a new test case to ensure that the user officer should not be able to send proposal submitted email to a proposal not submitted.
3. Modified the ProposalMutations class to include the ability to send event emails.
4. Created a new mutation to handle the sending of proposal event emails.

## Fixes UOP Issue

Closes: https://github.com/UserOfficeProject/issue-tracker/issues/630

## Depends On
https://github.com/isisbusapps/internal-scripts/pull/124


apps/backend/src/mutations/ProposalMutations.ts
```diff
@@ -752,4 +757,34 @@ export default class ProposalMutations {
 
     return submitted;
   }
+
+  @Authorized([Roles.USER_OFFICER])
+  async sendEventEmail(
+    agent: UserWithRole | null,
+    args: SendProposalEventEmailArgs
+  ) {
+    if (args.event !== Event.PROPOSAL_SUBMITTED) {
```
martin-trajanovski, line 766: As I see the endpoint is called `sendProposalEventEmail` but for now it only accept `Event.PROPOSAL_SUBMITTED`. Is this meant to be extended in the future or it is only used to resend emails for `PROPOSAL_SUBMITTED` event.

apps/backend/src/mutations/ProposalMutations.ts
```diff
@@ -752,4 +757,34 @@ export default class ProposalMutations {
...
+    if (!proposal) {
+      return rejection('Proposal not found', { args });
+    }
+
+    if (!proposal.submitted) {
+      return rejection('Proposal has not been submitted', { args });
+    }
+
+    return this.emailHandler({
+      type: Event.PROPOSAL_SUBMITTED,
```
martin-trajanovski, line 783: Shouldn't we use `type: args.event` here instead?

apps/backend/src/mutations/ProposalMutations.ts
mutambaraf: Wanted to limit it at the moment but let me extend it to include other emails.

mutambaraf: Working on extending it in this pr.

apps/backend/src/mutations/ProposalMutations.ts
mutambaraf: I have fixed this.

Scott-James-Hurley (commented):
I think there should be tests for successfully using the endpoint and non-user officers not being able to use it.

bolmsten:
I am unsure about this PR. 

I would suggest creating a new table for tracking status_actions that is visible for the user officer. Every time a status action runs we would log it into the table with the status of success/fail. If the action has a fail the user officer can decide to re-run it via the UI. 

This has several benefits. It is more generic and will solve problems for all status actions. It gives a overview for the user officer of both success and failures.

bolmsten (changes requested):
See comment

mutambaraf:
> I am unsure about this PR.
> 
> I would suggest creating a new table for tracking status_actions that is visible for the user officer. Every time a status action runs we would log it into the table with the status of success/fail. If the action has a fail the user officer can decide to re-run it via the UI.
> 
> This has several benefits. It is more generic and will solve problems for all status actions. It gives a overview for the user officer of both success and failures.

@bolmsten Its possible for me to use the event log table we just add a two or three columns depending if we what to capture the handler.

bolmsten:
> > I am unsure about this PR.
> > I would suggest creating a new table for tracking status_actions that is visible for the user officer. Every time a status action runs we would log it into the table with the status of success/fail. If the action has a fail the user officer can decide to re-run it via the UI.
> > This has several benefits. It is more generic and will solve problems for all status actions. It gives a overview for the user officer of both success and failures.
> 
> @bolmsten Its possible for me to use the event log table we just add a two or three columns depending if we what to capture the handler.I have included only backend changes .If l can get a comment will be great.

I think it will be hard as you are mixing events and proposal statuses.  We should not send emails based on events anymore, instead we should move over to the new way of doing it via the proposal workflow engine. This makes the email sending process dynamic and configurable. So create a new table that track proposal status actions, where one can replay failed actions.

@martin-trajanovski  comment?

martin-trajanovski:
> > > I am unsure about this PR.
> > > I would suggest creating a new table for tracking status_actions that is visible for the user officer. Every time a status action runs we would log it into the table with the status of success/fail. If the action has a fail the user officer can decide to re-run it via the UI.
> > > This has several benefits. It is more generic and will solve problems for all status actions. It gives a overview for the user officer of both success and failures.
> > 
> > 
> > @bolmsten Its possible for me to use the event log table we just add a two or three columns depending if we what to capture the handler.I have included only backend changes .If l can get a comment will be great.
> 
> I think it will be hard as you are mixing events and proposal statuses. We should not send emails based on events anymore, instead we should move over to the new way of doing it via the proposal workflow engine. This makes the email sending process dynamic and configurable. So create a new table that track proposal status actions, where one can replay failed actions.
> 
> @martin-trajanovski comment?

Yes I agree with this. Maybe it is better to handle this differently. To me it sounds like a more general solution which would be perfect to aim for. Event or status specific handlers are something that we don't want to introduce as long as it is something very important and cannot be done otherwise. So @mutambaraf try to think little bit more about this and if it is needed create another discussion or open issue where you can investigate more and find a nice and clean solution to the problem. Would be perfect to have some mechanism of replaying status actions as a user officer.

mutambaraf:
Will open another PR

PR #608, opened 2024-05-07
fix: prevent multiple save clicks

simonfernandes:
## Description

Fixes being able to click the "save" button multiple times. You can accidentally double click it or intentionally click it several times. When you do, it will generate a new database row and ID and fire updated and created events each time. I also suspect it's causing files to scan multiple times (the linked issue) which is how this was found.

## How Has This Been Tested

Manual testing that the button can't be clicked multiple times, and re-enables when there are validation errors.

## Fixes

Closes UserOfficeProject/issue-tracker#934


ACLay (approved):

martin-trajanovski (commented):
This looks good to me but would be nice if we can add an e2e test to just make sure this doesn't appear again in the future.

ACLay (approved):

martin-trajanovski (approved):

PR #609, opened 2024-05-10
feat: add User Office email recipient

simonfernandes:
## Description
Adds a new recipient option in the email status actions for "User Office". When it is selected it will send an email to the address(es) defined in the `USER_OFFICE_EMAIL` env var.

## Motivation and Context

STFC sometimes want to send the User Office emails for certain things. Because it is a common recipient (and for us is actually two email addresses) it makes sense for it to be a predefined recipient.

## How Has This Been Tested

Manual testing that the email sends to the new recipient

## Fixes

Part of UserOfficeProject/issue-tracker#1039

## Depends on
Requires https://github.com/isisbusapps/docker-orchestration/pull/559 (updating the environment variable name)


apps/backend/example.development.env
```diff
@@ -50,6 +50,7 @@ JWT_SECRET=qMyLZALzs229ybdQXNyzYRdju7X784TH
 TZ=Europe/Stockholm
 DATE_FORMAT=dd-MM-yyyy
 DATE_TIME_FORMAT=dd-MM-yyyy HH:mm
+USER_OFFICE_EMAIL=useroffice1@email.com; useroffice2@email.com
```
mutambaraf, line 53: Is it possible for us to make this an array.

simonfernandes: I don't think so, it seems docker-compose and kubernetes only allows string values. You can separate them with a comma or some other delimiter and then parse in the code, but that's not much different to this. The benefit of using `;` here is that you can directly use the entire string without parsing in the to address of an email and it will send to all the addresses listed.

martin-trajanovski: I just need better clarification if this is user office email or user officer email.

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -88,6 +89,55 @@ export const emailActionHandler = async (
           break;
         }
 
+        case EmailStatusActionRecipients.USER_OFFICE: {
+          const userOfficeEmail = process.env.USER_OFFICE_EMAIL;
+
+          if (userOfficeEmail == null) {
```
martin-trajanovski, line 95: I think it is better to use `===` triple equal when doing this comparison and another question is what happens if `userOfficeEmail` is `undefined`.  Is there any specific reason why we compare it only against `null` and not using something like: `if (!userOfficeEmail) {`

apps/backend/example.development.env
simonfernandes: This is the central user office email. They wouldn't have an account because it doesn't belong to an individual. Also in our case we would additionally send these emails to User Officer Emma as requested. But we wouldn't want all User Officers to automatically be included here. Just thought I'd write down what we just discussed in the meeting 🙂

bolmsten:
As discussed we would like this to be in the DB so that it can be changed in the App settings page

apps/backend/src/statusActionEngine/emailActionHandler.ts
simonfernandes: It would have checked for null and undefined with `==` due to type coercion. `===` seems a bit more verbose as we have to explicitly check both like `(userOfficeEmail === null || userOfficeEmail === undefined)`. But using `(!userOfficeEmail)` as you said sounds best as it will then additionally check for falsy values like an empty string.

martin-trajanovski (approved):
Now it seems good to me

mutambaraf (approved):

PR #610, opened 2024-05-13
feat: added re-login button on OAuth logout page

Junjiequan:
## Description

This PR adds functionality to the PingFederate idp/startSLO.ping logout template that allows the conditional display of a re-login button, which, when clicked, redirects users to a specified URL. This feature is controlled by the AUTH_RELOGIN_PARAMS environment variable.

Additionally, it removes the redundant updateOIDCSettings() function, which duplicated logic found in the `init` function of the `OAuthAuthorization` class. The original PR that added upateOIDCSettings() is https://github.com/UserOfficeProject/user-office-core/pull/191

## Motivation and Context

Enhances user experience by providing a straightforward option to re-authenticate after logging out, but only when specific conditions are met.

## How Has This Been Tested

Manual testing

## Changes

Logout Process Enhancement:  Introduced the logoutWithRedirectParam method to manage the logout URL based on environment settings. This method checks if AUTH_RELOGIN_PARAMS is provided and correctly formatted (from=REDIRECT_URL). If valid, the re-login button is enabled and the URL is appended to the logout process.

Code Cleanup: Removed the updateOIDCSettings() function as its functionality overlaps with that of the existing init method in the OAuthAuthorization class, streamlining the codebase


martin-trajanovski (commented):
Looks fine to me but I think you should add this `AUTH_RELOGIN_PARAMS` variable to the example env file for the backend and add some example value. You can also add a comment and explain little bit why and how it is used.

martin-trajanovski (commented):
I am happy to approve this but was thinking if there is possibility for testing it with an e2e test or it is not possible in the local development environment.

martin-trajanovski (approved):

PR #611, opened 2024-05-23
feat: add call filter option for short code

Scott-James-Hurley:
Closes: UserOfficeProject/issue-tracker/issues/1077

## Description

Adds ability to fetch calls by their short code.

## Motivation and Context

This would be useful for one of our downstream tools and is a good improvement in general imo. 

## How Has This Been Tested

Tested manually

## Changes

- `apps/backend/src/datasources/postgres/CallDataSource.ts`: filter calls by short code
- `apps/backend/src/resolvers/queries/CallsQuery.ts`: add short code filter option


apps/backend/src/datasources/postgres/CallDataSource.ts
```diff
@@ -44,6 +44,11 @@ export default class PostgresCallDataSource implements CallDataSource {
 
   async getCalls(filter?: CallsFilter): Promise<Call[]> {
     const query = database('call').select(['*']);
+
+    if (filter?.shortCode) {
+      query.where('call_short_code', filter.shortCode);
+    }
```
mutambaraf, line 50: Do we need it to be exact or we can benefit from making match with a like.

Scott-James-Hurley: Great point, thanks.

mutambaraf (approved):

EdwardHaynes (approved):

martin-trajanovski (approved):

PR #618, opened 2024-05-31
feat: make already generated proposal pdfs of migrated proposals available for download

deepaksftc:
## Description

Proposals migrated from old system to user-office has already generated proposal pdfs available for them with a different template. This change is to store the proposal pdfs of these older proposals in the database and to make it available for users and user officer to download. Proposal pdfs are stored Files table. During download pdf flow, it will check whether its a migrated proposal, and if yes, it will fetch and download the saved pdf from database. This feature will not be applicable for multi selected proposals pdf download.
Data will be updated in db by running script. (https://github.com/isisbusapps/internal-scripts/pull/126)

## How Has This Been Tested

Manually

## Fixes

Fixes https://github.com/UserOfficeProject/issue-tracker/issues/758


janosbabik:
We discussed your PR, and our suggestion is to store the old PDF files in file storage, for example. Using dependency injection, you can customize the factory middleware to check if there is an old file for a proposal. If there is, provide it instead of using 'legacy' fields in the database.

What do you think?

deepaksftc:
> We discussed your PR, and our suggestion is to store the old PDF files in file storage, for example. Using dependency injection, you can customize the factory middleware to check if there is an old file for a proposal. If there is, provide it instead of using 'legacy' fields in the database.
> 
> What do you think?

We had a discussion regarding your suggestion. We have a need for having an identifier for imported proposals for upcoming features also. If altering the existing tables is a concern, we can go ahead with creating a new table for storing the details of imported proposals alone if that's ok? All the naming's will be changed to 'imported/migrated' instead of using 'legacy'.

deepaksftc:
> When you create a migrated proposal, you can generate a special proposal ID based on the old one, such as 1234 -> MIGRATED_1234. Then, in your StfcProposalDataSource, you can implement your own cloneProposal function. This function should check if the proposal ID matches MIGRATED_*. If it does, the function can throw an error. During the PDF downloading process, the callFactoryService can be defined differently in your case (DI) to use logic that checks if the proposal is migrated. If it is, the proposal pdf can be downloaded from a storage location or any other source, for example. If you want to keep the ID, you can use the creation time to check if it is an old proposal.

Since most of the old proposals are already migrated, I am not sure if changing the proposal id for those proposals will be ideal. We already use them for reports. Check with creation time will also have some challenges I guess, as there are still proposals waiting to be migrated, like the Xpress one. If adding a new column cannot be done, I guess, we will make use the description field in call table. In that case, we will be having description comparison checks at couple of places to identify if its a imported proposal. Hope that is fine? 
Regarding storing the pdfs, I will once check the possibility of storing the pdfs within our files table before I plan to create a new schema?

apps/backend/src/factory/EssDownloadService.ts
```diff
@@ -0,0 +1,107 @@
...
+  DownloadType,
+  PDFType,
+  XLSXType,
+  ZIPType,
+  DownloadService,
+} from './DownloadService';
+
+const ENDPOINT = process.env.USER_OFFICE_FACTORY_ENDPOINT;
+
+export class EssDownloadService implements DownloadService {
```
janosbabik, line 21: I think we can call it DefaultDownloadService instead of EssDownloadService, as it is not an ESS-specific service.

apps/backend/src/factory/StfcDownloadService.ts
```diff
@@ -0,0 +1,149 @@
...
+        await new Promise((f) => setTimeout(f, 1000));
+        proposalPdfData.emit('end');
+      } catch (error) {
+        next({
+          error,
+          message: 'Could not download generated',
+        });
+      }
+    } else {
+      const controller = new AbortController();
```
janosbabik, line 79: From here, it is identical to the code in the EssDownloadService. Can we create a separate function like fetchDataAndStreamResponse(req, res, next, downloadType, type, properties) or something similar , and use it both here and there?

janosbabik (changes requested):
I left some comments, however, I like this solution. Thank you for your effort.

janosbabik (approved):

apps/backend/src/factory/StfcDownloadService.ts
```diff
@@ -0,0 +1,82 @@
...
+        `Migrated-${data.proposal.proposalId}.pdf`
+      );
+
+      if (proposalPdfData) {
+        isPdfAvailable = true;
+        fileName = `${data.proposal.proposalId}_${
+          data.principalInvestigator.lastname
+        }_${data.proposal.created.getUTCFullYear()}.pdf`;
+      }
+    }
```
mutambaraf, line 54: I am not seeing away we are handling multiple proposals and are we not going to allow zip for legacy proposals.

apps/backend/src/datasources/mockups/FileDataSource.ts
```diff
@@ -44,4 +46,8 @@ export default class FileDataSourceMock implements FileDataSource {
       new Date()
     );
   }
+
+  async getBlobdata(fileName: string): Promise<ReadStream | null> {
+    return null;
+  }
```
mutambaraf, line 52: Are we mocking only returning null or  if filename if provided we must return information.

apps/backend/src/factory/StfcDownloadService.ts
deepaksftc: Yes, currently its done only for single proposal pdf download. This was communicated with Emma.

apps/backend/src/datasources/mockups/FileDataSource.ts
deepaksftc: This will be the default case when the data for that particular proposal id is not present in db right. In that case, it will return null. If needed, I can change this to return a stream.

apps/backend/src/factory/StfcDownloadService.ts
mutambaraf: What happens if l send two proposals which are legacy and want zip

deepaksftc: It will follow our usual pdf download using pdf template one.  Focused more on users. User doesn't have the option for multi download I believe.

mutambaraf: This is not facing users only even the API can make a request to this end point.

mutambaraf: If we are not handling it we can just return a message for the  areas not implemented.

apps/backend/src/datasources/mockups/FileDataSource.ts
mutambaraf: It must potentially return a ReadStream from its signature.

apps/backend/src/factory/StfcDownloadService.ts
deepaksftc: ok sure. I have a doubt, since we are not introducing any new endpoint and for current scenarios, no scenario will be breaking, then is it suitable to return a message that its not implemented. For all other combinations, it will generate pdf normally using templates.

apps/backend/src/factory/StfcDownloadService.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+    } else {
+      fetchDataAndStreamResponse(
+        downloadType,
+        type,
+        properties,
+        req,
+        res,
+        next
+      );
```
mutambaraf, line 79: Do we need this else looks like its only two options.

apps/backend/src/factory/StfcDownloadService.ts
mutambaraf: Maybe we should open an issue for it if it not already open, because my understanding was that we can generate a proposal or get it as a legacy proposal from the database.

apps/backend/src/factory/StfcDownloadService.ts
```diff
@@ -0,0 +1,75 @@
...
+        res.setHeader('x-download-filename', querystring.escape(fileName));
+        res.setHeader('content-type', 'application/pdf');
+
+        proposalPdfData.pipe(res);
+        await new Promise((f) => setTimeout(f, 1000));
+        proposalPdfData.emit('end');
+      } catch (error) {
+        next({
+          error,
+          message: 'Could not download generated',
```
mutambaraf, line 68: Can this message be improved like "Could not generate proposal pdf"

mutambaraf (approved):

apps/backend/src/factory/StfcDownloadService.ts
```diff
@@ -0,0 +1,149 @@
...
+    let isPdfAvailable = false;
+    let proposalPdfData = null;
+    const data = properties?.data[0] as ProposalPDFData;
+
+    if (properties?.data?.length === 1) {
+      const fileDataSource = container.resolve<FileDataSource>(
+        Tokens.FileDataSource
+      );
+      proposalPdfData = await fileDataSource.getBlobdata(
+        `Migrated-${data.proposal.proposalId}.pdf`
```
simonfernandes, line 45: Sorry to only realise this now, but I think unfortunately there needs to be a way of fetching the reference number for a specific facility. Because there are some reference numbers that are shared between facilities. For example, 91012 is a reference number for both LSF and HPL.

deepaksftc: I have done the changes to save the pdf with facility name included in file name. Please have a look.

apps/backend/src/factory/StfcDownloadService.ts
deepaksftc: I think, since this method is a void method and we are not returning anything back, the method outside if case is also getting called and its logging error. So, I am reverting the change done as part of this comment.

simonfernandes (approved):

apps/backend/src/factory/StfcDownloadService.ts
mutambaraf: I thinking for it to terminate in the if statement we either have to send back a response or execute the next function otherwise it will continue executing the next statements after the if statement or we can just call return;

simonfernandes (changes requested):
Sorry again for realising late, but there is another issue with our reference numbers which is that they can clash with the randomly generated ones that are assigned on pre-submission. So there is a risk that a user could click download PDF on a draft and get another user's legacy proposal. 

I think we could get around this with a check to only download via the blob route if the proposal is submitted. Or by checking another identifier on the proposal (if there is one)?

simonfernandes (approved):

PR #619, opened 2024-05-31
fix: wait for database patches to be applied before setting variables

bolmsten:


janosbabik (approved):

martin-trajanovski:
@Scott-James-Hurley can you please add your suggestion here on how to fix the issue with massAssignFapReviews?

Scott-James-Hurley:
> @Scott-James-Hurley can you please add your suggestion here on how to fix the issue with massAssignFapReviews?

It turns out the mass assigning FAP reviews functionality I added was not what our User Officer was looking for. I'm working on a [PR](https://github.com/UserOfficeProject/user-office-core/pull/624) right now that amends it; this includes removing the tests that are failing for this PR. It might be an idea to just remove those tests here, unless you think it's worth investigating why they're failing.

PR #621, opened 2024-06-03
fix: only submitted proposals in fap

TCMeldrum:
Hotfix so only submitted proposals appear in FAP export

We had a support ticket in where proposals that were expired or in draft were showing up in the FAP data exports. This change makes it so only submitted proposals show up there.

### DIFF
--- a/apps/backend/src/datasources/postgres/FapDataSource.ts
+++ b/apps/backend/src/datasources/postgres/FapDataSource.ts
@@ -471,7 +471,8 @@ export default class PostgresFapDataSource implements FapDataSource {
         'p.status_id': 'ps.proposal_status_id',
       })
       .where('sp.fap_id', fapId)
-      .andWhere('ihp.instrument_id', instrumentId);
+      .andWhere('ihp.instrument_id', instrumentId)
+      .andWhere('p.submitted', true);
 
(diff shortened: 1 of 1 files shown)


simonfernandes (approved):

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -471,7 +471,8 @@ export default class PostgresFapDataSource implements FapDataSource {
         'p.status_id': 'ps.proposal_status_id',
       })
       .where('sp.fap_id', fapId)
-      .andWhere('ihp.instrument_id', instrumentId);
+      .andWhere('ihp.instrument_id', instrumentId)
+      .andWhere('p.submitted', true);
```
ACLay, line 475: I think in the teams thread you mentioned it was pulling in draft proposals and expired ones. I can see how this'll exclude any draft proposals, but is this enough, for current needs at least? Can something be submitted and then marked as expired? Or does expiring unset any submitted flag?

Also, could you put the details of why this needed into an issue so it's easier to look back on why we wanted the change made.

ACLay (approved):

martin-trajanovski (approved):

PR #627, opened 2024-06-13
feat: improve user officer proposals table sorting

martin-trajanovski:


apps/e2e/cypress/e2e/proposalAdministration.cy.ts
```diff
@@ -29,7 +30,13 @@ context('Proposal administration tests', () => {
   let createdInstrumentId: number;
 
   const instrument1 = {
-    name: faker.random.words(2),
+    name: `0000. ${faker.random.words(2)}`,
```
simonfernandes, line 33: Might be worth using `faker.string.alpha()` and `faker.string.alphanumeric()` instead? But I guess we may just swap all of the deprecated faker functions at once in the future.

simonfernandes (commented):
This looks good. I have a question from the PR description:

> Removed the sorting restrictions on "Principal Investigator" and "PI Email" columns in the proposals table.

It doesn't look like this has been affected? But I think maybe the bot made this comment and got it wrong or maybe you have changed the files since the description was made?

apps/e2e/cypress/e2e/proposalAdministration.cy.ts
martin-trajanovski: Yes I see that it is deprecated but we can do the swap at once in one single PR.

martin-trajanovski:
> This looks good. I have a question from the PR description:
> 
> > Removed the sorting restrictions on "Principal Investigator" and "PI Email" columns in the proposals table.
> 
> It doesn't look like this has been affected? But I think maybe the bot made this comment and got it wrong or maybe you have changed the files since the description was made?

Yeah you are right. This was the bot because I had some merge leftovers that I cleaned up after I opened the PR. I removed it from the description now as it is not touching that part.

simonfernandes (approved):

Junjiequan (approved):

PR #628, opened 2024-06-13
feat: technique creation

simonfernandes:
## Description

This PR creates "Techniques" which are accessible via a new tab in the User Officer role. You can create/edit/delete them and add a name, short code, description and attach instruments (experimental areas) to them.

Note that this is not hidden behind a feature flag so it is currently visible to all facilities.

## Motivation and Context

This is the first part of creating an STFC Xpress system in the software. You can see our plan so far and more discussion in https://github.com/UserOfficeProject/issue-tracker/discussions/1032#discussioncomment-9752530

## How Has This Been Tested

- General manual testing
- e2e tests
- Backend unit tests
- Manual graphql endpoint testing for schema validation

## Fixes

Closes https://github.com/UserOfficeProject/issue-tracker/issues/1095

## Changes

- Frontend pages for techniques
- Backend resolver/query/mutation/datasource flows for the various create/edit/update/delete functionality

## Depends on

The schema validation has already been merged via https://github.com/UserOfficeProject/user-office-lib/pull/194

https://github.com/UserOfficeProject/user-office-core/assets/149392207/ca90ff55-e257-459a-a858-56768a525073


apps/backend/db_patches/0157_AddTechniques.sql
```diff
@@ -0,0 +1,24 @@
+DO
+$$
+BEGIN
+  IF register_patch('0157_AddTechniques.sql', 'Deepak Jaison & Simon Fernandes', 'Add techniques', '2024-06-05') THEN
+    BEGIN
+
+      CREATE TABLE techniques (
```
yoganandaness, line 7:
```suggestion
      CREATE TABLE IF NOT EXISTS techniques (
```

apps/backend/db_patches/0157_AddTechniques.sql
```diff
@@ -0,0 +1,24 @@
...
+        technique_id SERIAL PRIMARY KEY,
+        name VARCHAR(200) NOT NULL,
+        short_code VARCHAR(20) NOT NULL,
```
yoganandaness, line 10: This fields sounds like UNIQUE to me. But could you please check it out once?

apps/backend/src/models/Technique.ts
```diff
@@ -0,0 +1,15 @@
+export class Technique {
+  constructor(
+    public id: number,
```
yoganandaness, line 3: Should it be `techniqueId` as per the db definition? 

https://github.com/UserOfficeProject/user-office-core/pull/628/files#diff-9f73e29295089d202fe5fa3d11befff47df1c37abc181ce3ded2a113b85d7706R8

apps/backend/src/datasources/TechniqueDataSource.ts
```diff
@@ -0,0 +1,23 @@
+import { Instrument } from '../models/Instrument';
+import { Technique } from '../models/Technique';
+import { CreateTechniqueArgs } from '../resolvers/mutations/CreateTechniqueMutation';
+
+export interface TechniqueDataSource {
+  create(args: CreateTechniqueArgs): Promise<Technique>;
+  getTechnique(instrumentId: number): Promise<Technique | null>;
```
yoganandaness, line 7:
```suggestion
  getTechnique(techniqueId: number): Promise<Technique | null>;
```

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
```diff
@@ -0,0 +1,159 @@
...
+  TechniqueHasInstrumentsRecord,
+  TechniqueRecord,
+  createInstrumentObject,
+} from './records';
+
+@injectable()
+export default class PostgresTechniqueDataSource
+  implements TechniqueDataSource
+{
+  createTechniqueObject(technique: TechniqueRecord): any {
```
yoganandaness, line 20: Is it possible to avoid `any` here?

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
```diff
@@ -0,0 +1,159 @@
...
+      .select()
+      .from('technique_has_instruments')
+      .where('technique_id', techniqueId)
+      .then((results: TechniqueHasInstrumentsRecord[]) =>
+        results.map((result) => {
+          return Int.parseValue(result.instrument_id);
+        })
+      );
+
+    return database
```
yoganandaness, line 93: I believe, We can do `Join` to merge these two queries into one query.

apps/backend/src/mutations/TechniqueMutations.ts
```diff
@@ -0,0 +1,125 @@
...
+@injectable()
+export default class TechniqueMutations {
+  constructor(
+    @inject(Tokens.TechniqueDataSource)
+    private dataSource: TechniqueDataSource,
+    @inject(Tokens.UserAuthorization) private userAuth: UserAuthorization
+  ) {}
+
+  @EventBus(Event.TECHNIQUE_CREATED)
+  @ValidateArgs(createTechniqueValidationSchema)
```
yoganandaness, line 33: I just went through the schema in the user-office-lib. I felt it could be improved a little bit. 

https://github.com/UserOfficeProject/user-office-lib/blob/master/packages/validation/src/Technique/index.ts#L5

For ex., The Short Code in DB has the length restriction of 20 as defined in the migration. However in the validation, no such condition is present. I think it is a good idea to align the DB level validation, backend Level validation and frontend level on par with each other.

apps/backend/src/mutations/TechniqueMutations.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+import {
+  dummyUserOfficerWithRole,
+  dummyUserWithRole,
+} from '../datasources/mockups/UserDataSource';
+import TechniqueMutations from './TechniqueMutations';
+
+const techniqueMutations = container.resolve(TechniqueMutations);
+
+describe('Test technique Mutations', () => {
+  test('A user cannot create a technique', () => {
```
yoganandaness, line 14: May be, we can have an additional test to check the ShortCode(more than 20 characters).

yoganandaness: Also Technique Name if possible.

apps/backend/src/mutations/TechniqueMutations.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  test('A user cannot remove an assigned instrument from a technique', async () => {
+    return expect(
+      techniqueMutations.removeInstrumentFromTechnique(dummyUserWithRole, {
+        instrumentId: 1,
+        techniqueId: 1,
+      })
+    ).resolves.toHaveProperty('reason', 'INSUFFICIENT_PERMISSIONS');
+  });
+});
```
yoganandaness, line 117: Can we also add an additional test to test the Unique key constraint in InstrumentTechniques table?

yoganandaness: ex., Trying to add the same InstrumentId and techniqueId combination.

apps/backend/src/resolvers/mutations/AssignInstrumentsToTechnique.ts
```diff
@@ -0,0 +1,34 @@
...
+export class AssignInstrumentsToTechniqueArgs {
+  @Field(() => [Int])
+  public instrumentIds: number[];
+
+  @Field(() => Int)
+  public techniqueId: number;
+}
+
+@Resolver()
+export class AssignScientsitsToInstrumentMutation {
```
yoganandaness, line 23: The Class name is conflicting with the purpose of this Mutation Resolver. Could you please re-check once?

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const techniqueId = selectedTechniques[0].id;
+    if (instruments?.length) {
+      await api({
+        toastSuccessMessage: `Instrument/s assigned to the selected technique successfully!`,
+      }).assignInstrumentsToTechnique({
+        instrumentIds: instruments.map((instrument) => instrument.id),
+        techniqueId,
+      });
+    }
+    setTechniques((techniques) =>
```
yoganandaness, line 97: What happens when the api fails. Are we still attaching the Instrument to the Technique in the frontend?

yoganandaness: I think it is a good ideal to move this(setTechniques) into api().then

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const techniqueList: TechniqueFragment[] = [];
+    techniqueList.push(technique);
+    setSelectedTechniques(techniqueList);
+    setOpenTechniqueAssignment(true);
+  }
+
+  const assignInstrumentsToTechniques = async (
+    instruments: InstrumentFragment[]
+  ): Promise<void> => {
+    const techniqueId = selectedTechniques[0].id;
```
yoganandaness, line 88: Kindly help me understand this. 

Are we going to attach the instruments only to the first selected technique? That means, the remaining techniques will not be attached to the instruments?

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+        }
+      })
+    );
+  };
+
+  const AssignedInstruments = React.useCallback(
+    ({ rowData }) => {
+      return <AssignedInstrumentsTable technique={rowData} />;
+    },
+    [setTechniques]
```
yoganandaness, line 115: setTechniques is a function and is never going to get changed after the first render. Is it ok to have it as a dependency for a useCallback?

I think rowData is the right candidate for the dependency.

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+  );
+
+  const removeIntrumentsFromTechnique = async (
```
yoganandaness, line 118: Since this follows the same pattern as `assignInstrumentsToTechniques`. Kindly check the comments attach to the `assignInstrumentsToTechniques`.

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+      close={(technique: TechniqueFragment | null) =>
+        !!editTechnique ? onUpdate(technique) : onCreate(technique)
+      }
+    />
+  );
+
+  return (
+    <>
+      <Dialog
+        aria-labelledby="simple-modal-title"
```
yoganandaness, line 164: Could we have the appropriate title instead of default placeholders?

apps/frontend/src/components/technique/CreateUpdateTechnique.tsx
```diff
@@ -0,0 +1,141 @@
...
+            {isExecutingCall && <UOLoader size={14} />}
+            {technique ? 'Update' : 'Create'}
+          </Button>
+        </Form>
+      )}
+    </Formik>
+  );
+};
+
+CreateUpdateTechnique.propTypes = {
```
yoganandaness, line 131: I feel this is not really required, since you have already enforced the Prop rules via typescript. `CreateUpdateTechniqueProps`.

yoganandaness: Correct me if i am wrong.

apps/frontend/src/components/technique/AssignInstrumentsToTechniques.tsx
```diff
@@ -0,0 +1,128 @@
...
+                  (selectedInstrumentId) => selectedInstrumentId.id === id
+                )
+              ) {
+                deletedInstruments.push(id);
+              }
+            });
+
+            await assignInstrumentsToTechniques(addedInstruments || null);
+
+            await removeIntrumentsFromTechnique(deletedInstruments);
```
yoganandaness, line 73: Should we do a IF to check if `deletedInstruments` is `[]` or not, before calling `removeIntrumentsFromTechnique`.

yoganandaness (changes requested):
Kindly go through my comments and reply if have any concerns. Overall it is a good work and a complete package.

apps/backend/src/mutations/TechniqueMutations.ts
deepaksftc: Added size restrictions for name and short code. Created a new PR https://github.com/UserOfficeProject/user-office-lib/pull/195

apps/backend/src/models/Technique.ts
simonfernandes: We tried this and unfortunately couldn't get it working with the `SuperMaterialTable` component. It expects an `id` property in the `TechniqueFragment`. It seems that other objects using the same table are also using `id`. Do you know of a way around it or are we fine to leave it as `id`?

apps/backend/src/mutations/TechniqueMutations.spec.ts
simonfernandes: Just to check, are you referring to the composite key in the `technique_has_instruments` table? If we do this as a backend test then I think this means we would only be testing the implementation of the mockup. And if we add this as an e2e test then it seems we would just be testing how postgres composite keys work. Or did you mean otherwise?

simonfernandes:
@yoganandaness These are some really good comments, thanks Yoganandan. We have addressed most of these now but have left a couple of things with comments to clarify.

yoganandaness (changes requested):
Thanks for the response. Found some more improvements, when i went through the PR. Could you please take a look. 

Kindly let me know in case of any concerns.

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
```diff
@@ -0,0 +1,150 @@
...
+    const [result]: TechniqueRecord[] = await database('techniques')
+      .update({
+        name: technique.name,
+        short_code: technique.shortCode,
+        description: technique.description,
+      })
+      .where('technique_id', technique.id)
+      .returning('*');
+
+    return technique ? this.createTechniqueObject(result) : Promise.reject();
```
mutambaraf, line 103: We may want to include the reason for this rejection.

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const result = await database('technique_has_instruments').insert(
+      dataToInsert
+    );
+
+    if (result) {
+      return true;
+    } else {
+      return false;
+    }
```
mutambaraf, line 132: Do we need the else looks like we have only two options.

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
```diff
@@ -0,0 +1,150 @@
...
+    const result = await database('technique_has_instruments')
+      .where('instrument_id', instrumentId)
+      .andWhere('technique_id', techniqueId)
+      .del();
+
+    if (result) {
+      return true;
+    } else {
+      return false;
+    }
```
mutambaraf, line 148: We only have two condition we may not need else.

apps/backend/src/datasources/mockups/TechniqueDataSource.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  async getTechnique(techniqueId: number): Promise<Technique | null> {
+    const technique = dummyTechniques.find((t) => t.id === techniqueId);
+
+    if (technique) {
+      return technique;
+    } else {
+      return null;
+    }
```
mutambaraf, line 35: Looks like we only have two conditions here.

apps/backend/src/resolvers/types/Technique.ts
```diff
@@ -0,0 +1,47 @@
...
+  async instruments(
+    @Root() technique: Technique,
+    @Ctx() context: ResolverContext
+  ): Promise<Instrument[] | null> {
+    const instruments =
+      context.queries.technique.dataSource.getInstrumentsByTechniqueId(
+        technique.id
+      );
+
+    return isRejection(instruments) ? [] : instruments;
```
mutambaraf, line 45: If this is rejected why are we returning an empty array .Unless we have some way to filter and log certain rejections and allow some.

apps/backend/src/mutations/TechniqueMutations.ts
```diff
@@ -0,0 +1,125 @@
...
+  async create(
+    agent: UserWithRole | null,
+    args: CreateTechniqueArgs
+  ): Promise<Technique | Rejection> {
+    const technique = await this.dataSource.create(args).catch((error) => {
+      return rejection(
+        'Could not create technique',
+        { agent, shortCode: args.shortCode },
+        error
+      );
```
mutambaraf, line 44: I am not seeing the underlining function propagating any error message error will most likely be always not set.

apps/backend/src/mutations/TechniqueMutations.ts
```diff
@@ -0,0 +1,125 @@
...
+    args: UpdateTechniqueArgs
+  ): Promise<Technique | Rejection> {
+    const updatedTechnique = await this.dataSource
+      .update(args)
+      .catch((error) => {
+        return rejection(
+          'Could not update technique',
+          { agent, techniqueId: args.id },
+          error
+        );
```
mutambaraf, line 64: We not be passing any error message from update method.

apps/frontend/src/components/technique/CreateUpdateTechnique.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const { t } = useTranslation();
+  const { api, isExecutingCall } = useDataApiWithFeedback();
+
+  const initialValues = technique
+    ? { ...technique }
+    : {
+        name: '',
+        shortCode: '',
+        description: '',
+      };
```
mutambaraf, line 31:
```suggestion
  const initialValues = {
        name: '',
        shortCode: '',
        description: '',
         ...technique 
      };
```

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
```diff
@@ -0,0 +1,150 @@
...
+  async create(args: CreateTechniqueArgs): Promise<Technique> {
+    const [technique]: TechniqueRecord[] = await database('techniques')
+      .insert({
+        short_code: args.shortCode,
+        name: args.name,
+        description: args.description,
+      })
+      .returning('*');
+
+    return technique ? this.createTechniqueObject(technique) : Promise.reject();
```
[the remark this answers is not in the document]
simonfernandes: It catches and rejects with a message at the mutation level, I thought this would be a better place to handle this kind of logic

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
```diff
@@ -0,0 +1,150 @@
...
+          query.offset(offset);
+        }
+      })
+      .then((techniques: TechniqueRecord[]) => {
+        const result = techniques.map((technique) =>
+          this.createTechniqueObject(technique)
+        );
+
+        return {
+          totalCount: techniques[0] ? techniques[0].full_count : 0,
```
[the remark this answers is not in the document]
simonfernandes: This is so that the total total count is accurate when using the query with an offset or limit

apps/backend/src/datasources/postgres/TechniqueDataSource.ts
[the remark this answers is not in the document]
simonfernandes: I see what you meant now 😁

apps/backend/src/mutations/TechniqueMutations.ts
simonfernandes: I missed this, thanks. Now the datasource throws errors with messages when something unexpected happens. Also made sure all mutations/queries catch the error and return a rejection including the reason and any relevant args.

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+      const techniqueId = selectedTechnique.id;
+      if (instrumentIds?.length) {
+        instrumentIds.forEach(async (instrumentId) => {
+          await api({
+            toastSuccessMessage: `Instrument/s unassigned from selected technique successfully!`,
+          })
+            .removeInstrumentFromTechnique({
+              instrumentId,
+              techniqueId,
+            })
```
mutambaraf, line 130: Do we need to do multiple backend calls or we can potentially move this to the backend and  let it handle the load and we can update the frontend.

simonfernandes:
@yoganandaness Deepak has now added a video above 🙂

apps/backend/src/eventHandlers/logging.ts
```diff
@@ -177,6 +182,58 @@ export default function createHandler() {
...
+            }
+
+            await eventLogsDataSource.set(
+              event.loggedInUserId,
+              event.type,
+              json,
+              '',
+              description
+            );
+          }
```
mutambaraf, line 235: Can we include the changedObjectId as the techniqueId rather than empty string same as above.

apps/backend/src/eventHandlers/logging.ts
```diff
@@ -223,15 +223,15 @@ export default function createHandler() {
...
-              event.type,
-              json,
-              '',
-              description
-            );
+              await eventLogsDataSource.set(
+                event.loggedInUserId,
+                event.type,
+                json,
+                obj[0].techniqueId,
```
mutambaraf, line 231: If this id not a string already we must need to convert it to a string.

deepaksftc: Since obj is from JSON.parse(), its of type any. It's getting converted to string type when getting passed to this method call.

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,233 @@
...
+          }}
+          actions={
+            isUserOfficer
+              ? [
+                  {
+                    icon: AssignmentIndIcon,
+                    tooltip: 'Assign/remove instruments',
+                    onClick: (_event: unknown, rowData: unknown): void =>
+                      setAssigningTechniqueId(rowData as TechniqueFragment),
+                  },
```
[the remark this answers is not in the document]
simonfernandes: Icon now looks like

image

mutambaraf (approved):
This looks good to me just , that one comment.

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -0,0 +1,256 @@
...
+        id: techniqueDeletedId as number,
+      });
+
+      return true;
+    } catch (error) {
+      return false;
+    }
+  };
+
+  const AssignmentIndIcon = (): JSX.Element => <ScienceIcon />;
```
yoganandaness, line 78: Can we use the Icon, that we use for assigning the Instrument to the Proposal or Call. 
Screenshot 2024-07-01 at 11 29 01

Available in `user-office-core/apps/frontend/src/components/common/icons/ScienceIcon.tsx`

yoganandaness (changes requested):

yoganandaness:
Overall, it looks great. Just a small comment. Thanks.

apps/frontend/src/components/technique/TechniqueTable.tsx
simonfernandes: Hi Yoganandan, I think we may have already changed this in https://github.com/UserOfficeProject/user-office-core/pull/628/commits/24be98e51c4797b7d0bf5583d1d5b280e23c9f9d, if I've understood you correctly.



This is what it looks like now:

image

yoganandaness: @simonfernandes Oh That's my bad. I checked the Video and it had the different Icon. But, I should have checked the code completely.

PR #647, opened 2024-07-02
feat: make tech revs optional

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1092

## Description
This PR makes the technical reviews an optional feature configurable by each individual call.

## Motivation and Context
This change is required to provide flexibility for calls that do not require technical reviews, thus simplifying the process and saving unnecessary effort.

## Changes
1. Introduced a new boolean field `need_tech_review` in the `call` table to indicate whether technical reviews are needed for a particular call.
2. Updated the `Call` model and data source to handle the new `need_tech_review` field.
3. Modified `ProposalDataSource` to check if a proposal needs technical review based on its associated call.
4. Updated `InstrumentMutations` to only set technical review if the proposal's call needs it.

## Fixes Jira Issue


bolmsten:
Why can we not solve this with the proposal workflows? If I have a call that does not require technical review I do not add feasibility review in the workflow for that call?

TCMeldrum:
When you assign an instrument to a proposal, it automatically creates and requests a tech review for it. This then shows up on instrument scientists pages as available to fill in and has been confusing. 

I get what you mean, though we could tie tech review creation to the workflow rather than the call.

bolmsten:
@martin-trajanovski @TCMeldrum  

An idea would be to check if a proposal has the feasibility review in the workflow, if not, do not show the feasibility tab. The same functionality could be used for the FAP reviews, I am thinking about commissioning proposals or rapid access that does not always have FAP.

TCMeldrum:
I think that would work, but last round some instrument scientists decided that not all of them want to fill in feasibility reviews and some of them did. So we have a use case where some proposals needed reviews before the fap and some didn't. Maybe worth discussing at next sprint catch up meeting.

PR #650, opened 2024-07-09
fix: remove expired proposals for instrument scientists

ellen-wright:
Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1089
## Description
Added an array to remove proposals of any given status from the instrument scientist table. At the moment the only status being removed is 'Expired' however, different statuses can be easily added/ removed from this list.
The same statuses in the list will also be removed from the status filter for instrument scientists.

## How Has This Been Tested
This has been tested manually


apps/backend/src/datasources/stfc/StfcProposalDataSource.ts
```diff
@@ -61,6 +61,7 @@ export default class StfcProposalDataSource extends PostgresProposalDataSource {
       .from('proposal_table_view')
       .whereIn('proposal_pk', proposals)
       .orderBy('proposal_pk', 'desc')
+      .where('proposal_status_id', '<>', '9')
```
mutambaraf, line 64: Can this be made to be a filter rather than hardcoding it we can pass a list of status which must be excluded

apps/frontend/src/hooks/proposal/useProposalsCoreData.ts
bashanlam, line 1: Have you performed regression testing on the ProposalTableOfficer.tsx component in apps/frontend/src/components/proposal/? This component utilizes the useProposalsCoreData.ts hook from apps/frontend/src/hooks/proposal/ that may be affected?

ellen-wright: Expired proposals still appear in the User Officer table, the `useProposalsCoreData `for user officer table is not effected if nothing is passed for the  `excludeProposalStatusIds`

bashanlam (approved):

janosbabik (commented):
My understanding of the problem is that when the status filter is set to ALL, EXPIRED proposals are also shown. Is that correct?

The only problem with this solution is that if I select the EXPIRED status explicitly, I do not get back any proposals even though there are EXPIRED proposals. So, somehow, it is a bug.

What if the statuses in the 'Status' filter could be selectable with checkboxes? Then we could fetch proposals with different statuses, and by default, we could select every status except EXPIRED.

ellen-wright:
Yeah so basically it was causing confusion with instrument scientists when EXPIRED proposals were showing in the instrument scientist table. I get what you mean about not being able to access these hidden proposals, I will look into the checkbox solution

apps/frontend/src/hooks/proposal/useProposalsCoreData.ts
mutambaraf: I do not know if Instrument Scientist needs to be able to view expired proposals given there is no further processes on those proposals. This is something we might need to check with the requirements.

mutambaraf:
I do not know if Instrument Scientist needs to be able to view expired proposals given there is no further processes on those proposals. This is something we might need to check with the requirements.

janosbabik:
> I do not know if Instrument Scientist needs to be able to view expired proposals given there is no further processes on those proposals. This is something we might need to check with the requirements.

I think this is not about the requirements. The frontend has a status filter that would not work correctly with this change (STFC side).
Instead of having hardcoded exclusions, I would improve the filter component to allow the selection of multiple proposal statuses, giving users control over what they want to see. 
By default, it could be set to include all statuses except EXPIRED for the Instrument scientist’s proposal table.
I was thinking of a filter like the one we have in the Scheduler application:
Screenshot 2024-08-01 at 13 54 28

mutambaraf:
> > I do not know if Instrument Scientist needs to be able to view expired proposals given there is no further processes on those proposals. This is something we might need to check with the requirements.
> 
> I think this is not about the requirements. The frontend has a status filter that would not work correctly with this change (STFC side). Instead of having hardcoded exclusions, I would improve the filter component to allow the selection of multiple proposal statuses, giving users control over what they want to see. By default, it could be set to include all statuses except EXPIRED for the Instrument scientist’s proposal table. I was thinking of a filter like the one we have in the Scheduler application: Screenshot 2024-08-01 at 13 54 28

Maybe we should remove the filter for expired proposals so that its not even available given expired proposals have no much use since they are no longer part of any process.

janosbabik (approved):

janosbabik (approved):

PR #655, opened 2024-07-15
fix: proposal export doesn't include all selected proposals

deepaksftc:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1083

## Description

Instead of row data, the export function is changed to consider all the selected proposals.

## Motivation and Context

When you select a list of proposals and "Export proposals in Excel", it will only include the first 20 proposals in the exported spreadsheet. All proposals need to show in the spreadsheet for it to be useful and to avoid repetition.

## How Has This Been Tested

Manually

## Fixes

https://github.com/UserOfficeProject/issue-tracker/issues/1083

## Changes

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx

### DIFF
--- a/apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
+++ b/apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
@@ -973,10 +973,10 @@ const ProposalTableOfficer = ({
           {
             icon: ExportIcon,
             tooltip: 'Export proposals in Excel',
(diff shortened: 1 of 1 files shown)


simonfernandes (approved):

martin-trajanovski (approved):
As a temporary fix this seems good to me. I know that we will need to refactor this a bit after my big PR gets merged with material-ui and table upgrade.

PR #660, opened 2024-07-17
feat: add conflict of interest feature flag

TCMeldrum:
Closes UserOfficeProject/issue-tracker#1074

## Description
This PR introduces a new feature flag for managing the display of conflict of interest warnings when assigning reviewers. 
Within STFC due to how the data is stored the logic around this warning would be difficult to implement and since Secs have been doing this anyway we are just going to disable the warning. 

## Motivation and Context
The motivation behind this change is to provide flexibility and control over the display of conflict of interest warnings. This is essential as it helps to prevent potential bias or unfairness in the review process.

## Changes
1. A new feature flag `USE_CONFLICT_OF_INTEREST_WARNING` has been added to the backend and frontend, which controls the display of conflict of interest warnings.
2. The feature flag is initially set to true, enabling the conflict of interest warnings by default.
3. The display of warnings in `FapProposalsAndAssignmentsTable` component is now conditional, based on the state of the feature flag.


Scott-James-Hurley:
Did this come out of the FAP review feedback?

TCMeldrum:
> Did this come out of the FAP review feedback?

Yeah, the logic for what a conflict was at stfc differed between CLF and ISIS and the FAP secs were already hyper aware of any conflict of interest so it was decided just to remove it for STFC

apps/e2e/cypress/e2e/FAPs.cy.ts
```diff
@@ -24,6 +24,16 @@ const fapMembers = {
   reviewer2: initialDBData.users.user3,
 };
 
+function clickConfirmOk() {
```
joshhdawes, line 27: From what I can see, this puts the conflict-of-interest feature under a feature flag (which is what we discussed), but if that feature flag is enabled we still have a broken conflict-of-interest feature?

TCMeldrum: That would be the case, yes. I thought we chose to turn it off for stfc then fix, was this not what we discussed?

joshhdawes: Yeah I think it was actually - from the meeting we had it doesn't sound like it'll ever actually be used, so I wouldn't be surprised if we just deleted it completely one day.

Scott-James-Hurley (approved):

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
```diff
@@ -173,6 +181,11 @@ const FapProposalsAndAssignmentsTable = ({
     settingsFormatToUse: SettingsId.DATE_FORMAT,
   });
   const { enqueueSnackbar } = useSnackbar();
+  const { featuresMap } = useContext(FeatureContext);
+
+  const useWarning = featuresMap.get(
```
yoganandaness, line 186:
```suggestion
  const isUseConflictOfInterestWarningEnabled = featuresMap.get(
```

yoganandaness: Just to get the variable naming more aligned with the feature flag name, so scanning the code is more easier.

yoganandaness (changes requested):

yoganandaness (approved):

PR #663, opened 2024-07-19
fix: proposal pdf download fails when instrument picker question is used in proposal question template

Bhaswati1148:
## Description

- Code fix to address the download proposal pdf issue when instrument picker question(along with time request option) is included in the question template. 
- Separate end to end test case has been added to test the above scenario.
- PR also involves some minor changes like adding the request time field in radio button group and removing the ability to input decimal values in time request field.

## How Has This Been Tested
This has been tested manually as well as executing e2e test.

## Fixes
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1111

## Changes

- Code changes in the logic how values are extracted from an answer for an instrument picker question data type at two places `addTopicInformation` and `collectProposalPDFData` in  `apps/backend/src/factory/pdf/proposal.ts`.
- Added a separate e2e test case in `apps/e2e/cypress/e2e/proposals.cy.ts` to test proposal pdf download when instrument picker question is included in template.
- Removing the step size from request time field in `apps/frontend/src/components/questionary/questionaryComponents/InstrumentPicker/QuestionaryComponentInstrumentPicker.tsx`
- Adding the request time field in radio button group in `apps/frontend/src/components/questionary/questionaryComponents/InstrumentPicker/QuestionaryComponentInstrumentPicker.tsx`


martin-trajanovski (approved):
This seems fine to me. The only thing I am concerned is the data migration. I know we mentioned that but not sure if it is really needed or not. Maybe that should be part of another PR but just making sure that we are keeping track of it.

Bhaswati1148:
> This seems fine to me. The only thing I am concerned is the data migration. I know we mentioned that but not sure if it is really needed or not. Maybe that should be part of another PR but just making sure that we are keeping track of it.

I have added the migration scripts in the [comments](https://github.com/UserOfficeProject/issue-tracker/issues/1111#issuecomment-2239302104) section of the issue https://github.com/UserOfficeProject/issue-tracker/issues/1111

I am not sure if the scripts should be kept in the db patches in repo, so didn't include them in this PR. Basically there are two scripts, one for single instrument and another for multiple instruments, both will have separate question ids so thought that one should manually run them when required in production.

apps/backend/db_patches/0158_UpdateOldInstrumentPickerAnswers.sql
```diff
@@ -0,0 +1,34 @@
+DO
+$$
+BEGIN
```
martin-trajanovski, line 3: I will try to run this against our develop database and get back with approval if everything works.

martin-trajanovski: Hey @Bhaswati1148 after running this migration locally I noticed that it works well when we had value just as number for example:
this answer:
```typescript
{
  "value": 4
}
``` 
was converted to this: 
```typescript
{
  "value": {
    "instrumentId": "4",
    "timeRequested": null
  }
}
``` 

but it seems to have a problem when converting some other cases like for example:
this one:
```typescript
{
  "value": [4]
}
``` 
was converted to this:
```typescript
{
  "value": {
    "instrumentId": "[4]",
    "timeRequested": null
  }
}
``` 
which I think it is wrong. It should be something like:
```typescript
{
  "value": [{
    "instrumentId": "4",
    "timeRequested": null
  }]
}
``` 
And then another example is where we have:
```typescript
{
  "value": [19, 20, 21]
}
``` 
converted to:
```typescript
{
  "value": {
    "instrumentId": "[19, 20, 21]",
    "timeRequested": null
  }
}
``` 
which I also think it is wrong and it should be:
```typescript
{
  "value": [{
    "instrumentId": "19",
    "timeRequested": null
  },
  {
      "instrumentId": "20",
      "timeRequested": null
   },
  {
      "instrumentId": "21",
      "timeRequested": null
  }]
}
```

martin-trajanovski (changes requested):
Please check my comment on the migration script

martin-trajanovski:
@Bhaswati1148 I have run the whole file and that's how I got into the states that I showed by example. So something is not working that good. Can you please for example add manually a record in the database which was using the old structure and try to run the script after that to see the result?

Bhaswati1148:
> @Bhaswati1148 I have run the whole file and that's how I got into the states that I showed by example. So something is not working that good. Can you please for example add manually a record in the database which was using the old structure and try to run the script after that to see the result?

From our discussion on zoom today, it came out that 

- There were some weird combinations of questions and answers in your dev database, so the migration script didn't work properly for such cases. You corrected the incorrect data and now the migration script should work fine. 

- There was another anomaly found for answers like below. I have updated the script to handle such scenario. 
```
{
  "value": {
    "instrumentId": "10",
    "timeRequested": null
  }
}
```
Please rerun the updated script and let me know

martin-trajanovski (approved):
Seems fine to me right now.

apps/e2e/cypress/e2e/proposals.cy.ts
```diff
@@ -1616,4 +1619,154 @@ context('Proposal tests', () => {
...
+              instrumentFapIds: [{ instrumentId: result.createInstrument.id }],
+            });
+          });
+        }
+      });
+      cy.login('officer');
+      cy.visit('/');
+    });
+    it('Should be able to download proposal pdf for a proposal which contains instrument picker question', function () {
+      if (!featureFlags.getEnabledFeatures().get(FeatureId.SCHEDULER)) {
```
TCMeldrum, line 1712: is there any reason we are not running these in e2e mode and only stfc mode?

Bhaswati1148: The test case was failing without this check and now I don't remember the message shown for the failed test case. 
However when I checked in other places like [here](https://github.com/UserOfficeProject/user-office-core/blob/4ab6e3c9984d1ba180dbfce7e84b3ac20b35b504/apps/e2e/cypress/e2e/proposalAdministration.cy.ts#L474) and [here](https://github.com/UserOfficeProject/user-office-core/blob/4ab6e3c9984d1ba180dbfce7e84b3ac20b35b504/apps/e2e/cypress/e2e/proposalAdministration.cy.ts#L505) I found this check was there, so I added that in the new test case and it worked.

TCMeldrum: This flag is often used to check which config we are running in (e2e vs stfc) so we are skipping the test in e2e mode (which aligns more closly with ess config) since this is not something that is stfc specific (correct me if im wrong) I think we should have it working in both modes

PR #673, opened 2024-07-30
fix(stfc): prevent errors when an external user number doesn't exist

ACLay:
## Description

Returns a dummy user from the STFC user data source when the user number can't be found in our external user service.

Currently in draft until I've had a review meeting to check we want to go with this approach.

## Motivation and Context

The UOP is incredibly config driven. While this makes it flexible to support multiple templates and facilities, it means there's an overwhelming amount of config that needs setting up in local environments if we want other software to interact with it in a way that is a fair approximation of our production environment.

I've been looking at creating a postgres image that imports a dump of our prod database, providing access to all the same questions, calls, templates, faps... that we have on prod.

An issue I found while testing this is that many user numbers that exist in our prod environment don't in our local one. When UOP encounters a missing user, it returns null for secretaries, investigators, instrument scientists... where there should never be null, causing the site to become non functional. 

I can't think of a good way to decouple non existent user numbers from the rest of the data without losing vast quantities or creating mismatches. What we can do instead, is return a dummy user that clearly states it's not a real one.

## Fixes

Part of https://github.com/UserOfficeProject/issue-tracker/issues/343

## Depends on

This could go in before them, but this is to use the database from:

https://github.com/isisbusapps/local-db/pull/45
https://github.com/isisbusapps/docker-orchestration/pull/598

### DIFF
--- a/apps/backend/src/datasources/stfc/StfcUserDataSource.ts
+++ b/apps/backend/src/datasources/stfc/StfcUserDataSource.ts
@@ -75,6 +75,24 @@ export function toEssBasicUserDetails(
   );
 }
 
+function essBasicUserDetailsForMissingUser(
+  userNumber: number
+): BasicUserDetails {
+  return new BasicUserDetails(
+    userNumber,
+    'Missing',
+    'User ' + userNumber,
+    'Missing',
+    '',
+    1,
+    '',
+    new Date(),
+    true,
+    '',
+    ''
+  );
+}
+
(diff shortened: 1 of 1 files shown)


ACLay:
At a review meeting it was suggested that rather than mocking users here, we insert them into the users oracle database. That way we don't have to worry about missing user numbers affecting our other systems that consume data from the UOP API.

PR #678, opened 2024-08-01
feat: integrate EAM asset creation json api

yoganandaness:
## Description

Update the EAM API call from SOAP to Rest. (XML to JSON)

## Motivation and Context

Adapt to the latest version of EAM

## How Has This Been Tested

Manually

## Changes

Removed all the XML related code and added the json payload. 


apps/backend/package-lock.json
```diff
@@ -50,6 +50,7 @@
         "string-strip-html": "^8.4.0",
         "tsyringe": "^4.8.0",
         "type-graphql": "2.0.0-beta.1",
+        "undici": "^6.19.5",
```
janosbabik, line 53: Please remove the `package-lock.json` file from the changes.

apps/backend/package.json
```diff
@@ -124,4 +124,4 @@
     "npm": ">=9.0.0",
     "node": ">=18.0.0"
   }
-}
+}
```
janosbabik, line 127: No changes are needed in the package.json file, please remove it.

apps/backend/src/services/assetRegistrar/eam/utils/performApiRequest.ts
```diff
@@ -1,32 +1,33 @@
...
+        INFOR_USER: getEnvOrThrow('EAM_AUTH_USER'),
+        INFOR_PASSWORD: getEnvOrThrow('EAM_AUTH_PASSWORD'),
+        INFOR_TENANT: getEnvOrThrow('EAM_AUTH_TENANT'),
+        INFOR_ORGANIZATION: getEnvOrThrow('EAM_AUTH_ORGANIZATION'),
       },
     });
-
-    const data = await response.text();
-
+    const data = await response.json();
```
janosbabik, line 17: Could you move this line after the response.ok check?

apps/backend/src/services/assetRegistrar/eam/api/createContainer.ts
```diff
@@ -151,18 +151,12 @@ export async function createContainer(shipmentId: number) {
...
-
-  const response = await performApiRequest(request);
-
-  const regexFindEquipmentCode =
-    /<ns2:EQUIPMENTCODE>([0-9]*)<\/ns2:EQUIPMENTCODE>/;
-  const result = response.match(regexFindEquipmentCode);
-
-  if (!result || result.length < 2) {
+  const response = await performApiRequest('/equipment', request);
+  if (!response) {
```
janosbabik, line 155: I believe this is always true if the response is OK. If it is not, the `performApiRequest` fn throws an error.

janosbabik (changes requested):

apps/backend/src/services/assetRegistrar/eam/utils/performApiRequest.ts
yoganandaness: @janosbabik in case of any 400 errors, we get a error response as a json and we need to log it. Hence, i need to call this before doing the response.ok check, so that i can send the data into the log.

janosbabik (approved):

TCMeldrum (approved):
LGTM. STFC don't use this but happy to approve it as I've been requested a review

PR #679, opened 2024-08-01
fix: fix fap review and member selections

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1075

## Description
This PR focuses on fixing FAP review and member selections within the application. There were a few bugs i found with the new data fetching which I also fixed in this PR.

## Motivation and Context
The changes were necessary to resolve issues linked with the FAP review and member selections, which were not working as expected. The changes help in enhancing the user experience and the overall functionality of the system.

## Changes
- Added email search functionality option to the member selection in AssignFapMemberToProposalModal.
- Enhanced the PeopleTable component by adding email search functionality option and improving the user selection process.
  
The changes made are aimed at improving the functionality of FAP review and member selection, thereby making the system more efficient and user-friendly.


apps/frontend/src/components/user/PeopleTable.tsx
```diff
@@ -367,12 +368,33 @@ const PeopleTable = ({
             userRole: query.userRole,
           });
 
+          const paginatedFilteredData = data
```
martin-trajanovski, line 371: I am wondering isn't it better to do the backend filter and search instead of frontend one or maybe I am not getting the point of this so well.

apps/frontend/src/components/fap/Proposals/AssignFapMemberToProposalModal.tsx
```diff
@@ -69,6 +69,7 @@ const AssignFapMemberToProposalModal = ({
           title="Select reviewers"
           selection={true}
           data={members}
+          emailSearch={false}
```
martin-trajanovski, line 72: Is there a specific reason why we want to disable this emailSearch here?

apps/frontend/src/components/user/PeopleTable.tsx
TCMeldrum: Sorry didn't think of this at the time, looking into this I would need to paginate currently 3 places where the data field is used in 3 different places: 

- https://github.com/UserOfficeProject/user-office-core/blob/1075-fix-fap-review-page/apps/frontend/src/components/questionary/questionaryComponents/ProposalBasis/QuestionaryComponentProposalBasis.tsx#L44
- https://github.com/UserOfficeProject/user-office-core/blob/1075-fix-fap-review-page/apps/frontend/src/components/proposalBooking/CreateUpdateVisit.tsx#L88
- https://github.com/UserOfficeProject/user-office-core/blob/1075-fix-fap-review-page/apps/frontend/src/components/fap/Proposals/AssignFapMemberToProposalModal.tsx#L71

As far as I can tell, the first 2 look like they tricky to paginate, as the data that is passed to the PeopleTable is used elsewhere in the file. Am I understanding it correctly

martin-trajanovski: Maybe it is just good to open a ticket about it for now and then we can take it from there.

apps/frontend/src/components/fap/Proposals/AssignFapMemberToProposalModal.tsx
martin-trajanovski: Is it the same for everybody or this is only STFC specific?

TCMeldrum: Currently stfc specific

apps/frontend/src/components/user/PeopleTable.tsx
TCMeldrum: Opened https://github.com/UserOfficeProject/issue-tracker/issues/1141

apps/frontend/src/components/fap/Proposals/AssignFapMemberToProposalModal.tsx
martin-trajanovski: If it is so we can make it maybe as an app setting and if we or someone else want to use it then we can choose.

TCMeldrum: Sorry I'm a bit confused, you want to make the email search feature into a setting? The email search I am disabling here is only used by stfc at the moment.

joshhdawes:
Since there are some UI changes in this (adding the email search), it seems to me that there should be some additions to e2e tests (however straightforward they may be).

joshhdawes:
> > @TCMeldrum since there are some UI changes in this (adding the email search), it seems to me that there should be some additions to e2e tests (however straightforward they may be). Unless there are already tests to cover this.
> 
> @joshhdawes The only UI changes are changing the dropdown to select the table size that seemed to small to add a test for, but I'll go back and add one. I am taking away an email search from a component that does not need it so I don't think I need to test that.

@TCMeldrum if there are already tests to capture the new behaviour that you expect from the change, then that's fine. Otherwise, the tests need updating.

TCMeldrum:
@joshhdawes Sorry got my PRs mixed up I was thinking of #681, there is no new behaviour in this PR, just fixing broken ones.

apps/frontend/src/components/fap/Proposals/AssignFapMemberToProposalModal.tsx
martin-trajanovski: Ah OK that's fine then. Sorry for missunderstanding

Scott-James-Hurley:
image
When testing this locally it looks like the navigation buttons at the bottom are greyed out, even if there more members to display.

Scott-James-Hurley (approved):
ta

apps/frontend/src/components/fap/Proposals/AssignFapMemberToProposalModal.tsx
bolmsten: I think it is fine

janosbabik (approved):

PR #680, opened 2024-08-02
fix: technical review check should be performed against the shortCode instead of name

martin-trajanovski:


janosbabik (approved):

mutambaraf:
But will it not be better to pass this as a filter from the frontend.

TCMeldrum (approved):

PR #681, opened 2024-08-02
feat: on proposal assignment page show all proposals in table

TCMeldrum:
Closes UserOfficeProject/issue-tracker#1091

## Description
This PR introduces a feature that displays all proposals in the table on the proposal assignment page.

## Motivation and Context
The change is required to enhance user experience by providing a comprehensive view of all proposals at once, eliminating the need to navigate through multiple pages to view different proposals.

## Changes
- Enabled the display of all proposals on the proposal assignment page.
- Added the functionality to adjust the page size based on the number of proposals, with options to display 5, 10, 20, or all proposals.
- Enhanced the selection feature to accommodate the new change.


apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
```diff
@@ -765,6 +765,15 @@ const FapProposalsAndAssignmentsTable = ({
           options={{
             search: true,
             selection: true,
+            pageSize: Math.min(10, FapProposalsWitIdAndFormattedDate.length),
+            pageSizeOptions: [
```
martin-trajanovski, line 769: This looks good to me but you can maybe just put the array in a variable called `pageSizeOptions` or something similar and then just use that one here instead of doing the calculation and filtering directly in the html code.

martin-trajanovski (changes requested):

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
```diff
@@ -765,6 +767,10 @@ const FapProposalsAndAssignmentsTable = ({
           options={{
             search: true,
             selection: true,
+            pageSize: Math.min(10, maxPageLength),
+            pageSizeOptions: [5, 10, 20, maxPageLength]
```
martin-trajanovski, line 771: I was thinking to put even all of this in another variable would be nice. Something like:
```typescript
const pageSizeOptions = [5, 10, 20, maxPageLength]
              .sort()
              .filter((n) => n <= maxPageLength)
```

martin-trajanovski (approved):

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
```diff
@@ -765,6 +771,8 @@ const FapProposalsAndAssignmentsTable = ({
           options={{
             search: true,
             selection: true,
+            pageSize: Math.min(10, maxPageLength),
```
ACLay, line 774: Seems odd to show 5 as an option for page size if we're defaulting to the lower of everything or 10

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
```diff
@@ -716,6 +716,12 @@ const FapProposalsAndAssignmentsTable = ({
     })
   );
 
+  const maxPageLength = FapProposalsWitIdAndFormattedDate.length;
+
+  const pageSizeOptions = [5, 10, 20, maxPageLength]
+    .sort()
```
janosbabik, line 722: Just a small thing: using the sort function on numbers without a compare function gives the wrong order. [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description)

janosbabik (changes requested):

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
TCMeldrum: The 5 comes from the tables comments default options which as used across the application: https://github.com/material-table-core/core/blob/6303bd78399666e129a91ed28154ddd4d38c22c2/src/defaults/props.options.js#L26 and the 10 is what we set as default before. Should I remove the 5 in this case?

ACLay (approved):

janosbabik (approved):

PR #689, opened 2024-08-06
fix: update fap proposal count to only include proposals currently in fap review

TCMeldrum:
Closes UserOfficeproject/issue-tracker#1124
Closes UserOfficeproject/issue-tracker#1136

## Description
This PR updates the FAP proposal count to include only proposals currently in FAP review.

## Motivation and Context
The change is required to accurately reflect the number of proposals under FAP review. It solves the problem of inflated counts which included proposals not currently under review.

## Changes
1. A new method `getCurrentFapProposalCount` has been added to the `FapDataSource` which fetches the count of proposals currently in FAP review.
2. The `getFapProposalCount` method in the `FapDataSourceMock` has been updated to reflect the current count.
3. In the `PostgresFapDataSource`, a new query has been added to fetch only the proposals where the review has not ended.
4. The `FapResolvers` has been updated to use the new method to fetch the current proposal count.
5. The `FapsTable` component now displays the current proposal count.


apps/e2e/cypress/e2e/FAPs.cy.ts
```diff
@@ -818,9 +847,6 @@ context('Fap reviews tests', () => {
 
       cy.login('officer');
 
-      cy.visit(`/FapPage/${createdFapId}?tab=1`);
-      cy.get('[data-cy="fap-reviewers-table"]').contains('0');
```
martin-trajanovski, line 822: I am wondering why this got removed

TCMeldrum: It used to be that proposal totals were only populated for reviewers after the call ended, but I removed this filter as proposals can in a review status before then.  This part of the tests was testing that they were not appearing before call ends, I think. Saying that, I should remove the call ending in this tests as well.

martin-trajanovski (approved):

apps/backend/src/datasources/FapDataSource.ts
```diff
@@ -76,6 +76,7 @@ export interface FapDataSource {
     callId: number
   ): Promise<BasicUserDetails[]>;
   getFapProposalCount(fapId: number): Promise<number>;
+  getCurrentFapProposalCount(fapId: number): Promise<number>;
```
ACLay, line 79: Would it make sense for the new field be renamed to `getFapProposalCountCurrentRound`? That way it'd be a consistent naming scheme with `getFapReviewerProposalCount` and `getFapReviewerProposalCountCurrentRound` below

TCMeldrum: Yeah, we should make them consistent, but I think I am in favour of removing the "round" from  `getFapReviewerProposalCountCurrentRound` as there is no concept of a round in the software only open calls which in theory there could be multiple of going on at once.

ACLay (approved):

PR #690, opened 2024-08-07
feat: add ELI configuration

gnyiri:
## Description
This PR introduces the ELI configuration to the application. 

## Motivation and Context
This change was necessary to configure the application specifically for ELI's environment and requirements. It allows the application to run smoothly in the ELI environment and caters to the specific needs of the ELI user base.

## Changes
1. A new file `dependencyConfigELI.ts` was added in the backend config which maps the necessary dependencies for the ELI environment.
2. The `configureELIEnvironment.ts` file was created to set up the ELI environment specific settings, such as colour theme and timezone.
3. The new ELI configuration was integrated into the application's main configuration file. This ensures that the ELI configuration is loaded when the application is running in the ELI environment.
   
## How Has This Been Tested
The changes have been tested locally by running the application in the ELI environment and verifying that the correct configuration is loaded and all dependencies are correctly mapped. The functionality of the application was tested to ensure it works as expected in this new environment.

## Fixes Jira Issue


apps/backend/src/config/eli/configureELIEnvironment.ts
```diff
@@ -0,0 +1,108 @@
...
+  );
+  await db.updateSettings({
+    settingsId: SettingsId.DEFAULT_INST_SCI_REVIEWER_FILTER,
+    settingsValue: 'ME',
+  });
+  await db.updateSettings({
+    settingsId: SettingsId.DEFAULT_INST_SCI_STATUS_FILTER,
+    settingsValue: 'FEASIBILITY_REVIEW',
+  });
+  /*await db.updateSettings({
```
martin-trajanovski, line 96: Do we want to remove this or enable it. Would be nice to avoid having commented code like this I guess.

apps/backend/src/eventHandlers/email/eliEmailHandler.ts
```diff
@@ -0,0 +1,324 @@
...
+          substitution_data: {
+            piPreferredname: principalInvestigator.preferredname,
+            piLastname: principalInvestigator.lastname,
+            proposalNumber: event.proposal.proposalId,
+            proposalTitle: event.proposal.title,
+            commentForUser: event.proposal.commentForUser,
+          },
+          recipients: [
+            { address: principalInvestigator.email },
+            /*{
```
martin-trajanovski, line 249: This comment here as well can be cleaned up

martin-trajanovski (changes requested):
It looks good in general. Just few comments that need some attention

apps/backend/src/config/eli/configureELIEnvironment.ts
gnyiri: Fixed, thanks.

apps/backend/src/eventHandlers/email/eliEmailHandler.ts
gnyiri: Fixed, thanks.

janosbabik (commented):
Hello @gnyiri, I am not sure about the new event you created, USER_CREATED. Also, the USER_PASSWORD_RESET_EMAIL event, which is used in the eliEmailHandler service, they are not published anywhere.

gnyiri:
> Hello @gnyiri, I am not sure about the new event you created, USER_CREATED. Also, the USER_PASSWORD_RESET_EMAIL event, which is used in the eliEmailHandler service, they are not published anywhere.

Hello @janosbabik, I simply removed those events, it is an old code from our early tests. We need to implement a new, non-SMTP  email service anyway.

janosbabik (approved):

TCMeldrum (approved):

bolmsten:
I think it would have been better to use the status actions for sending emails @janosbabik @gnyiri

PR #693, opened 2024-08-07
feat: make optional tech reviews setting

TCMeldrum:
Closes UserOfficeProject/issue-tracker#1144

## Description
This PR introduces an optional setting for technical reviews in the proposal workflow. 

## Motivation and Context
The change is required to provide flexibility in the proposal workflow, allowing certain proposals to bypass the technical review stage based on their workflow status. This feature is particularly useful when only a subset of proposals require technical review.

## Changes
1. A new setting `TECH_REVIEW_OPTIONAL_WORKFLOW_STATUS` has been added to the settings table, which defines the workflow status that triggers the creation of a technical review.
2. The `PostgresProposalDataSource` class has been updated to check this setting when determining if a proposal needs a technical review.
3. The e2e tests have been updated to accommodate the new optional technical review setting.


apps/backend/src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -1073,6 +1077,16 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
   }
 
   async doesProposalNeedTechReview(proposalPk: number): Promise<boolean> {
+    const getWorkflowStatus = (
```
martin-trajanovski, line 1080: Maybe it is better to name this just `workflowStatus` instead of using prefix get because it contains the status and not a function to get it.

apps/backend/db_patches/0160_OptionalTechReviewsSetting.sql
```diff
@@ -0,0 +1,16 @@
+DO
+$$
+BEGIN
+  IF register_patch('0160_OptionalTechReviewsSetting.sql', 'TCMeldrum', 'Optional Technical Reviews Setting', '2024-07-17') THEN
```
martin-trajanovski, line 4: Just the date should be fixed here

martin-trajanovski (commented):
In general it looks great! I left some comments that might need little attention but otherwise I am happy to approve. I hope for us at ESS this setting will be empty by default right? And it will include all proposals for tech reviews

TCMeldrum:
> In general it looks great! I left some comments that might need little attention but otherwise I am happy to approve. I hope for us at ESS this setting will be empty by default right? And it will include all proposals for tech reviews

Cool, thanks, that was the idea if it is `null` or `""` it should make a tech review.

martin-trajanovski (approved):

Scott-James-Hurley (approved):

PR #696, opened 2024-08-09
1123 create document store for faps

TCMeldrum:


TCMeldrum:
I don't know if other facilities would like this feature, if not I am happy to put it behind a feature flag

janosbabik:
> I don't know if other facilities would like this feature, if not I am happy to put it behind a feature flag

Hi @TCMeldrum, we can keep it as a default functionality.

apps/backend/src/models/Settings.ts
```diff
@@ -37,4 +37,5 @@ export enum SettingsId {
   GRADE_PRECISION = 'GRADE_PRECISION',
   USER_OFFICE_EMAIL = 'USER_OFFICE_EMAIL',
   SMTP_BCC_EMAIL = 'SMTP_BCC_EMAIL',
+  TECH_REVIEW_OPTIONAL_WORKFLOW_STATUS = 'TECH_REVIEW_OPTIONAL_WORKFLOW_STATUS',
```
janosbabik, line 40: Is this change connected to the current PR?

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -403,6 +403,7 @@ export interface FapRecord {
   readonly custom_grade_guide: boolean | null;
   readonly active: boolean;
   readonly full_count: number;
+  readonly files: string;
```
janosbabik, line 406:
```suggestion
  readonly files: string | null;
```

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -105,7 +105,8 @@ export default class PostgresFapDataSource implements FapDataSource {
     numberRatingsRequired: number,
     gradeGuide: string,
     customGradeGuide: boolean,
-    active: boolean
+    active: boolean,
+    files: string
```
janosbabik, line 109:
```suggestion
    files: string | null
```

apps/frontend/src/components/fap/FapPage.tsx
```diff
@@ -51,6 +53,10 @@ const FapPage = () => {
         />
       ),
     },
+    {
+      name: 'Documents',
+      element: <FapFilesView data={fap} />,
```
janosbabik, line 58: As a UserOfficer I believe that showing the upload component is sufficient. What do you think?

```suggestion
      element: isUserOfficer ? (
          <FapFiles
            data={fap}
            onFapUpdate={(newFap: Fap): void => setFap(newFap)}
          />
        ) : (
          <FapFilesView data={fap} />
        ),
```

janosbabik (changes requested):
Great feature @TCMeldrum! I have a few suggestions

apps/frontend/src/components/fap/FapPage.tsx
TCMeldrum: I left it so the user officer could see both so that they are able to see what the users of the system will see and where the documents "end up" and have a better understanding of where the documents are "going". I'm happy to remove this from their view if we think it is unnecessary.

janosbabik (approved):

ellen-wright (approved):

PR #697, opened 2024-08-12
feat: double-check if proposal workflow needs to be run multiple times and do some UI improvements

martin-trajanovski:
## Description
The PR introduces improvements to the proposal workflow and the user interface of the application.

## Motivation and Context
The change is required to ensure the proposal workflow runs multiple times if needed and enhances the user interface for better usability and user experience.

## Changes
1. Implemented a double-check mechanism in the workflow engine to verify if the proposal workflow needs to run multiple times.
2. Removed the 'question-search-toggle' button from various components, enhancing the user interface by reducing clutter and improving usability.
**Before:**
before

**After:**

after

3. Removed unnecessary states and props from ProposalFilterBar component, simplifying the component and making it more efficient.

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4113](https://jira.esss.lu.se/browse/SWAP-4113)


janosbabik (changes requested):
Hi @martin-trajanovski, nice update! Could you wait for the `checkIfConditionsForNextStatusAreMet` call at line 247 and the `workflowEngine` call at line 112? (workflowEngine/index.ts)

martin-trajanovski:
> Hi @martin-trajanovski, nice update! Could you wait for the `checkIfConditionsForNextStatusAreMet` call at line 247 and the `workflowEngine` call at line 112? (workflowEngine/index.ts)

Yes @janosbabik maybe this is a good idea. First I thought that it might be good to run it without waiting for it but the workflow engine is asynchronous anyway so there is not a problem even if we wait.

janosbabik:
> > Hi @martin-trajanovski, nice update! Could you wait for the `checkIfConditionsForNextStatusAreMet` call at line 247 and the `workflowEngine` call at line 112? (workflowEngine/index.ts)
> 
> Yes @janosbabik maybe this is a good idea. First I thought that it might be good to run it without waiting for it but the workflow engine is asynchronous anyway so there is not a problem even if we wait.

@martin-trajanovski , usually it needs to wait for them because if any of them throws an error, it will not be caught (resulting in uncaught errors...)

apps/backend/src/workflowEngine/index.ts
```diff
@@ -57,6 +59,67 @@ const shouldMoveToNextStatus = (
...
+            proposalEventsKey as keyof ProposalEventsRecord
+          ]
+      );
+
+      const nextStatusRulesFulfilled = allProposalCompleteEvents.includes(
+        sce.statusChangingEvent.toLowerCase()
+      );
+
+      if (sce.statusChangingEvent && nextStatusRulesFulfilled)
+        workflowEngine([
```
janosbabik, line 112:
```suggestion
        await workflowEngine([
```

janosbabik (changes requested):

apps/backend/src/workflowEngine/index.ts
martin-trajanovski: @janosbabik I think we don't really want to await this and block the running of the other code because of it. The idea was that the whole workflow runs in the background and not blocking the main thread.

janosbabik (approved):

Scott-James-Hurley (approved):

PR #707, opened 2024-08-16
feat: add retry of status actions emails

mutambaraf:
## Description
This PR introduces a retry mechanism for status actions emails in the system.

## Motivation and Context
The change is required to ensure that status actions emails are reliably sent to users. Previously, in case of an email sending failure, there was no system in place to retry the action, potentially leading to missed communications. This also include publishing proposal events when a status action is replayed on a proposal these includes a failure or success.

## Changes
1. Introduced new tables `status_actions_logs` and `status_actions_log_has_proposals` in the database to log status action attempts.
2. Added new mutations and queries for `statusActionsLogs` in the backend context.
3. Integrated `StatusActionsLogsDataSource` into the dependency configuration of the system.
4. Have also refactor addConnectionStatusActions in StatusActionsDataSource.ts when updated status actions we nolonger remove the old but we just update.

## How Has This Been Tested?
Added E2E tests.

## Fixes Jira Issue


bolmsten:
Hi @mutambaraf,

thanks for the work, I think this is a really nice feature to have. I have some feedback first regarding the UI:

Screenshot 2024-08-26 at 14 01 06

It would be nice if you can filter on status FAIL. Also replay of a message is going to always look like a success as the frontend shows the user a green box and does not refresh.  As this list will eventually become huge even more filters will probably be needed, such as filtering on call. 

I have not been able to see the RabbitMQ statuses here, is this missing?

apps/backend/src/datasources/postgres/StatusActionsLogsDataSource.ts
```diff
@@ -0,0 +1,230 @@
...
+  statusActionsSuccessful: 'status_actions_successful',
+  statusActionsStep: 'status_actions_step',
+};
+export default class PostgresStatusActionsLogsDataSource
+  implements StatusActionsLogsDataSource
+{
+  private createStatusActionsLogObject(
+    statusActionLog: StatusActionsLogRecord
+  ) {
+    return new StatusActionsLog(
```
bolmsten, line 32: Usually the creation is broken out and done in records I think

apps/backend/src/resolvers/queries/StatusActionsLogsQuery.ts
```diff
@@ -0,0 +1,104 @@
...
+  public statusActionsLogId: number | null;
+
+  @Field(() => Int)
+  public connectionId: number;
+
+  @Field(() => Int)
+  public actionId: number;
+
+  @Field(() => EmailStatusActionRecipients)
+  public statusActionsStep: EmailStatusActionRecipients;
```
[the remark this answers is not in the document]
mutambaraf: I have updated the name

apps/backend/src/statusActionEngine/statusActionUtils.ts
```diff
@@ -279,30 +281,76 @@ export const getInstrumentScientistsAndFormatOutputForEmailSending = async (
...
+  ) {
+    const statusActionsLogsArgs: StatusActionsLogsArgs = {
+      ...args,
+      statusActionsLogId: args?.statusActionsLogId || null,
+      statusActionsBy: args?.statusActionsBy || null,
+      statusActionsSuccessful,
+      statusActionsMessage,
+    };
+
+    if (statusActionsLogsArgs.statusActionsLogId) {
```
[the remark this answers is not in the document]
mutambaraf: I have improved the test to use truthy or false

apps/backend/src/datasources/postgres/StatusActionsLogsDataSource.ts
```diff
@@ -0,0 +1,230 @@
...
+        status_actions_message: args.statusActionsMessage,
+        status_actions_by: args.statusActionsBy,
+        status_actions_successful: args.statusActionsSuccessful,
+      })
+      .where({ status_actions_log_id: args.statusActionsLogId })
+      .returning('*')
+      .then((statusActionsLog) => {
+        return this.createStatusActionsLogObject(statusActionsLog[0]);
+      });
+  }
```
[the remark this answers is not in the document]
mutambaraf: This is coming from prettier config it is not allowing me to add spaces.

mutambaraf:
@bolmsten For the update on the UI I tried not to lockup the UI while waiting for the success/failure in the sending of emails .I added a refresh button next to search .I have also included search by proposal ID and a filter for Failed and successful actions. For us to filter by calls will have to add another endpoint for giving us calls that have status actions logs ,if necessary l can add the filter.

bashanlam (approved):

apps/backend/db_patches/0162_AddStatusActionsLogsTables.sql
```diff
@@ -0,0 +1,27 @@
...
+$$
+BEGIN
+  IF register_patch('0162_AddStatusActionsLogsTables.sql', 'Farai Mutambara', 'Add Status Actions Logs', '2024-09-02') THEN
+    BEGIN
+       CREATE TABLE IF NOT EXISTS status_actions_logs(
+            status_actions_log_id serial PRIMARY KEY,
+            connection_id int NOT NULL,
+            action_id int NOT NULL,
+            email_status_action_recipient text DEFAULT NULL,
+            status_actions_by int REFERENCES users (user_id) ON UPDATE CASCADE,
```
yoganandaness, line 11: do we need to have this tight integrity? Will this affect the user deletion?

apps/backend/db_patches/0162_AddStatusActionsLogsTables.sql
```diff
@@ -0,0 +1,27 @@
...
+            status_actions_successful boolean DEFAULT FALSE,
+            status_actions_message text DEFAULT NULL,
+            status_actions_tstamp TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+             FOREIGN KEY (connection_id, action_id) REFERENCES proposal_workflow_connection_has_actions (connection_id, action_id) ON UPDATE CASCADE ON DELETE CASCADE
```
yoganandaness, line 15: Do we really need to delete this log, if the corresponding workflow action is deleted?(ON DELETE CASCADE)

I would recommend that since this is the log, which needs to be persisted for auditing purpose, FK constraint is not really needed to avoid deletion.

yoganandaness (changes requested):

apps/backend/db_patches/0162_AddStatusActionsLogsTables.sql
mutambaraf: On our side we may still want to resend even successful emails for example when we edit an email template or user has not received the email for other reasons. We can decide to keep the the record of a status action but  it will have no link to any connection_id, action_id which we require to retry the status action. If it is for auditing we do pass the user information when a status is replayed which will be in the proposal event logs for each proposal affected by the status action.
image

mutambaraf: We can also rename this feature if the word logs is misleading.

apps/backend/db_patches/0162_AddStatusActionsLogsTables.sql
mutambaraf: Have removed this column since it is not necessary in this table.

apps/backend/src/mutations/StatusActionsLogsMutations.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    emailActionHandler(statusAction, workflowEngineProposals, {
+      statusActionsLogId: statusActionsLog.statusActionsLogId,
+      loggedInUserId: agent?.id,
+      statusActionRecipients:
+        statusActionsLog.emailStatusActionRecipient as EmailStatusActionRecipients,
+    });
+
+    return !!statusActionsLog;
```
yoganandaness, line 115: This is always going to be True. Should we return True instead.

apps/backend/src/mutations/StatusActionsLogsMutations.ts
```diff
@@ -0,0 +1,137 @@
...
+      );
+
+    if (!statusActionsLog) {
+      return rejection('Could not retrieve status actions log', {
+        agent,
+        statusActionsLogId,
+      });
+    }
+
+    return this.executeStatusActionsLog(statusActionsLog, agent);
```
yoganandaness, line 135: This is always going to be True. Should we return True instead.

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -21,203 +22,379 @@ import {
...
+};
 
-          const otherRecipients: EmailReadyType[] =
-            recipientWithTemplate.otherRecipientEmails.map((email) => ({
-              id: recipientWithTemplate.recipient.name,
-              email: email,
-              proposals: proposals,
-              template: recipientWithTemplate.emailTemplate.id,
-            }));
+export const emailStatusActionRecipients = async (
```
yoganandaness, line 83:
```suggestion
export const emailStatusActionRecipient = async (
```

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -21,203 +22,379 @@ import {
...
+        ${statusActionRecipients}`
+      );
+    }
+    emailStatusActionRecipients(
+      recipientWithTemplate,
+      proposalStatusAction,
+      proposals,
+      statusActionsLogId,
+      loggedInUserId,
+      recipientWithTemplate.recipient.name
```
yoganandaness, line 65: This is the redundant argument. `recipientWithTemplate` is already passed as the first parameter. Could we remove this?

Correct me if i am wrong.

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -21,203 +22,379 @@ import {
...
 };
 
-const sendMail = async (recipientsWithData: EmailReadyType[]) => {
+const sendMail = async (
+  recipientsWithData: EmailReadyType[],
+  statusActionLogger: (
+    actionSuccessful: boolean,
+    message: string
+  ) => Promise<void>,
+  statusActionsLogId?: number | null,
```
yoganandaness, line 321: I would recommend passing option arguments `successfulMessage` and `failMessage` instead of `statusActionsLogId`. This way, `sendEmail` need not know anything about the Status Action Log Feature.

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -21,203 +22,379 @@ import {
...
+          return res;
+        } catch (err) {
           logger.logError('Could not send email', {
             error: err,
           });
-        });
-    })
-  );
+
+          await publishMessageToTheEventBus(
```
yoganandaness, line 375: Could you also add this in the PR description, so that it gets logged. 

Publishing message in the event of email failure.

apps/backend/src/datasources/postgres/StatusActionsDataSource.ts
```diff
@@ -156,35 +182,63 @@ export default class PostgresStatusActionsDataSource
   }
 
   async addConnectionStatusActions(
-    input: AddConnectionStatusActionsInput
+    connectionStatusActionsInput: AddConnectionStatusActionsInput
```
yoganandaness, line 185: I can understand that there have been some refactoring on this function. Could you add that to the PR description for logging purpose.

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -21,203 +22,379 @@ import {
...
+            recipientWithData.proposals,
+            `${failMessage} to: ${recipientWithData.email} recipient: ${recipientWithData.id}`,
+            undefined,
+            loggedInUserId || undefined
+          );
+          throw err;
+        }
+      })
+    );
+    if (
```
yoganandaness, line 385: I think we need to handle this for each Email response.

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -21,203 +22,379 @@ import {
...
+      !!mailServiceResponse.length &&
+      mailServiceResponse[0].results.total_rejected_recipients === 0
```
yoganandaness, line 387: This only ensures the successfulness of first email response, however the other things are not checked.

yoganandaness (changes requested):
@mutambaraf Sorry for the getting back very late. 

I was able to come up with few more points, as i looked a bit deeper. Please resolve as applicable. 

Kindly let me know in case of any concerns.

apps/backend/src/statusActionEngine/emailActionHandler.ts
mutambaraf: I have added a check for all responses.

mutambaraf: Have added the check of all responses

apps/backend/src/statusActionEngine/emailActionHandler.ts
mutambaraf: Have added the checks

yoganandaness (approved):
Seems good to me.

bashanlam (approved):

PR #711, opened 2024-08-20
feat: switch stfc email to use status actions

mutambaraf:
## Description
This is a pull request to switch the STFC email to use status actions instead of the previously used method.

## Motivation and Context
The previous method of handling STFC emails was cumbersome and prone to errors. There was also a need to clean up the code and improve readability. This change simplifies the email handling process and makes the code easier to maintain and understand.

## Changes
The primary changes are done in two files: `SMTPMailService.ts` and `stfcEmailHandler.ts`. 

1. In `SMTPMailService.ts`, the variable `_email` has been renamed to `emailTemplates` for clarity and better readability. 
2. In `stfcEmailHandler.ts`, several unnecessary imports, functions, and variables have been removed. This significantly simplifies the file, making it easier to understand and maintain. 
3. The email handling process has been switched to use status actions, which simplifies the process and reduces the potential for errors.

## Fixes
Part of : https://github.com/UserOfficeProject/issue-tracker/issues/1011
## Depends On
https://git.devs.facilities.rl.ac.uk/isisbusapps/BisAppSettings/pulls/453


bolmsten (approved):

bolmsten:
I think this is a good step. In the future it would be good to make the SMTP service less STFC specific, at the moment you have hardcoded values from STFC in the file.

bolmsten:
@gnyiri  

Would be good to do the same for ELI and combine the SMTP services

bashanlam (approved):

gnyiri:
> @gnyiri
> 
> Would be good to do the same for ELI and combine the SMTP services

We will probably not need our ELI specific SMTP service, the specialized event handler would be enough if the SMTP service will be generic without STFC specific code.

mutambaraf:
> @gnyiri
> 
> Would be good to do the same for ELI and combine the SMTP services

I think the best way to solve this is to allow upload of email templates then query the table for uploaded templates.

PR #714, opened 2024-08-22
fix: add validation to files endpoints

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1152

## Description
This PR introduces validation to the file endpoints to ensure only authorized users can upload or download files.

## Motivation and Context
Currently, there is no validation in the file endpoints which could potentially allow unauthorized access to sensitive data. This change is necessary to enhance the security of our application by ensuring that only authorized users can upload or download files.

## Changes
1. Added a validation middleware on the file endpoints that checks if the user is authenticated and if their external token is valid.
2. The frontend now sends an authorization header with the user's token when making a POST request to upload a file.
3. Implemented context to get the user's token in the file upload hook.

## Fixes Jira Issue


janosbabik (changes requested):
Thanks @TCMeldrum , for fixing this issue! Just one suggestion from me.

janosbabik (approved):

ellen-wright (approved):

gnyiri:
We tested upload/download at ELI and found that the validation prevents downloading files (401 error).

TCMeldrum:
Yeah sorry forgot to add the request header to the download links thought that kind of thing would have been picked up by the tests, working on a fix now.

PR #747, opened 2024-09-09
fix: resolve scientist proposal page next page button is greyed out and disabled problem #1160

bashanlam:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1160

This is a hotfix to resolve the problem of the next page button is greyed out and disabled, so scientists can only view 5, 10 or 20 proposals (using the "rows per page" dropdown) and cannot view more.

### DIFF
--- a/apps/frontend/src/components/proposal/ProposalTableInstrumentScientist.tsx
+++ b/apps/frontend/src/components/proposal/ProposalTableInstrumentScientist.tsx
@@ -82,6 +82,9 @@ type QueryParameters = {
   };
   searchText?: string | undefined;
 };
+
+const currentPage = 0;
+
 const getFilterReviewer = (selected: string | ReviewerFilter) =>
   selected === ReviewerFilter.ME ? ReviewerFilter.ME : ReviewerFilter.ALL;
 
@@ -257,7 +260,6 @@ const ProposalTableInstrumentScientist = ({
 }: {
   confirm: WithConfirmType;
 }) => {
-  const currentPage = 0;
   const [rowsPerPage, setRowsPerPage] = useState(20);
   const [selectedProposals, setSelectedProposals] = useState<
     ProposalViewData[]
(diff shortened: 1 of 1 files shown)


yoganandaness (changes requested):

apps/frontend/src/components/proposal/ProposalTableInstrumentScientist.tsx
```diff
@@ -257,7 +257,7 @@ const ProposalTableInstrumentScientist = ({
 }: {
   confirm: WithConfirmType;
 }) => {
-  const [currentPage, setCurrentPage] = useState(0);
+  const [currentPage] = useState(0);
```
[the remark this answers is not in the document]
bashanlam: Updated.

deepaksftc (approved):

apps/frontend/src/components/proposal/ProposalTableInstrumentScientist.tsx
```diff
@@ -257,7 +257,7 @@ const ProposalTableInstrumentScientist = ({
 }: {
   confirm: WithConfirmType;
 }) => {
-  const [currentPage, setCurrentPage] = useState(0);
+  const currentPage = 0;
```
yoganandaness, line 260: @bashanlam I have a suggestion. Could you check if it make sense. Since this variable doesn't have any dependency on the component now, can we just move it outside the component like this?

`const CURRENT_PAGE = 0` OR `const STARTING_PAGE = 0`. 

It doesn't have to be in the useEffect, as the value is not going to change at all.

yoganandaness (changes requested):

apps/frontend/src/components/proposal/ProposalTableInstrumentScientist.tsx
bashanlam: I've moved the currentPage variable outside the component while retaining its name. However, the page crashes when currentPage is undefined, suggesting it might be a mandatory prop for MaterialTable.

yoganandaness (approved):

PR #771, opened 2024-09-16
fix: instrument picker instrument selection

TCMeldrum:
## Description
This PR fixes the issue with instrument selection in the instrument picker.

## Motivation and Context
The change is required to correct the behavior of the instrument picker, which was not properly removing previously assigned instruments before assigning new ones. This was causing confusion and errors in instrument assignments.

## Changes
- Introduced a call to `removeProposalsFromInstrument` function in `InstrumentPicker.ts` to remove previously assigned instruments before new assignment.
- Adjusted the value format of the selection in `QuestionaryComponentInstrumentPicker.tsx` to ensure correct instrument selection in the picker.

## Fixes Jira Issue

### DIFF
--- a/apps/backend/src/models/questionTypes/InstrumentPicker.ts
+++ b/apps/backend/src/models/questionTypes/InstrumentPicker.ts
@@ -108,6 +108,15 @@ export const instrumentPickerDefinition: Question<DataType.INSTRUMENT_PICKER> =
         return;
       }
 
+      const instrumentDataSource = container.resolve<InstrumentDataSource>(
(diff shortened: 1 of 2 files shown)


TCMeldrum:
Oh I think your right missed that when I looked into it. Removing the length check seems to remove the instrument, tho I'm not sure what the length of the arrays is ment to do?

simonfernandes:
I'm not sure either what it's for, my only guess is it's trying to avoid removing the instruments if they haven't changed

PR #773, opened 2024-09-17
fix: instrument picker not removing instruments from proposal

TCMeldrum:
closes: UserOfficeProject/issue-tracker#1170

HOTFIX

## Description
This PR fixes an issue where the instrument picker was not properly removing instruments from a proposal.

## Motivation and Context
The current behavior of the instrument picker led to a problem where removed instruments were still appearing in the proposal, causing confusion and potential inaccuracies in the proposal data. This change is required to ensure that when an instrument is removed, it is reflected correctly in the proposal.

## Changes
- Added code to identify instruments on a proposal that need to be removed based on the updated selection in the instrument picker.
- Modified the condition to trigger the removal of instruments from a proposal, ensuring it occurs whenever there are instruments to remove.
- Adjusted the value assignment in the instrument picker component to handle single and multiple selections consistently, ensuring the correct data type is used.

### DIFF
--- a/apps/e2e/cypress/e2e/proposals.cy.ts
+++ b/apps/e2e/cypress/e2e/proposals.cy.ts
@@ -1542,6 +1542,7 @@ context('Proposal tests', () => {
       cy.contains('Proposals').click();
       cy.contains(title).parent().contains(instrument2.name);
       cy.contains(title).parent().find('[aria-label="View proposal"]').click();
+      cy.contains('td', instrument1.name).should('not.exist');
       cy.contains('td', instrument2.name).should('exist');
     });
 
--- a/apps/e2e/cypress/e2e/proposals.cy.ts
+++ b/apps/e2e/cypress/e2e/proposals.cy.ts
@@ -1542,7 +1542,7 @@ context('Proposal tests', () => {
       cy.contains('Proposals').click();
       cy.contains(title).parent().contains(instrument2.name);
       cy.contains(title).parent().find('[aria-label="View proposal"]').click();
-      cy.contains('td', instrument1.name).should('not.exist');
+      cy.contains('td', instrument.name).should('not.exist');
       cy.contains('td', instrument2.name).should('exist');
     });
 
--- a/apps/e2e/cypress/e2e/proposals.cy.ts
+++ b/apps/e2e/cypress/e2e/proposals.cy.ts
@@ -1490,6 +1490,61 @@ context('Proposal tests', () => {
       cy.contains('td', instrument.name).should('exist');
     });
 
+    it('Instrument should be automatically un assigned to the proposal when a new instrument is selected', () => {
+      cy.login('user1', initialDBData.roles.user);
+      cy.visit('/');
+      cy.contains('New Proposal').click();
+      cy.get('[data-cy=call-list]').find('li:first-child').click();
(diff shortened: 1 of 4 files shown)


apps/frontend/src/components/questionary/questionaryComponents/InstrumentPicker/QuestionaryComponentInstrumentPicker.tsx
```diff
@@ -216,7 +216,7 @@ export function QuestionaryComponentInstrumentPicker(
             value={
               Array.isArray(stateValue)
                 ? stateValue?.filter((i) => i).map((i) => i.instrumentId) || []
-                : stateValue?.instrumentId || '0'
+                : [stateValue?.instrumentId] || ['0']
```
janosbabik, line 219: this expression is always truthy
```suggestion
                : stateValue?.instrumentId ? [stateValue.instrumentId] : []
```

janosbabik (changes requested):
Hi @TCMeldrum, just a small recommendation from me. Also, it would be good to have an E2E test. What do you think?

TCMeldrum:
I agree there should be e2e tests, I would be happy to add them, that being said we have a call out at the moment and we are seeing this bug on prod at the moment so would like to get a fix out.

ACLay (approved):

simonfernandes:
Do we need to consider

TCMeldrum:
Oh I forgot about that, guess this needs to got out as part of a bigger release on our side. The breaking change seems to be a React update, I guess I'll delay this until we have discussed it.

janosbabik (approved):

apps/e2e/cypress/e2e/proposals.cy.ts
```diff
@@ -1490,6 +1490,61 @@ context('Proposal tests', () => {
...
+      cy.get('[data-cy="save-and-continue-button"]').focus().click();
+      cy.finishedLoading();
+      cy.notification({ variant: 'success', text: 'Saved' });
+
+      cy.login('officer');
+      cy.visit('/');
+      cy.contains('Proposals').click();
+      cy.contains(title).parent().contains(instrument2.name);
+      cy.contains(title).parent().find('[aria-label="View proposal"]').click();
+      cy.contains('td', instrument2.name).should('exist');
```
ACLay, line 1545: Should this also check that the first instrument isn't still there?

apps/e2e/cypress/e2e/proposals.cy.ts
```diff
@@ -1490,6 +1490,62 @@ context('Proposal tests', () => {
...
+      cy.get('[data-cy="save-and-continue-button"]').focus().click();
+      cy.finishedLoading();
+      cy.notification({ variant: 'success', text: 'Saved' });
+
+      cy.login('officer');
+      cy.visit('/');
+      cy.contains('Proposals').click();
+      cy.contains(title).parent().contains(instrument2.name);
+      cy.contains(title).parent().find('[aria-label="View proposal"]').click();
+      cy.contains('td', instrument1.name).should('not.exist');
```
ACLay, line 1545: Earlier in the test we used `instrument` instead of `instrument1`

```suggestion
      cy.contains('td', instrument.name).should('not.exist');
```

ACLay (approved):

PR #778, opened 2024-09-18
feat: send rabbitMQ messages when proposals move between faps

ACLay:
## Description

We have downstream software that manages a sharepoint datasource that needs informing when a proposal has been moved between access panels. The UOP event triggered by this wasn't being sent to rabbitMQ, so this PR adds it to the list of events that are.

## How Has This Been Tested

By writing the update for the downstream software that will consume the events and updating test proposals in my local environment. (https://github.com/isisbusapps/proposal-allocations/pull/577)

## Fixes

Fixes UserOfficeProject/issue-tracker#1174


TCMeldrum (approved):

janosbabik (approved):

martin-trajanovski:
@ACLay I was thinking if this is the right way of hardcoding the email sending or using the workflow to do it is much better. For example on status change to "PROPOSAL_FAPS_SELECTED" we can trigger email sending to configured people or message to RabbitMQ consumers as well. Let me know what do you think about this.

ACLay:
We don't have `PROPOSAL_FAPS_SELECTED` causing any status transitions in our workflows.

The case I'm trying to capture with this is that a user officer can manually move a proposal between FAPs, as some instruments can be reviewed in one of many FAPs. That moving can be done at any point in the workflow up to the reviews taking place, and doesn't affect their status, so I don't see how we could capture it practically in a workflow.

martin-trajanovski:
> We don't have `PROPOSAL_FAPS_SELECTED` causing any status transitions in our workflows.
> 
> The case I'm trying to capture with this is that a user officer can manually move a proposal between FAPs, as some instruments can be reviewed in one of many FAPs. That moving can be done at any point in the workflow up to the reviews taking place, and doesn't affect their status, so I don't see how we could capture it practically in a workflow.

Yes you are right then it is a bit hard to catch that as we don't have event based actions but only status. Maybe in the future it will be good to think about that and automate this by adding some event actions that will be executed if some event is triggered. But it is not that relevant for today as we already have some hardcoded events triggering some action.

PR #793, opened 2024-09-25
feat: add Xpress management page

simonfernandes:
## Description
This PR adds an Xpress management page for user officers and experiment scientists. It is hidden behind a feature flag so that it's only visible if enabled.

At the moment this is a basic table view that will show all proposals that are attached to techniques. User officers can see all proposals attached to techniques, experiment scientists can only see proposals on techniques that they are part of. Once we have the feature of being able to identify Xpress proposals via the call's workflow status (in development), we can optimise this further.

## Motivation and Context
The Xpress management page is a necessary addition to improve the proposal management process. The current system lacks an efficient way to manage, sort and filter proposals. The Xpress management page solves this issue by providing a dedicated interface for these tasks. 

## Changes
1. Implemented the Xpress management page feature, which is only visible if the associated feature flag is enabled.
2. Added a new 'submitted_date' column to the proposals table to keep track of when each proposal was submitted.
3. Created a new view 'proposal_table_view' to facilitate efficient querying of proposal data.
4. Added two new tables 'technique_has_scientists' and 'technique_has_proposals' to establish relationships between techniques, proposals, and scientists.
5. Updated the ProposalDataSource and TechniqueDataSource interfaces to include methods for retrieving proposals associated with a scientist and techniques associated with a proposal, respectively.

## How Has This Been Tested?
The changes have been tested by running the existing unit tests for the modified components. All tests passed, indicating that the changes do not break any existing functionality. Additionally, manual testing was conducted to verify the correct operation of the new Xpress management page.

## Fixes Jira Issue

Closes https://github.com/UserOfficeProject/issue-tracker/issues/1140

https://github.com/user-attachments/assets/b6d6b1f6-6c59-4bf4-8ef2-1501ea11d36b


Scott-James-Hurley:
Do you want the STFC-side reviewer to be from Submissions?

simonfernandes:
@Scott-James-Hurley That's probably a good idea, I'll change it

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,363 @@
...
+      customFilterAndSearch: () => true,
+    },
+  ];
+
+  addColumns(columns, instrumentManagementColumns(t));
+  addColumns(columns, techniquesColumns());
+
+  columns = setSortDirectionOnSortField(columns, sortField, sortDirection);
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mutambaraf, line 170: Should we be disabling linting rules.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,363 @@
...
+            options={{
+              search: true,
+              searchText: search || undefined,
+              selection: true,
+              headerSelectionProps: {
+                inputProps: { 'aria-label': 'Select All Rows' },
+              },
+              debounceInterval: 600,
+              columnsButton: true,
+              pageSize: 20,
```
mutambaraf, line 311: Are we not allowing page size changes here.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,363 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [selectedProposals, setSelectedProposals] = useState<
+    ProposalViewData[]
+  >([]);
+  const { techniques, loadingTechniques } = useTechniquesData();
+  const { proposalStatuses, loadingProposalStatuses } =
+    useProposalStatusesData();
+  const [currentPage, setCurrentPage] = useState(0);
+  const [rowsPerPage, setRowsPerPage] = useState(20);
+  const PREFETCH_SIZE = 200;
```
mutambaraf, line 32: Will this not be best be server side rather than use prefetching we can get slices based on page size

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,363 @@
...
+      offset: 0,
+    },
+    searchText: search ?? undefined,
+  });
+
+  const [proposalFilter, setProposalFilter] = useState<ProposalsFilter>({
+    callId: callId ? +callId : undefined,
+    instrumentFilter: {
+      instrumentId: instrument ? +instrument : null,
+      showAllProposals: !instrument,
```
mutambaraf, line 67: If l select a technique with multiple instrument will this filter all or just one.

apps/frontend/src/components/common/proposalFilters/TechniqueFilter.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const { t } = useTranslation();
+
+  if (techniques === undefined) {
+    return null;
+  }
+
+  /**
+   * NOTE: We might use https://material-ui.com/components/autocomplete/.
+   * If we have lot of dropdown options to be able to search.
+   */
```
mutambaraf, line 48: Is these relevant in the current context if it is we can use autocomplete then know its sorted.

apps/frontend/src/components/common/proposalFilters/TechniqueFilter.tsx
```diff
@@ -0,0 +1,124 @@
...
+              <MenuItem value={TechniqueFilterEnum.MULTI}>Multiple</MenuItem>
+            )}
+            <ListSubheader sx={{ lineHeight: 1 }}>
+              <Divider>Techniques</Divider>
+            </ListSubheader>
+            {techniques.map((technique) => (
+              <MenuItem key={technique.id} value={technique.id}>
+                {technique.name}
+              </MenuItem>
+            ))}
```
mutambaraf, line 116: The list sub header should be shown if there are techniques here.

apps/frontend/src/components/xpress/XpressProposalFilterBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+  InstrumentFragment,
+  ProposalsFilter,
+  ProposalStatus,
+  TechniqueFragment,
+} from 'generated/sdk';
+
+type ProposalFilterBarProps = {
+  calls?: { data: Call[]; isLoading: boolean };
+  instruments?: { data: InstrumentFragment[]; isLoading: boolean };
+  techniques?: { data: TechniqueFragment[]; isLoading: boolean };
```
mutambaraf, line 21: Do not techniques have related instruments .

apps/frontend/src/components/xpress/XpressProposalFilterBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+    data: PropTypes.array.isRequired,
+    isLoading: PropTypes.bool.isRequired,
+  }),
+  techniques: PropTypes.shape({
+    data: PropTypes.array.isRequired,
+    isLoading: PropTypes.bool.isRequired,
+  }),
+  setProposalFilter: PropTypes.func.isRequired,
+  filter: PropTypes.object.isRequired,
+};
```
mutambaraf, line 138: If we have already set types on the function do we still need these

bolmsten:
Is experimental area the same as instrument?

simonfernandes:
@bolmsten Yes, it's our translation for it

apps/frontend/src/components/xpress/XpressProposalTable.tsx
deepaksftc: Hope I understood the question correctly. If a technique alone is selected in technique filter, it will check for proposal which has that particular technique attached, independent of instruments in technique. 
If any instrument is selected from instrument filter, it will now check for the technique to which that particular instrument is attached and return all proposal which has that particular technique.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
deepaksftc: 20 which is set here will be the default size. We do can change the page size/the number of rows from UI.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,363 @@
...
+      showMultiTechniqueProposals: false,
+    },
+    dateFilter: {
+      to: to ? to : null,
+      from: from ? from : null,
+    },
+    referenceNumbers: proposalId ? [proposalId] : undefined,
+    proposalStatusId: proposalStatusId ? +proposalStatusId : undefined,
+    text: search,
+    excludeProposalStatusIds: [9],
```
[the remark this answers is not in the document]
deepaksftc: This is for filtering out the expired proposals while displaying the proposals for technique/instrument scientists. Other xpress proposals are expected to have only the xpress specific statuses.

apps/frontend/src/components/xpress/XpressProposalFilterBar.tsx
deepaksftc: Yes, techniques have related instruments. We pass only the instruments that are part of the techniques, to be listed in filter dropdown.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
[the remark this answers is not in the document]
mutambaraf: For the proposal status we might need to only show those related to Xpress proposals.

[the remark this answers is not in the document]
deepaksftc: Yes, we should only show xpress statuses, this will be handled as part of upcoming issue for xpress status implementation (Once the statuses are confirmed).

apps/frontend/src/components/xpress/XpressProposalTable.tsx
deepaksftc: @mutambaraf, yes we can do it. Shall we handle it as part of another issue, given that this won't cause any issue now and considering the target date?

mutambaraf: I think this is just passing two additional filters to the backend.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
mutambaraf: How are we then updating it when the use select a new page size.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
deepaksftc: Sure, will include it.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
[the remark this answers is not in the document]
mutambaraf: I thinks we have some which are relevant and then we can add the additional as they are decided. I can see we have some which are being included and may not be relevant to these proposals.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+  const [tableData, setTableData] = useState<ProposalViewData[]>([]);
+  const [preselectedProposalsData, setPreselectedProposalsData] = useState<
+    ProposalViewData[]
+  >([]);
+
+  const { instruments, loadingInstruments } = useXpressInstrumentsData(
+    proposalsData,
+    techniques
+  );
```
mutambaraf, line 188: Is this query also necessary because if we already have techniques and calls with instruments we can still get instruments related to the technique and call or filter them with calls on proposals.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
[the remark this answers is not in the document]
simonfernandes: I'm currently working on this issue to be able to change statuses and exclude the ones which aren't applicable so it will be changed very soon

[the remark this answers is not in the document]
mutambaraf: Can we have a comment on this if it will be fixed through another issue and we can also include the issue number please.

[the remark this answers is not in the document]
deepaksftc: Status related logics will modified/fixed as part of https://github.com/UserOfficeProject/issue-tracker/issues/1138.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,430 @@
...
+
+    if (isMounted) {
+      refreshTableData();
+    }
+
+    return () => {
+      isMounted = false;
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [JSON.stringify(proposalFilter)]);
```
mutambaraf, line 98: We can also create a function that we pass to XpressProposalFilterBar component that will update our setProposalFilter and update the table.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,389 @@
...
+
+const XpressProposalTable = () => {
+  const tableRef = React.useRef<MaterialTableCore<ProposalViewData>>();
+  const refreshTableData = () => {
+    tableRef.current?.onQueryChange({});
+  };
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [selectedProposals, setSelectedProposals] = useState<
+    ProposalViewData[]
+  >([]);
```
mutambaraf, line 41: We can try to not disable this linting rule if possible.

apps/frontend/src/components/xpress/XpressProposalTable.tsx
```diff
@@ -0,0 +1,389 @@
...
+            techniques={{
+              data: techniques,
+              isLoading: loadingTechniques,
+            }}
+            proposalStatuses={{
+              data: proposalStatuses,
+              isLoading: loadingProposalStatuses,
+            }}
+            handleFilterChange={handleFilterChange}
+            filter={proposalFilter}
```
mutambaraf, line 331: We can also move filter into to handleFilterChange function and not pass it if possible since it is not being used for any computations in XpressProposalFilterBar. Looks good just those last comments.

deepaksftc: @mutambaraf, sorry I didn't understand it properly. 'filter' holds the currently selected values of each filters right. We would need it to know the current filter values inside XpressProposalFilterBar.

mutambaraf (approved):

yoganandaness (changes requested):
I have re-reviewed the PR. Some of the comments, i have made previously have been removed after thoroughly going through the requirement. Now, i have made few comments and hope that is going to make sense. Requesting to take a look. 

@simonfernandes 
@deepaksftc

deepaksftc:
> I have re-reviewed the PR. Some of the comments, i have made previously have been removed after thoroughly going through the requirement. Now, i have made few comments and hope that is going to make sense. Requesting to take a look.
> 
> @simonfernandes @deepaksftc

@yoganandaness, thanks for the comments. We have made the changes.

yoganandaness (approved):
Looks good to me.

PR #797, opened 2024-09-27
fix: hide multiple download option when only one proposal is checked in the table

yoganandaness:
## Description

1. When only one proposal is selected in the table, in the toolbar, dont show multiple download as an option. Show only single download. 
2. When more than on proposal has been selected in the table, in the toolbar, show multiple download as an option along with single download as an option

## Motivation and Context

Improve UX

## How Has This Been Tested

Attached e2e

- [X] I have added tests to cover my changes.


apps/frontend/src/components/proposal/TableActionsDropdownMenu.tsx
```diff
@@ -112,6 +118,18 @@ const TableActionsDropdownMenu = ({
...
+  const [searchParams] = useSearchParams();
+  const { search } = useLocation();
+
+  useEffect(() => {
+    if (searchParams.getAll('selection').length <= 1) {
+      setMenuItems(getMenuItems(false));
+    } else {
+      setMenuItems(getMenuItems());
+    }
+  }, [search, searchParams]);
```
janosbabik, line 131:
```suggestion
  }, [searchParams]);
```

apps/frontend/src/components/proposal/TableActionsDropdownMenu.tsx
```diff
@@ -112,6 +118,18 @@ const TableActionsDropdownMenu = ({
     null
   );
   const [open, setOpen] = React.useState<boolean>(false);
+  const [menuItems, setMenuItems] = React.useState(getMenuItems());
+  const [searchParams] = useSearchParams();
+  const { search } = useLocation();
```
janosbabik, line 123: Do you think this is necessary?

janosbabik (changes requested):
Hi @yoganandaness, nice improvement! I have just one question: Do you think it's necessary to use `useLocation`?

apps/frontend/src/components/proposal/TableActionsDropdownMenu.tsx
yoganandaness: Yes. The reason is, 

When there is a change in the query params caused by different components, the current component doesnt know about it. The `searchParams` from `useSearchParams` does not reinstantiate and hence, we can't use it in `useEffect`, but the `search` from `useLocation` does side effect. 

https://reactrouter.com/en/main/hooks/use-location

@janosbabik

janosbabik (approved):

TCMeldrum (approved):

PR #802, opened 2024-10-02
fix: insert all available user information at creation time

gnyiri:
## Description

Some user attributes are only updated after the second login of the user (first login will create the user in the users table, the second login will update the existing record).

## Motivation and Context

It is confusing that some user attributes will only be set after the second login. 

## How Has This Been Tested

Manual testing + e2e.

## Fixes

Synchronize the inserted / updated user attributes. 

## Changes

Update upsert method in OAuthAuthorization.ts.

## Depends on

N/A

- [ x] All relevant doc has been updated


bolmsten (approved):
This is a good improvement. We should go back and clean up the user table from all the unused values here

deepaksftc (approved):

PR #803, opened 2024-10-03
fix: user search for instrument scientist and internal reviewer

gnyiri:
## Description

Add missing user role filters when searching for internal reviewers / instrument scientists. 

## Motivation and Context

It is possible to add internal reviewer / instrument scientist for a proposal / instrument without having the proper role. 

## How Has This Been Tested

Manual test & e2e

## Fixes

Add missing roles for the user lookup.

## Changes

Code changes in frontend.

## Depends on

N/A

- [ x] All relevant doc has been updated

### DIFF
--- a/apps/frontend/src/components/instrument/CreateUpdateInstrument.tsx
+++ b/apps/frontend/src/components/instrument/CreateUpdateInstrument.tsx
@@ -75,7 +75,12 @@ const CreateUpdateInstrument = ({
         .getUsers({ filter: value, userRole: UserRole.INSTRUMENT_SCIENTIST })
         .then((data) => {
           if (data.users?.totalCount == 0) {
-            setFieldError('surname', 'No users found with that surname');
+            setFieldError(
+              'surname',
+              'No ' +
+                i18n.format(t('instrumentSci'), 'plural') +
+                ' found with that surname'
+            );
           } else {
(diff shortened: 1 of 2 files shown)


TCMeldrum (approved):

apps/frontend/src/components/instrument/CreateUpdateInstrument.tsx
```diff
@@ -72,7 +72,7 @@ const CreateUpdateInstrument = ({
 
     try {
       await api()
-        .getUsers({ filter: value })
+        .getUsers({ filter: value, userRole: UserRole.INSTRUMENT_SCIENTIST })
         .then((data) => {
           if (data.users?.totalCount == 0) {
             setFieldError('surname', 'No users found with that surname');
```
ACLay, line 78: Could this error message be updated to reflect the filter change? There's a similar error message in the other updated file that I think should be updated too.
```suggestion
            setFieldError('surname', 'No instrument scientists found with that surname');
```

ACLay: Actually, I think this might need to use the `instrumentSci` field from the translation file to be accurate at each facility.

ACLay (approved):

PR #811, opened 2024-10-09
fix: review uows caches

Bhaswati1148:
## Motivation and Context
The UOWS is seeing large volumes of traffic from UOP, potentially causing errors. It appears that our caching is insufficient for the number of users being processed, as the number of requested users is massively higher than the cache size, so a lot of refetches are performed.
Custom implementation of LRUCache algorithm was used until this PR. This PR aims to replace that custom implementation with external library TTLCache https://www.npmjs.com/package/@isaacs/ttlcache

## Fixes
https://github.com/UserOfficeProject/issue-tracker/issues/1099

## Changes
Switching over from LRUCache to TTLCache https://www.npmjs.com/package/@isaacs/ttlcache 


ACLay:
Would it be worth adding some environment variables to use as override values for the default cache size and ttl parameters? That way if we need to increase the sizes we wouldn't need to do another release.

apps/backend/src/utils/Cache.ts
```diff
@@ -23,8 +23,15 @@ export class Cache<T> implements ICache<T> {
...
-    this.millisecondsToLive = secondsToLive ? secondsToLive * 1000 : 10000;
+    this.maxEntries =
+      Number(process.env.CACHE_MAX_ENTRIES) > maxEntries
+        ? Number(process.env.CACHE_MAX_ENTRIES)
+        : maxEntries;
+    this.millisecondsToLive = secondsToLive
+      ? Number(process.env.CACHE_TTL) > secondsToLive
+        ? Number(process.env.CACHE_TTL) * 1000
+        : secondsToLive * 1000
+      : 10000;
```
ACLay, line 34: It feels odd to me to have override parameters that will only be used if they're larger. And this layout also means that the CACHE_TTL variable will be ignored if the constructor is called without a secondsToLive parameter. I'd have thought the logic might go something like `this.millisecondsToLive = (Number(process.env.CACHE_TTL ?? secondsToLive ?? 10)) * 1000;`

And could I be inconvenient and ask that CACHE_TTL be renamed to include SECONDS so it's immediately clear in the config what units it uses

PR #820, opened 2024-10-14
fix: prevent multiple instances of the StfcUserDataSource being created

ACLay:
## Motivation and Context

In STFC we use an external service to get user details, and cache results to improve performance. I noticed that we're logging statistics for multiple parallel instances of these, which means we're not getting the full benefit of those caches, as different parts of the app may be requesting the same data from the slow external source instead of using already requested and cached data.

## How Has This Been Tested

I browsed to local pages where the updated data sources would be used and they seemed to load fine. Other than that I'm just relying on the test suite to check for regressions.

## Fixes

Closes UserOfficeProject/issue-tracker#1188

### DIFF
--- a/apps/backend/src/datasources/stfc/StfcProposalDataSource.ts
+++ b/apps/backend/src/datasources/stfc/StfcProposalDataSource.ts
@@ -1,5 +1,6 @@
-import { injectable } from 'tsyringe';
+import { container, injectable } from 'tsyringe';
 
+import { Tokens } from '../../config/Tokens';
 import { Call } from '../../models/Call';
 import { Proposal } from '../../models/Proposal';
 import { ProposalView } from '../../models/ProposalView';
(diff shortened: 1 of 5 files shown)


apps/backend/src/datasources/stfc/StfcProposalDataSource.ts
```diff
@@ -19,10 +20,12 @@ import { ProposalsFilter } from './../../resolvers/queries/ProposalsQuery';
 import PostgresProposalDataSource from './../postgres/ProposalDataSource';
 import { StfcUserDataSource } from './StfcUserDataSource';
 
-const stfcUserDataSource = new StfcUserDataSource();
-
 @injectable()
 export default class StfcProposalDataSource extends PostgresProposalDataSource {
+  protected stfcUserDataSource: StfcUserDataSource = container.resolve(
+    Tokens.UserDataSource
+  ) as StfcUserDataSource;
```
mutambaraf, line 27: We can also inject it into the constructor.

ACLay: I could, but many of our data sources don't have an explicit constructor. If I were to implement one with the `@inject` annotation I'd have to also add annotations and parameters for the ones we'd then need to pass to the parent classes constructor, and it ends up being a fair bit more code than this.

mutambaraf (approved):

janosbabik (approved):

PR #823, opened 2024-10-16
perf: reduce data quantity from high use getCalls query

ACLay:
## Description

This is the second most used query in our monitoring logs, and for most places where it's used the vast majority of fields are unnecessary, especially user data, which for us comes from a slow external service.

## Motivation and Context

We're approaching a proposal call deadline and are seeing high load levels. This may be a significant culprit.


TCMeldrum (approved):

Bhaswati1148 (approved):

gnyiri:
Its OK, but it could be a better (?) choice to add a new parameter to the existing useCall function instead of introducing a new useCalllsMinimalData function with almost identical implementation.   

Example for this: includeTemplate parameter in useActiveTemplates.

gnyiri (commented):
There are many references of useCallsData in the code. As MINIMAL is the default parameter,  did you check that CallsTable is the only reference where all call data are needed?

ACLay:
Yeah, I followed all the places where useCallsData is used, and the only place that looked like it needed extra fields was the calls table.

janosbabik (approved):

gnyiri (approved):

ACLay:
> Yeah, I followed all the places where useCallsData is used, and the only place that looked like it needed extra fields was the calls table.

Looks like xpress stuff that merged in since I did those checks also required extended data

TCMeldrum (approved):

PR #824, opened 2024-10-18
feat: assign instruments to Xpress proposals

deepaksftc:
## Description
This pull request introduces the feature of assigning instruments to Xpress proposals.

## Motivation and Context
This change was required to provide the ability for users to assign specific instruments to Xpress proposals, enhancing the granularity and specificity of proposal handling.

## Changes
1. Added a `checkProposalsHasTechniques` method in `TechniqueDataSource`, `TechniqueDataSourceMock`, and `PostgresTechniqueDataSource` to verify if proposals have techniques.
2. Implemented a new mutation `assignXpressProposalsToInstruments` in `InstrumentMutations` and `AssignProposalsToInstrumentsMutation` to assign instruments to proposals.
3. Added a check in `AssignProposalsToInstrumentsMutation` to validate if the proposals are Xpress before assigning instruments.
4. Added e2e tests in `xpress.cy.ts` to verify the new functionality.
5. Updated the frontend component `XpressProposalTable.tsx` to include the new feature of assigning instruments to proposals.

## Fixes Issue

https://github.com/UserOfficeProject/issue-tracker/issues/1137

image

### DIFF
--- a/apps/backend/src/datasources/TechniqueDataSource.ts
+++ b/apps/backend/src/datasources/TechniqueDataSource.ts
@@ -37,8 +37,4 @@ export interface TechniqueDataSource {
     techniqueId: number
   ): Promise<boolean>;
   getTechniquesByProposalPk(proposalPk: number): Promise<Technique[]>;
-  isXpressInstrumentAndProposal(
-    proposalPk: number,
-    instrumentId: number
-  ): Promise<boolean>;
 }
--- a/apps/backend/src/datasources/TechniqueDataSource.ts
+++ b/apps/backend/src/datasources/TechniqueDataSource.ts
@@ -37,4 +37,5 @@ export interface TechniqueDataSource {
     techniqueId: number
   ): Promise<boolean>;
   getTechniquesByProposalPk(proposalPk: number): Promise<Technique[]>;
+  checkProposalsHasTechniques(proposalPks: number[]): Promise<boolean>;
 }
--- a/apps/backend/src/datasources/TechniqueDataSource.ts
+++ b/apps/backend/src/datasources/TechniqueDataSource.ts
@@ -37,5 +37,8 @@ export interface TechniqueDataSource {
     techniqueId: number
   ): Promise<boolean>;
   getTechniquesByProposalPk(proposalPk: number): Promise<Technique[]>;
-  checkProposalsHasTechniques(proposalPks: number[]): Promise<boolean>;
+  isXpressInstrumentAndProposal(
+    proposalPk: number,
+    instrumentId: number
+  ): Promise<boolean>;
 }
--- a/apps/backend/src/mutations/InstrumentMutations.ts
+++ b/apps/backend/src/mutations/InstrumentMutations.ts
@@ -473,15 +473,11 @@ export default class InstrumentMutations {
         techniquesWithProposal.map((technique) => technique.id)
       );
 
-    let isXpress = false;
-
-    if (instrumentWithTechnique && instrumentWithTechnique.length > 0) {
-      isXpress = instrumentWithTechnique.find(
-        (instruments) => instruments.id === args.instrumentIds[0]
-      )
-        ? true
-        : false;
-    }
+    const isXpress = instrumentWithTechnique.find(
+      (instruments) => instruments.id === args.instrumentIds[0]
+    )
+      ? true
+      : false;
 
     if (!isXpress) {
       return rejection('No permission to assign instrument for this proposal', {
--- a/apps/backend/src/mutations/InstrumentMutations.ts
+++ b/apps/backend/src/mutations/InstrumentMutations.ts
@@ -14,10 +14,11 @@ import { Tokens } from '../config/Tokens';
 import { FapDataSource } from '../datasources/FapDataSource';
 import { InstrumentDataSource } from '../datasources/InstrumentDataSource';
 import { ReviewDataSource } from '../datasources/ReviewDataSource';
+import { TechniqueDataSource } from '../datasources/TechniqueDataSource';
 import { Authorized, EventBus, ValidateArgs } from '../decorators';
 import { Event } from '../events/event.enum';
 import { Instrument, InstrumentsHasProposals } from '../models/Instrument';
-import { rejection, Rejection } from '../models/Rejection';
+import { isRejection, rejection, Rejection } from '../models/Rejection';
 import { Roles } from '../models/Role';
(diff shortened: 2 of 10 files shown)


apps/e2e/cypress/e2e/xpress.cy.ts
```diff
@@ -839,5 +839,66 @@ context('Xpress tests', () => {
...
+      cy.get('table.MuiTable-root tbody tr').should(
+        'not.contain',
+        instrument1.name
+      );
+      cy.get('[role="listbox"]').contains(instrument1.name).click();
+      cy.finishedLoading();
+      cy.contains(instrument1.name);
+    });
+
+    it('Instrument scientist able to select, assign and unassign an instrument for a proposal', function () {
```
simonfernandes, line 872: We can change the name now that there is no ability to unassign

apps/backend/src/datasources/TechniqueDataSource.ts
```diff
@@ -37,5 +37,8 @@ export interface TechniqueDataSource {
     techniqueId: number
   ): Promise<boolean>;
   getTechniquesByProposalPk(proposalPk: number): Promise<Technique[]>;
-  checkProposalsHasTechniques(proposalPks: number[]): Promise<boolean>;
+  isXpressInstrumentAndProposal(
```
simonfernandes, line 40: Might be better to split the Xpress check to the STFC-specific mutation and keep the datasource to checking if the proposal belongs to the technique. It's a bit less optimised but then keeps the datasource generic and free of Xpress logic. Happy if you think it's best left though

simonfernandes (approved):

apps/backend/src/resolvers/mutations/AssignProposalsToInstrumentsMutation.ts
```diff
@@ -67,4 +67,51 @@ export class AssignProposalsToInstrumentsMutation {
...
+    @Ctx() context: ResolverContext
+  ) {
+    const techniquesWithProposal =
+      await context.queries.technique.getTechniquesByProposalPk(
+        context.user,
+        args.proposalPks[0]
+      );
+
+    if (!techniquesWithProposal || techniquesWithProposal.length < 1) {
+      return false;
```
simonfernandes, line 83: I think it would be better to return a Rejection instead of false so that it's clearer when validation has failed and we can provide some context as to why.

apps/backend/src/resolvers/mutations/AssignProposalsToInstrumentsMutation.ts
```diff
@@ -67,4 +67,52 @@ export class AssignProposalsToInstrumentsMutation {
...
+    @Ctx() context: ResolverContext
+  ) {
+    const techniquesWithProposal =
+      await context.queries.technique.getTechniquesByProposalPk(
+        context.user,
+        args.proposalPks[0]
+      );
+
+    if (!techniquesWithProposal || techniquesWithProposal.length < 1) {
+      return rejection(
```
simonfernandes, line 83: Would be good to add the user and args as context for these rejections

apps/backend/src/resolvers/mutations/AssignProposalsToInstrumentsMutation.ts
```diff
@@ -67,4 +67,52 @@ export class AssignProposalsToInstrumentsMutation {
       args
     );
   }
+
+  @Mutation(() => Boolean)
+  async assignXpressProposalsToInstruments(
```
simonfernandes, line 72: I had another look and I think this might be the only resolver with validation, everywhere else does it in the mutation. From a technical standpoint it seems it can be done either way and this looks good, but I think we should ideally follow our existing convention

apps/backend/src/mutations/InstrumentMutations.ts
```diff
@@ -450,6 +453,45 @@ export default class InstrumentMutations {
...
+    }
+
+    const instrumentWithTechnique =
+      await this.techniqueDataSource.getInstrumentsByTechniqueIds(
+        techniquesWithProposal.map((technique) => technique.id)
+      );
+
+    let isXpress = false;
+
+    if (!isRejection(instrumentWithTechnique)) {
```
simonfernandes, line 478: I don't think this can return a rejection, only an empty instruments array

apps/backend/src/mutations/InstrumentMutations.ts
```diff
@@ -450,6 +453,45 @@ export default class InstrumentMutations {
...
+      isXpress =
+        instrumentWithTechnique.length > 0 &&
+        instrumentWithTechnique.filter(
+          (instruments) => instruments.id === args.instrumentIds[0]
+        ).length > 0;
+    } else {
+      return instrumentWithTechnique;
```
simonfernandes, line 485: Similarly here I don't think this is going to return a rejection

apps/backend/src/mutations/InstrumentMutations.ts
```diff
@@ -450,6 +453,45 @@ export default class InstrumentMutations {
...
+      await this.techniqueDataSource.getInstrumentsByTechniqueIds(
+        techniquesWithProposal.map((technique) => technique.id)
+      );
+
+    let isXpress = false;
+
+    if (!isRejection(instrumentWithTechnique)) {
+      isXpress =
+        instrumentWithTechnique.length > 0 &&
+        instrumentWithTechnique.filter(
```
simonfernandes, line 481: array.find() may be more suitable here

apps/backend/src/mutations/InstrumentMutations.ts
```diff
@@ -475,14 +475,12 @@ export default class InstrumentMutations {
...
 
-    if (!isRejection(instrumentWithTechnique)) {
-      isXpress =
-        instrumentWithTechnique.length > 0 &&
-        instrumentWithTechnique.filter(
-          (instruments) => instruments.id === args.instrumentIds[0]
-        ).length > 0;
-    } else {
-      return instrumentWithTechnique;
+    if (instrumentWithTechnique && instrumentWithTechnique.length > 0) {
```
simonfernandes, line 478: Looks good, though this check shouldn't be needed as it will be an array of instruments or empty array. Using find() on an empty array will still work without error

simonfernandes (approved):

deepaksftc:
Hi @janosbabik, could you please have a look at this PR and suggest if there is any changes required.

janosbabik (approved):

PR #831, opened 2024-10-22
perf: cache stfc user service requests instead of responses

ACLay:
## Description

Instead of caching responses from the STFC users service, cache the requests. This way if a piece of external data is being requested, and additional requests for it come in while that first one is being made, we don't have to make repeated requests to the external service.

When loading the user officer homepage with empty caches in my local setup, this reduced the number of queries made to get user data from 155 to 56.

I also removed an unnecessary & unused initialisation of the STFC users datsource, which caused unnecessary duplicate cache statistics to be logged. I believe this must have come from merge conflict resolution some time after I got rid of them in #820

## Motivation and Context

We use a comparitively slow external service to provide user data, and cache responses to improve performance. However, if a piece of data has expired from the caches, it can potentially be requested in multiple places for one or several queries at the same time, resulting in us making unnecessary requests at the same time for the same data. During periods of high load we've often made more simultaneous requests than the external service can handle, resulting in degraded performance or requests failing altogether.

## How Has This Been Tested

A bit of general browsing the site locally, and running queries against logs to check that query loads on the external service are reduced. Otherwise I'm just relying on the test suite in this PR since it's internal refactoring.

## Fixes

Part of https://github.com/UserOfficeProject/issue-tracker/issues/1197

### DIFF
--- a/apps/backend/src/datasources/stfc/StfcUserDataSource.ts
+++ b/apps/backend/src/datasources/stfc/StfcUserDataSource.ts
@@ -204,7 +204,7 @@ export class StfcUserDataSource implements UserDataSource {
   private async getStfcBasicPersonByEmail(
     email: string,
     searchableOnly?: boolean
-  ): Promise<StfcBasicPersonDetails | null> {
+  ): Promise<StfcBasicPersonDetails | undefined> {
     const cache = searchableOnly
       ? this.uowsSearchableBasicUserDetailsCache
       : this.uowsBasicUserDetailsCache;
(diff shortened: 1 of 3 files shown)


ACLay:
While sharing requests to the UOWS should reduce peak loads, it could have impacts on performance in other ways. If some of those shared queries are anomalously slow then any users trying to simultaneously use the result of that request would be impacted instead of just one. I think it would be worth the risk though if it can reduce the peak request rates and UOWS overloading, and we can always undo this if it turns out to be more hinderance than help.

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -143,66 +146,92 @@ export class StfcUserDataSource implements UserDataSource {
...
-  ): Promise<StfcBasicPersonDetails | null> {
+  ): Promise<StfcBasicPersonDetails | undefined> {
     const cache = searchableOnly
       ? this.uowsSearchableBasicUserDetailsCache
       : this.uowsBasicUserDetailsCache;
 
-    const cachedUser = cache.get(email);
+    const cachedUser = await cache.get(email);
     if (cachedUser) {
       return cachedUser;
```
TCMeldrum, line 213: If I am understanding this correctly, then when since it cacheing the request and then returns this request. If that requests fails then are we not going to return a failed request until the cache times out? Sorry if I am miss understanding

ACLay: I was going for if the request fails, then resolving the cached Promise should return `undefined`, this return statement will be skipped, and we make a new request.

TCMeldrum: Cool thanks just wanted to clarify that!

TCMeldrum (approved):

gnyiri:
Did you consider an event-driven approach in the long term where UOS gets notified by the source (uOWS/Keycloak/..) whenever a user has changed his/her profile? I think that the users will rarely change their registration data during a call and most of the cache misses due to expiration are unnecessary.

ACLay:
That'd be nice, but our users service doesn't have any sort of notification system

simonfernandes (approved):

PR #832, opened 2024-10-28
feat: add client identifying metadata for apollo studio analytics

ACLay:
## Description

Adds a metadata header to graphql requests that names the client, allowing requests to be broken down by client in the apollo studio analytics suite. ([apollo docs](https://www.apollographql.com/docs/graphos/platform/insights/client-segmentation#setup))

My only concern is that this would mean we're sending user numbers to the external analytics suite. I don't think this would be a problem, but I'd like sign off from @srconway as to whether that passes GDPR muster on our side.

## Motivation and Context

This will allow us to tell which clients are producing the highest volume or longest running requests, allowing us to more easily locate bugs or targets for optimisation.

By including the user number where available my hope is that we can identify future occurances of https://github.com/UserOfficeProject/issue-tracker/issues/1094 with much greater ease too.

## How Has This Been Tested

Using local browser dev tools to verify that the headers are set on requests to the backend.

## Fixes

Closes UserOfficeProject/issue-tracker#1203


srconway:
I've not looked at the code itself but on the principle of whether we can use the user ID I think it's covered by our privacy statement under "to enable administration and delivery of the service". We do specifically mention tracking visitors through Google Analytics and we ought really to look at re-wording that statement to make it a bit broader - "3rd party systems like Google Analytics". Could you raise a separate issue to look at that one but given the first statement I don't think it should prevent us adding this in to Apollo as am I right in saying that only holds the data for 12 months.

> Your information may be shared with other organizations outside of STFC, only to enable administration and delivery of the services you requested or for the lawful, disclosed purposes of STFC. For example we will share information with taxis, hotels, airlines etc. as is necessary to deliver the travel and accommodation services your requested; we will share proposal details with organisations outside of STFC for proposal review or other purposes as part of internationally-funded programmes.
> 
> ...
> We use cookies on our website:

> ...
> to track visitor use of our websites and to compile statistical reports on website activity to help us improve our website to make it easier for you to access the services you require.
> ...
> We use Google Analytics to track visitor use of our websites. We have set the data retention period to the minimum available (14 months).

janosbabik:
Hi @ACLay, did you read my message on MTeam that I sent Friday morning? Did you try the approach where we just update the request's header after the auth middleware? If that works, wouldn’t it be simpler?

ACLay:
Yes, I saw the message. I'll give that a try, just had a lot of meetings on friday, and we've finally somehow managed to replicate https://github.com/UserOfficeProject/issue-tracker/issues/1094 so haven't had a chance to work on this.

ACLay:
@janosbabik I've had a look at setting it in the backend. It can be done there but you lose some level of specificity for anonymous calls. Setting it in the frontend we can specify a value like `UOP - anonymous`, but if we set something like that in the backend any anonymous calls from other consumers or systems would end up in the same category. It also just feels kinda hacky to be setting a request header once we've recieved the request.

janosbabik:
@ACLay I forgot that you have this identification from other systems. Okay, then the frontend should send it.

apps/frontend/src/hooks/common/useDataApi.ts
```diff
@@ -15,6 +15,26 @@ import { RequestQuery } from 'utils/utilTypes';
...
+  userId?: number | undefined
+) {
+  if (includeUserId) {
+    return (client: GraphQLClient) =>
+      client.setHeader(
+        clientNameHeader,
+        userId ? `UOP frontend - user ${userId}` : anonymousClientName
+      );
+  } else {
+    return setAnonymousClientNameHeader;
```
janosbabik, line 34: I would not add anything to the header if the feature is turned off. Could we use this?:
```suggestion
    return (client) => client;
```

janosbabik (changes requested):
Hi @ACLay ,it looks good! Just one small change from my side: can we remove the additional header item when the feature is turned off?

apps/frontend/src/hooks/common/useDataApi.ts
ACLay: Sorry, I thought the feature you were after was wether to include user numbers when available, and if the flag's not set always use the anonymous header.

I think that'd put us in a similar situation to setting the header on the backend. When determining whether to set the header based on a feature flag we need to make at least that one API request to get those feature flags. Not setting the anonymous header by default would mean that there's still unidentified requests being made that can't be attributed to the frontend in the analytics.

ACLay: @janosbabik I'd really like to have all our requests identified in the analytics where possible. If the default header was just "UOP frontend", instead of "UOP frontend - anonymous" would that be acceptable?

janosbabik: @ACLay Yes, I can work with this, but I have one last thought. I'm not sticking to the backend (BE) solution, but there are a few things to keep in mind:
With this approach, the server can receive any header values without validation, meaning I could send requests on behalf of other users or add custom messages. It’s not a huge issue, but it doesn’t feel thrustworthy.
If we handle this on the backend, you’d be able to control and validate incoming requests using user tokens and API tokens, which would cover all cases. Also with DI it can be turned off. 
I know that modifying request headers can seem hacky, but the frontend (FE) solution is a bit unconventional too.

ACLay: How about I always set an anonymous "UOP frontend" header value in the frontend, then in the backend, if the header value starts with that I append user data from the token when available? That way we identify all calls, without being entirely trusting of data that a users browser could manipulate

Scott-James-Hurley (approved):

janosbabik (approved):

PR #839, opened 2024-11-01
perf: stop overfetching of instrument data

TCMeldrum:
Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1223

## Description
This PR optimizes the fetching of instrument data by introducing a minimal version of instrument data fetching. 

## Motivation and Context
This change is required to improve the performance of the application by minimizing the amount of data fetched when full instrument data is not necessary.

## Changes
1. Introduced a minimal version of instrument data fetching in various components where full instrument data was previously fetched.
2. Updated the types and usages of instrument data in various components to use the minimal version.
3. Replaced the calls to `useInstrumentsData()` with `useInstrumentsMinimalData()` in various components.

## Additional Context
This change will reduce the load on the server and improve the user experience by speeding up the loading times.

## Fixes Jira Issue

### DIFF
--- a/apps/frontend/src/hooks/instrument/useInstrumentsMinimalData.ts
+++ b/apps/frontend/src/hooks/instrument/useInstrumentsMinimalData.ts
@@ -1,0 +1,78 @@
+import {
+  useEffect,
+  useState,
+  SetStateAction,
+  Dispatch,
+  useContext,
+} from 'react';
+
+import { UserContext } from 'context/UserContextProvider';
+import { InstrumentMinimalFragment, UserRole } from 'generated/sdk';
+import { useDataApi } from 'hooks/common/useDataApi';
+
+export function useInstrumentsMinimalData(callIds?: number[]): {
+  loadingInstruments: boolean;
+  instruments: InstrumentMinimalFragment[];
+  setInstruments: Dispatch<SetStateAction<InstrumentMinimalFragment[]>>;
+} {
+  const [instruments, setInstruments] = useState<InstrumentMinimalFragment[]>(
+    []
+  );
+  const [loadingInstruments, setLoadingInstruments] = useState(true);
+  const { currentRole } = useContext(UserContext);
+
+  const api = useDataApi();
+
+  useEffect(() => {
+    let unmounted = false;
+
+    setLoadingInstruments(true);
+    if (
+      currentRole &&
+      [
+        UserRole.USER_OFFICER,
+        UserRole.FAP_REVIEWER,
+        UserRole.FAP_CHAIR,
+        UserRole.FAP_SECRETARY,
+        UserRole.USER,
+      ].includes(currentRole)
+    ) {
+      api()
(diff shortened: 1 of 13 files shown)


apps/frontend/src/hooks/instrument/useInstrumentsMinimalData.ts
```diff
@@ -0,0 +1,78 @@
+import {
```
Scott-James-Hurley, line 1: This looks very similar to `useInstrumentsData.ts`. Should duplicated code be reduced or would it be preferable to keep them entirely separate?

TCMeldrum: I tried to, but I could not for the life of me get the typing to work in some way you would have something along the lines of `InstrumentDataMinimal | InstrumentData` which I could not get to type check in the places that needed `InstrumentData`. Happy to be proven wrong, but I could not make it work when I tryed.

TCMeldrum:
> This is grand. If optimisation is going to be an ongoing thing across teams, should best practice and design be discussed and agreed upon in order to make sure this effort is sustainable and doesn't go to waste down the road?

Maybe I can bring it up in the performance project meeting

apps/frontend/src/hooks/instrument/useInstrumentsMinimalData.ts
Scott-James-Hurley: image
I'll take that pay raise now, please.

Scott-James-Hurley (approved):

janosbabik (approved):
Hi @TCMeldrum, It looks good to me. However, do you have any evaluation data from before and after the changes? Having some statistics would be helpful when making performance-related changes.

TCMeldrum:
These numbers are against my local so take with a grain of salt.
We have about 40 instruments each with about 5-10 instrument scis attached, so is a comparison between these 2 calls:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/ofg87678/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/ofg87678/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#0563C1" vlink="#954F72">

Minimal | Full
-- | --
0.039487s | 1.592049s
0.010098s | 1.487671s
0.013802s | 1.058054s
0.008792s | 1.163378s
0.008592s | 1.495123s
0.006798s | 1.543326s



</body>

</html>


Keep in mind this query was uses across multiple pages including the landing pages for all roles bar `user`.

PR #846, opened 2024-11-06
fix: fap reviews not being deleted along with assignments

TCMeldrum:
Closes UserOfficeProject/issue-tracker#1232

## Description
This PR addresses the issue of removing FAP assignments, specifically handling the process in a transaction to ensure data consistency.

## Motivation and Context
This change is required to properly handle the removal of FAP members from a proposal. Previously, the removal of a member and their associated review were handled separately, which could lead to inconsistent states if one operation failed.

## Changes
1. The `removeMemberFromFapProposal` method in `FapDataSource.ts` and `PostgresFapDataSource.ts` now accepts an additional `reviewId` parameter, and the removal of the member and the associated review are handled in a transaction.
2. The `FapMutations.ts` file is updated to pass the `reviewId` to the `removeMemberFromFapProposal` method.
3. The `AssignMembersToFapMutation.ts` file is updated to include a `reviewId` field in the `RemoveFapReviewerFromProposalArgs` class.
4. The frontend component `FapProposalsAndAssignmentsTable.tsx` is updated to call the `removeMemberFromFapProposal` API with the `reviewId`.
5. The GraphQL mutation `removeMemberFromFapProposal` in `removeFapMemberFromProposal.graphql` is updated to include the `reviewId` field.

## Fixes Jira Issue

### DIFF
--- a/apps/backend/src/datasources/postgres/FapDataSource.ts
+++ b/apps/backend/src/datasources/postgres/FapDataSource.ts
@@ -858,28 +858,46 @@ export default class PostgresFapDataSource implements FapDataSource {
     fapId: number
   ) {
     await database.transaction(async (trx) => {
-      await trx<FapAssignmentRecord>('fap_assignments')
-        .insert(
-          assignments.map((assignment) => ({
-            proposal_pk: assignment.proposalPk,
-            fap_member_user_id: assignment.memberId,
-            fap_id: fapId,
-            fap_proposal_id: assignment.fapProposalId,
-          }))
+      try {
+        const fapAssignment = await database<FapAssignmentRecord>(
+          'fap_assignments'
         )
-        .returning<FapAssignmentRecord[]>(['*']);
-      await trx<ReviewRecord>('fap_reviews')
-        .insert(
-          assignments.map((assignment) => ({
-            user_id: assignment.memberId,
-            proposal_pk: assignment.proposalPk,
-            status: ReviewStatus.DRAFT,
-            fap_id: fapId,
-            fap_proposal_id: assignment.fapProposalId,
-            questionary_id: assignment.questionaryId,
-          }))
-        )
-        .returning<ReviewRecord[]>(['*']);
+          .insert(
+            assignments.map((assignment) => ({
+              proposal_pk: assignment.proposalPk,
+              fap_member_user_id: assignment.memberId,
+              fap_id: fapId,
+              fap_proposal_id: assignment.fapProposalId,
+            }))
+          )
+          .transacting(trx);
+        const reviewRecord = await database<ReviewRecord>('fap_reviews')
+          .insert(
+            assignments.map((assignment) => ({
+              user_id: assignment.memberId,
+              proposal_pk: assignment.proposalPk,
+              status: ReviewStatus.DRAFT,
+              fap_id: fapId,
+              fap_proposal_id: assignment.fapProposalId,
+              questionary_id: assignment.questionaryId,
+            }))
+          )
+          .transacting(trx);
+
+        if (!(fapAssignment && reviewRecord)) {
+          throw new GraphQLError('Could not delete reviewer');
+        }
+
+        return await trx.commit();
+      } catch (error) {
+        logger.logException(
+          `Could not delete reviewer ags: '${JSON.stringify(assignments)}'`,
+          error
+        );
+        trx.rollback();
+
+        throw new GraphQLError('Could not delete reviewer');
+      }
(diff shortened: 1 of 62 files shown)


Scott-James-Hurley (commented):
Is there a circumstance where you want to keep a review not associated with an assignment? For example, someone wrote a review and then they were unassigned from the proposal? If not, could we not delete `fap_review` entries based on `fap_id`, `proposal_pk` and `member_id` instead of passing the `review_id`?

TCMeldrum:
> Is there a circumstance where you want to keep a review not associated with an assignment? For example, someone wrote a review and then they were unassigned from the proposal? If not, could we not delete `fap_review` entries based on `fap_id`, `proposal_pk` and `member_id` instead of passing the `review_id`?

I can't think of a situation, this is the only referance to ` removeMemberFromFapProposal` I saw.
Yeah we could maybe something to do in https://github.com/UserOfficeProject/issue-tracker/issues/1231

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -894,17 +894,27 @@ export default class PostgresFapDataSource implements FapDataSource {
...
-      .del()
-      .where('fap_id', fapId)
-      .andWhere('proposal_pk', proposalPk)
-      .andWhere('fap_member_user_id', memberId);
+    await database.transaction(async (trx) => {
+      await trx
+        .from('fap_reviews')
+        .where('review_id', reviewId)
+        .returning('*')
+        .del();
```
ACLay, line 905: Rather than add a reviewId parameter, could we use the existing filtering arguments? This table already has equivalent `fap_id`, `proposal_pk` and `user_id` columns to match the ones filtered on in the fap_assignments delete, and that should ensure that an API client can't call this mutation with a `reviewId` that doesn't line up with the other parameters.

ACLay (changes requested):

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
```diff
@@ -355,10 +355,11 @@ const FapProposalsAndAssignmentsTable = ({
...
       }
 
-      allProposalReviews.push(...proposalReviews);
-      allProposalReviews.map(
-        (proposalReview) => (proposalReview.proposalPk = proposalPk)
+      const proposalReviewsWithProposalPk = proposalReviews.map(
+        (proposalReview) => ({ ...proposalReview, proposalPk })
       );
+
+      allProposalReviews.push(...proposalReviewsWithProposalPk);
```
ACLay, line 362: I think we could simplify away the intermediate array
```suggestion
      proposalReviews.forEach(
        (proposalReview) => allProposalReviews.push({ ...proposalReview, proposalPk })
      );
```

ACLay (dismissed):

apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
TCMeldrum: testing locally looks good

William-Edwards-STFC (approved):

RasmiaKulan (approved):

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -896,15 +896,26 @@ export default class PostgresFapDataSource implements FapDataSource {
...
+
+      await trx('fap_assignments')
+        .del()
+        .where('fap_id', fapId)
+        .andWhere('proposal_pk', proposalPk)
+        .andWhere('fap_member_user_id', memberId)
+        .returning<number>(['*']);
+    });
```
gnyiri, line 915: Do we need error handling and check the result of the transaction? The new implementation will not check if the transaction has failed whereas the old one did (and I have seen error handling for transactions in CallDataSource::update for instance).

gnyiri: My comment was raised for the delete not the insert. Can you add the error check for removeMemberFromFapProposal as well?

TCMeldrum: Sorry will add there too 🤦

TCMeldrum: @gnyiri please can you look again please!

gnyiri (approved):

PR #847, opened 2024-11-06
fix: proposal reviewer table fixes

TCMeldrum:
## Description
This PR optimizes the proposal reviewer table by refining the status check and improving the selection functionality.

## Motivation and Context
The changes in this PR are required to handle the edge case where the status could be undefined and to manage the selection state more effectively, thus enhancing the user experience and functionality of the proposal reviewer table.

## Changes
- In the `ReviewDataSource.ts` file, the status check has been improved to handle both `null` and `undefined` cases.
- In the `ProposalTableReviewer.tsx` file, the `selection` state is now derived from `searchParams` rather than being a constant. This allows it to update dynamically based on changes to the `searchParams`.
- The dependency array in the `useEffect` hook has been updated to include `searchParams` to ensure that changes to these parameters trigger a re-render.

## Fixes Jira Issue

### DIFF
--- a/apps/backend/src/datasources/postgres/ReviewDataSource.ts
+++ b/apps/backend/src/datasources/postgres/ReviewDataSource.ts
@@ -372,7 +372,7 @@ export default class PostgresReviewDataSource implements ReviewDataSource {
               qb.where('instrument_has_proposals.instrument_id', instrumentId);
             }
 
-            if (status !== null) {
+            if (status !== null && status !== undefined) {
               qb.where('fap_reviews.status', status);
             }
           })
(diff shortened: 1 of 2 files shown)


apps/backend/src/datasources/postgres/ReviewDataSource.ts
```diff
@@ -372,7 +372,7 @@ export default class PostgresReviewDataSource implements ReviewDataSource {
               qb.where('instrument_has_proposals.instrument_id', instrumentId);
             }
 
-            if (status !== null) {
+            if (status !== null && status !== undefined) {
```
janosbabik, line 375:
```suggestion
            if (status) {
```

janosbabik (changes requested):
Hi @TCMeldrum, just a small suggestion from me.

apps/backend/src/datasources/postgres/ReviewDataSource.ts
TCMeldrum: Review status is of type:
`export enum ReviewStatus {
  DRAFT = 0,
  SUBMITTED = 1,
}` So could be 0 which would not pass this check when we need it too.

janosbabik (approved):

William-Edwards-STFC (approved):

PR #857, opened 2024-11-18
feat: remove the merge user tool’s stored procedure #1155

bashanlam:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1155

## Content
Remove the merge user tool’s stored procedure.

## Motivation and Context
Remove the merge user tool’s stored procedure.

## How Has This Been Tested
Manually check relocation. 

## Changes
Remove the store procedure.


bashanlam:
PR to relocate the function the merge user tool's stored procedure script of the new submission system to the designated STFC location is https://github.com/isisbusapps/internal-scripts/pull/147

mutambaraf (commented):
Maybe lets use a patch that just drop the function from the database since it already exists.

bashanlam:
> Maybe lets use a patch that just drop the function from the database since it already exists.

I prefer dropping the function manually since we will no longer use a patch to create it. Additionally, we do not want it to remain in the User-office-core repository, as suggested by ESS.

mutambaraf:
> > Maybe lets use a patch that just drop the function from the database since it already exists.
> 
> I prefer dropping the function manually since we will no longer use a patch to create it. Additionally, we do not want it to remain in the User-office-core repository, as suggested by ESS.

This is something to discuss with ESS and ELI if they are fine with manually remove the function from their databases then.

bashanlam:
> > > Maybe lets use a patch that just drop the function from the database since it already exists.
> > 
> > 
> > I prefer dropping the function manually since we will no longer use a patch to create it. Additionally, we do not want it to remain in the User-office-core repository, as suggested by ESS.
> 
> This is something to discuss with ESS and ELI if they are fine with manually remove the function from their databases then.

A patch will be used to drop the function from the database.

bashanlam:
A new pull to drop store procedure will be created instead.

PR #861, opened 2024-11-19
perf: optimize getCallByQuestionId

TCMeldrum:


apps/backend/src/resolvers/queries/CallQuery.ts
```diff
@@ -14,10 +14,10 @@ export class CallQuery {
   }
 
   @Query(() => Call, { nullable: true })
-  callByQuestionId(
-    @Arg('questionId', () => String) questionId: string,
+  getCallAllocatedTimeUnit(
```
ACLay, line 17: While we're only using this in the frontend for the time unit, as a backend query it could be used for anything on the Call object, so would something like `getCallOfAnswersProposal` be better here too?

ACLay (commented):
Looks like a good optimisation, I just wonder if the name of the new query is a bit obtuse, it definitely took me a minute to figure out the logic of the links. Would something like `getCallOfAnswersProposal` be clearer than `getCallByAnswerId`.

simonfernandes (approved):

janosbabik (approved):

apps/backend/src/resolvers/queries/CallQuery.ts
yoganandaness: I agree with @ACLay. This must be frontend agnostic.

apps/backend/src/datasources/CallDataSource.ts
```diff
@@ -28,5 +28,5 @@ export interface CallDataSource {
   getCallsByInstrumentScientist(scientistId: number): Promise<Call[]>;
   isCallEnded(callId: number, checkIfInternalEnded: boolean): Promise<boolean>;
   isCallEnded(callId: number): Promise<boolean>;
-  getCallByQuestionId(questionId: string): Promise<Call>;
+  getCallOfAnswersProposal(answerId: number): Promise<Call>;
```
yoganandaness, line 31:
```suggestion
  getCallByAnswerIdProposal(answerId: number): Promise<Call>;
```

apps/backend/src/resolvers/queries/CallQuery.ts
```diff
@@ -14,10 +14,13 @@ export class CallQuery {
...
```
yoganandaness, line 17:
```suggestion
  getCallByAnswerId(
```

yoganandaness (changes requested):

yoganandaness (approved):
Seems fine to me

PR #876, opened 2024-11-29
chore: swap 4228 uo fe use search params type implementation

yoganandaness:
## Description

Add strong typings to the query params implementation. This is the improvement to the PR https://github.com/UserOfficeProject/user-office-core/pull/748

## Motivation and Context

In the current implementation of useSearchParams from react-router, it is not typed. To improve the quality of the code, a hook should be created that adds the layer of typings to the various query parameters. 

## How Has This Been Tested

Manually. e2e tests are already covered. 

### DIFF
--- a/apps/frontend/src/components/common/SimpleTabs.tsx
+++ b/apps/frontend/src/components/common/SimpleTabs.tsx
@@ -3,8 +3,9 @@ import { useTheme } from '@mui/material/styles';
 import Tab from '@mui/material/Tab';
 import Tabs from '@mui/material/Tabs';
 import useMediaQuery from '@mui/material/useMediaQuery';
-import React from 'react';
-import { useSearchParams } from 'react-router-dom';
+import React, { useMemo } from 'react';
+
+import { useTypeSafeSearchParams } from 'hooks/common/useTypeSafeSearchParams';
 
 interface TabPanelProps {
   children?: React.ReactNode;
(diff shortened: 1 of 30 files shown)


apps/frontend/src/components/common/SimpleTabs.tsx
```diff
@@ -58,11 +59,25 @@ const SimpleTabs = ({
...
+      tab: null,
+    }),
+    []
+  );
+
+  const [typedParams, setTypedParams] = useTypeSafeSearchParams<{
+    verticalTab: number | null;
+    modalTab: number | null;
+    tab: number | null;
+  }>(initialParams);
```
ACLay, line 76: Here it's stating that the values are numbers, but how can that be the case? Internally to `useTypeSafeSearchParams` the values are pulled from the URL, so wouldn't all values be strings? And the only type conversion I can see within that hook is to turn things into strings.

ACLay (commented):
I'm not quite sure what benefit this is providing. When I look at the code changes in this it just seems longer with more wrappers and boilerplate. It also replaces names like `searchParams` with `typedParams` which feels less descriptive, as rather than about the URL's search parameters, it could be any sort of parameters.

apps/frontend/src/components/common/SimpleTabs.tsx
yoganandaness: That's the idea. To make the query params consistent with types

`useTypeSafeSearchParams`
This takes care of typecasting any types to string.

Using the above hook, any component can register typed query params, which makes the application more consistent. For ex., I cannot pass proposal_id as `abcd`, since it is an integer and my implementation will throw type error. Just like how the default React useState behaves.

apps/frontend/src/hooks/common/useTypeSafeSearchParams.ts
```diff
@@ -0,0 +1,60 @@
...
+          return searchParams;
+        });
+
+        return newParams;
+      });
+    },
+    [setSearchParams]
+  );
+
+  return [typedParams, setTypedSearchParams];
```
jekabs-karklins, line 59: Nitpick: Maybe the return values should not be bound to the implementation, but to the concept. So, instead of:
```typescript
return [typedParams, setTypedSearchParams];
```
simply
```typescript
return [params, setSearchParams];
```

apps/frontend/src/hooks/common/useTypeSafeSearchParams.ts
```diff
@@ -0,0 +1,60 @@
+import { useState, useEffect, useCallback } from 'react';
```
jekabs-karklins, line 1: question: I do not understand if this hook actually does the type conversion or just marks the type. 
I think searchParams always returns strings.
so would not this result in the following?
```ts
convertToTypedParams({userId: "1"}) => {userId: "1"}:{userId: number}
```
So the type would be number while it is actually a string?

jekabs-karklins (commented):
I see two sides for this PR.

Upside:
It is nice that the params gets the types, so it is less error prone, and needs less type assertion

Downside:
It is a little more boilerplate and somewhat takes more mental power to read for the person who is used to the standard useParams and useSearchParams syntax form ReactRouter, but I guess it is just a matter of getting used to it.

PR #881, opened 2024-12-02
fix: show whole review on preview

TCMeldrum:
In the meetings they would like to be able to see the whole review when looking at the FAP reviews


RasmiaKulan (approved):

Scott-James-Hurley:
Is this so urgent it needs to be hotfixed?

janosbabik (approved):

PR #883, opened 2024-12-03
chore: mergeback after hotfix

TCMeldrum:
## Description
This PR removes the use of the `truncateString` function in the 'Comment' field of the Review Questionary Review component.

It also adds a `stripHtml` call to remove markup from access panel excel exports.

## Motivation and Context
The `truncateString` function was unnecessarily limiting the length of the 'Comment' field which could potentially cut off important information. The removal of this function allows for full visibility of the comments.

The html markup as plaintext in the excel exports made the columns unusable.

## Changes
- The `truncateString` function import statement was removed from 'ReviewQuestionaryReview.tsx'.
- The function call to `truncateString` was removed from the 'Comment' field. Now, the 'Comment' field value only uses the `stripHtml` function to clean the input, but doesn't limit its length.
- The access panel excel export adds a call to `stripHtml` to make the output readable

## Fixes Jira Issue


jekabs-karklins (approved):
LGTM

Scott-James-Hurley (approved):

PR #884, opened 2024-12-03
perf: optimise answer topic mutation

Scott-James-Hurley:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1243

## Description

AnswerTopic mutation now only returns changed answers. The following times are for a topic with one answer, the most flattering comparison, averaged over 5 times.

Old answerTopic time: 254ms
New answerTopic time: 88ms

Arguments:
image

## Motivation and Context

This reduces the data passed to the frontend and therefore load, increasing performance.

## How Has This Been Tested

Unit, integration and E2E tests have been run to test for regression.

## Changes

- Logging has been updated for the new return value
- AnswerTopic mutation return changed answers, instead of the whole topic
- Frontend logic has been updated to handle new answertopic return value


apps/frontend/src/models/questionary/QuestionarySubmissionState.ts
```diff
@@ -243,12 +255,26 @@ export function QuestionarySubmissionModel<
...
           const stepIndex = draftState.questionary.steps.findIndex(
-            (step) => step.topic.id === updatedStep.topic.id
+            (step) => step.topic.id === action.topicId
           );
-          draftState.questionary.steps[stepIndex] = updatedStep;
 
+          draftState.questionary.steps[stepIndex].fields =
+            draftState.questionary.steps[stepIndex].fields.map((f) => {
+              const updatedAnswer = action.answers.find(
+                (u) => u.questionId === f.question.id
```
TCMeldrum, line 265: Could you use more meaning variable names here?

TCMeldrum (changes requested):
Looks good only one small nit pick sorry. 

Could you add some data to the spreadsheet I have linked you in teams around the answer topic mutation.

EdwardHaynes (approved):

TCMeldrum (approved):

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -147,6 +149,15 @@ export default class QuestionaryMutations {
           answer.questionId,
           answer.value
         );
+
+        const updatedAnswer = await this.dataSource.getAnswer(
```
yoganandaness, line 153: What if we do bulk `select` instead of single query? - This way. only one query calls, instead of n

Infact, we can even get the answer details from the `updateAnswer` by using `returning` in the postgres query. This way, we don't need to make a query to fetch at all.

yoganandaness (changes requested):

apps/backend/src/mutations/QuestionaryMutations.ts
Scott-James-Hurley: Thanks for the feedback. To clarify: do you mean modifying `updateAnswer` so it returns the data I need, not just the question id?

yoganandaness (approved):
Good work!!!

PR #886, opened 2024-12-03
fix: update what comment the fap sheet export uses

TCMeldrum:


apps/backend/db_patches/0166_UpdateWhichCommentOnReviewProposalView.sql
```diff
@@ -0,0 +1,57 @@
+DO
+$$
+BEGIN
+    IF register_patch('0166_ExcludeExpiredProposalsFromReviewProposalView.sql', 'Thomas Cottee Meldrum', 'Update review column in the view', '2024-12-03') THEN
```
ACLay, line 4: Can the filename passed to register_patch be updated to match?

TCMeldrum: Missed that thanks!

jekabs-karklins (approved):

PR #892, opened 2024-12-12
perf: cache raw stfc rolls call

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1290

## Description
This PR improves the performance of our application by caching raw STFC roles call.

## Motivation and Context
This change is required to optimize the application and reduce unnecessary calls to the STFC service. By caching the raw STFC roles call, we are able to significantly reduce the load time and enhance the user experience.

## Changes
1. The method `getRolesForUser` in the `StfcUserAuthorization` class has been updated to fetch roles from `StfcUserDataSource` instead of directly from the client.
2. A new cache `uowsRawRolesCache` has been introduced in the `StfcUserDataSource` class to store raw STFC roles.
3. The method `getRolesForUser` in the `StfcUserDataSource` class has been updated to use the newly introduced cache. If the roles are not in the cache, it fetches them from the client, stores them in the cache, and then returns them.
4. The method `getUserRoles` in the `StfcUserDataSource` class has been updated to use the `getRolesForUser` method instead of directly calling the client. This allows it to benefit from the caching mechanism.

## Fixes Jira Issue


apps/backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -203,9 +203,9 @@ export class StfcUserAuthorization extends UserAuthorization {
 
     // Auto-assign users to instruments.
     // This will happen if the user is an instrument scientist for an STFC facility
-    const stfcRoles: stfcRole[] | null = (
-      await client.getRolesForUser(process.env.EXTERNAL_AUTH_TOKEN, userNumber)
-    )?.return;
+    const stfcRoles: stfcRole[] | null = await (
+      this.userDataSource as StfcUserDataSource
```
ACLay, line 207: Is the cast to StfcUserDataSource necessary? It looks like the variable is already typed that way.

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -128,6 +128,11 @@ export class StfcUserDataSource implements UserDataSource {
     StfcUserDataSource.rolesCacheSecondsToLive
   ).enableStatsLogging('uowsRolesCache');
 
+  private uowsRawRolesCache = new Cache<Promise<stfcRole[] | null>>(
```
ACLay, line 131: I see where you're coming from with the `raw` names, but I wonder if it might be more intuitive if this new cache was called `uowsRolesCache` or `stfcRolesCache` and the existing one was renamed to `uopRolesCache`?

jekabs-karklins (approved):
This looks ok, but one thing I want to ask if maybe this is making more sense to be cached at UOWSSoapClient level rather than datasource. 

Of course it is up to you, I'm just thinking that putting it in there seems more logical as if there is another component relying on UOWSSoapClient it would  benefit from the cache.

TCMeldrum:
The `UOWSSoapClient` is auto generated, so don't think it would be feasible to put it in there. I think this is fine for us for now as we are migrating away from the soap client, maybe we could look at putting it in the new client.

PR #897, opened 2024-12-17
feat: migrate stfc users service to rest

William-Edwards-STFC:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1060

## Description
This PR introduces REST changes to the application, replacing SOAP with REST in STFC UserOfficeWebService, and updating relevant configurations and dependencies.

## Motivation and Context
The change is required to migrate from the SOAP-based STFC UserOfficeWebService to a REST-based service. This will provide a more modern, efficient, and easy-to-maintain approach to data communication, ultimately improving the performance and reliability of the application.

Please find the mock server pr below.

https://github.com/UserOfficeProject/mock-uows/pull/18

Please find the docker orchestration pr below.

https://github.com/isisbusapps/docker-orchestration/pull/687
## Changes
1. Updated .prettierignore, .vscode/launch.json, README.md, apps/backend/.eslintignore, apps/backend/.gitignore, and apps/backend/docker-compose-stfc.e2e.yml files to reflect the migration from SOAP to REST.
2. Introduced the openapi-typescript-codegen package to generate TypeScript/JavaScript client APIs from OpenAPI Specification (OAS).
3. Removed the SOAP client for STFC UserOfficeWebService and replaced it with an auto-generated REST client.

## Fixes Jira Issue


.github/workflows/test-build.yml
```diff
@@ -177,7 +177,7 @@ jobs:
             sdk-frontend-
 
       - run: npm ci
-
+
```
jekabs-karklins, line 180: This is probably accidental change

William-Edwards-STFC: we were just doing this for testing purposes and it has been reverted since then

.github/workflows/test-build.yml
```diff
@@ -343,7 +343,7 @@ jobs:
           EMAIL_FOOTER_IMAGE_PATH: /config/logos/STFC-Logo-small.png
           EXTERNAL_AUTH_LOGIN_URL: http://localhost:9003/auth/Login.aspx
           EXTERNAL_AUTH_LOGOUT_URL: http://localhost:9003/auth/Login.aspx
-          EXTERNAL_AUTH_SERVICE_URL: http://localhost:1080/ws/UserOfficeWebService?wsdl
+          EXTERNAL_UOWS_API_URL: http://localhost:1080/users-service
```
mutambaraf, line 346: For this do we also need to set the new values in docker-orchestration

apps/backend/Dockerfile
```diff
@@ -7,8 +7,10 @@ RUN mkdir -p /home/node/app
 WORKDIR /home/node/app
 
 COPY --chown=node:node package*.json ./
+COPY openapi.yaml openapi.yaml
```
mutambaraf, line 10: If possible can we move this file to docker-orchestration or busapp settings .If possible we can load it up as config maps.

TCMeldrum: I'm not sure we can but happy to be proved worng. Right now the main 2 issues I see are:
- The linter doesn't pass without the generated files meaning the linter fails for anyone who would not have access to docker-orchestration or busapp settings like github actions or other facilitys working on this project
- The test would also fail on github actions for the same reason as above

mutambaraf: That must be true the only other way is to change this into a package but all depends what ELI and ESS say l guess.

ACLay: If the problem is that the linter doesn't like the generated files, we just need to tell it to ignore them by adding the generated directory to the `.eslintignore` file.

TCMeldrum: It is a problem that were functions from the generated files are used. Like the `STFCUserDataSource` doesnt type check if the gerenerated files dont exist.

apps/backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -164,46 +172,67 @@ export class StfcUserAuthorization extends UserAuthorization {
...
-    const userNumber = parseInt(stfcUser.userNumber);
-    const dummyUser =
-      await this.userDataSource.ensureDummyUserExists(userNumber);
+    let userNumber;
+    let dummyUser;
+    if (stfcUser!.userNumber != null) {
+      userNumber = parseInt(stfcUser!.userNumber);
+      dummyUser = await this.userDataSource.ensureDummyUserExists(userNumber);
+    } else {
+      throw error;
```
mutambaraf, line 222: Can we also throw this error with a message .

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -293,9 +341,11 @@ export class StfcUserDataSource implements UserDataSource {
   }
 
   async getByEmail(email: string): Promise<User | null> {
-    return this.getStfcBasicPersonByEmail(email).then((stfcUser) =>
+    const dd = this.getStfcBasicPersonByEmail(email).then((stfcUser) =>
```
mutambaraf, line 344: We can also improve the variable name here.

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -306,20 +356,19 @@ export class StfcUserDataSource implements UserDataSource {
...
+  async getRolesForUser(id: number): Promise<RoleDTO[]> {
     const cachedRoles = this.stfcRolesCache.get(String(id));
     if (cachedRoles) {
       return cachedRoles;
     }
 
-    const stfcRawRolesRequest = client
-      .getRolesForUser(token, id)
-      .then((response): stfcRole[] | null => response?.return);
+    const stfcRawRolesRequest = UOWSClient.role.getRolesForUser(id.toString());
```
mutambaraf, line 365: Can we also catch some of the errors when calling the client and specify which endpoint of the user office client we are calling in the error message.

ACLay (dismissed):
I mentioned in another comment's replies adding the generated files to the `.eslintignore`. In addition to that, since they get generated at runtime can we also add them to the dockerignore so we don't end up copying in any outdated client files from the host machine that could cause conflicts.

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -164,28 +193,36 @@ export class StfcUserDataSource implements UserDataSource {
...
+            undefined,
+            undefined,
+            cacheMisses
+          )
+        : await UOWSClient.basicPersonDetails.getBasicPersonDetails(
+            undefined,
+            undefined,
+            undefined,
+            cacheMisses
+          );
```
mutambaraf, line 207: Here also we can catch any errors from the client.

apps/backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -164,46 +172,67 @@ export class StfcUserAuthorization extends UserAuthorization {
...
+    }
+    const stfcUserTemp: BasicPersonDetailsDTO[] | null =
+      await UOWSClient.basicPersonDetails
+        .getBasicPersonDetails(undefined, undefined, undefined, [
+          loginBySession.userId,
+        ])
+        .then((rawStfcUser) => rawStfcUser)
+        .catch((error) => {
+          const rethrowMessage =
+            'Failed to fetch user details for STFC external authentication';
```
mutambaraf, line 192: We can also add the end point we are calling in the error log

.github/workflows/test-build.yml
William-Edwards-STFC: We do need to change it in two places I have made a PR for this here https://github.com/isisbusapps/docker-orchestration/pull/687

apps/backend/src/datasources/stfc/UOWSClient.ts
```diff
@@ -0,0 +1,14 @@
...
+import { UOWS } from '../../generated';
+
+dotenv.config({ path: process.env.CONFIG_PATH });
+
+export function createUOWSClient() {
+  return new UOWS({
+    BASE: process.env.EXTERNAL_UOWS_API_URL,
+    HEADERS: {
+      Authorization: `Api-key ${process.env.API_KEY}`,
+    },
```
mutambaraf, line 12: I know the old key will still be working but can you have a chat with users about updating to new keys.

apps/e2e/cypress/fixtures/exampleCallFapExport.json
jekabs-karklins, line 1: questions: I see that for many files this PR changes the permissions from 755->644
Is this accidental? If not I see that it is a good idea to mark them as not executable but maybe that can be part of a separate PR and have it done repository wide.

apps/backend/src/datasources/stfc/UOWSClient.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { UOWS } from '../../generated';
+
+dotenv.config({ path: process.env.CONFIG_PATH });
+
+export function createUOWSClient() {
+  return new UOWS({
+    BASE: process.env.EXTERNAL_UOWS_API_URL,
+    HEADERS: {
+      Authorization: `Api-key ${process.env.API_KEY}`,
```
ACLay, line 11: `API_KEY` looks like it's a new setting, but it looks like there's still a reference to the old `EXTERNAL_AUTH_TOKEN` one in StfcUserDataSource, and I can't find a PR in our config repo to add the new value.

mutambaraf, line 11: Lets update here also EXTERNAL_UOWS_API_KEY

apps/backend/src/datasources/stfc/UOWSClient.ts
```diff
@@ -0,0 +1,14 @@
+import dotenv from 'dotenv';
+
+import { UOWS } from '../../generated';
+
+dotenv.config({ path: process.env.CONFIG_PATH });
```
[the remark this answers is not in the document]
William-Edwards-STFC: Yeah I moved it here its not done twice

apps/backend/.dockerignore
```diff
@@ -5,4 +5,5 @@
 !db_patches
 !package*
 !tsconfig.json
-!index.ts
\ No newline at end of file
+!index.ts
+src/generated
```
mutambaraf, line 9: Worthy checking if this is working because be are including the whole src directory.

mutambaraf (commented):
Looks good to me but have we resolved adding `openapi.yaml` to this repo

ACLay:
I don't see how we could avoid including an api spec or client. We need a client for the code to compile, so we either include that as client code, or the service definition to generate one at build time. I guess it might be possible to have the client as a separate npm package, but the extra infrastructure for that seems excessive compared to just adding an npm command and yaml file here. And it's not like having a client/spec here is new, the old `UOWSSoapInterface.ts` is generated source code for the SOAP client.

TCMeldrum:
I agree alex I don't see a way around adding the open-api file or the compiled code to this repo without massive hassle to both us and other collaborators. I think I prefer adding the `open-api.yaml` rather than a folder of regenerated code.

mutambaraf:
We have more like a month to come up with a package which l think might be a good solution here.

ACLay:
Sure we've got a month, but what is the problem that having a separate package would address? It'd be adding extra steps to our update process if we need a new client to pull in UOWS updates, and giving us more infrastructure to maintain, so there'd have to be some benefit for that to be worth it, but I don't see what that would be.

.github/workflows/test-build.yml
```diff
@@ -343,7 +376,7 @@ jobs:
...
```
mutambaraf, line 379: Also l may have missed but this port might be different for `http://localhost:1080/users-service`

mutambaraf (approved):

.github/workflows/test-build.yml
```diff
@@ -108,6 +108,18 @@ jobs:
...
+      - name: Cache UOWS
+        id: cached-UOWS
+        uses: actions/cache@v4
+        with:
+          path: apps/backend/generated
+          key: uows-client-${{ hashFiles('apps/backend/openapi.yaml') }}-${{ env.NODE_VERSION }}
+
+      - name: Generate backend UOWS
+        run: |
+          cd apps/backend
```
jekabs-karklins, line 120: Suggestion: Do we want to run this every time or only when there is no cache hit?

if: steps.cached-UOWS.outputs.cache-hit != 'true'

.vscode/launch.json
```diff
@@ -1,24 +1,25 @@
...
-            "name": "Attach to Windows Docker container",
-            "type": "node",
-            "request": "attach",
-            "port": 9229,
-            "protocol": "inspector",
-            "localRoot": "${workspaceFolder}",
-            "remoteRoot": "C:\\app"
-        }
-    ]
+  // Use IntelliSense to learn about possible attributes.
```
jekabs-karklins, line 2: Question: Is this change intentional?

jekabs-karklins (approved):
Approved! But please don't forget to review the two comments I've added, and adjust as you see fit.

.vscode/launch.json
William-Edwards-STFC: I have reverted this back to normal

.github/workflows/test-build.yml
William-Edwards-STFC: Thats a good idea I'll add that thank you

PR #908, opened 2025-01-14
feat: add co proposer claim to invite

jekabs-karklins:
## Description
This pull request adds co-proposer claim to the invite
Although 20+  files have been touched, the main changes are in the following files
- apps/backend/db_patches/0168_AddCoProposerClaims.sql
- apps/backend/src/auth/InviteAuthorizer.ts
- apps/backend/src/datasources/postgres/CoProposerInviteDataSource.ts
- apps/backend/src/mutations/InviteMutations.ts

## Motivation and Context
Now it is possible to not only add a role, but also a co-proposer claim. This allows to create invites which when claimed assigns user a role.

## Changes
1. Added a new table `co_proposer_invites` in the PostgreSQL database to store co-proposer invite data.
2. Created the `InviteAuthorizer.ts` to include a new `InviteAuthorization` class to handle co-proposer invite authorization.
3. Updated package.json to use latest duo-validation for argument validation.
4. Updated the dependency configuration files to include the new `CoProposerInviteDataSource` and `InviteAuthorization` classes.

## How Has This Been Tested?

Added Unit tests

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4395](https://jira.esss.lu.se/browse/SWAP-4395)

### DIFF
--- a/apps/backend/db_patches/0168_AddCoProposerClaims.sql
+++ b/apps/backend/db_patches/0168_AddCoProposerClaims.sql
@@ -4,11 +4,11 @@ BEGIN
   IF register_patch('0168_CreateCoProposerInvite', 'Jekabs Karklins', 'Adding co-proposer claims', '2025-01-10') THEN
     BEGIN
       CREATE TABLE co_proposer_invites (
-        role_invite_id SERIAL PRIMARY KEY,
+        co_proposer_invite_id SERIAL PRIMARY KEY,
         invite_code_id INT NOT NULL REFERENCES invite_codes(invite_code_id) ON DELETE CASCADE,
-        role_id INT NOT NULL REFERENCES roles(role_id) ON DELETE CASCADE
+        proposal_pk INT NOT NULL REFERENCES proposals(proposal_pk) ON DELETE CASCADE
       );
-      CREATE INDEX co_proposer_invites_invite_code_id_idx ON co_proposer_invites(invite_code_id);
+      CREATE INDEX co_proposer_co_proposer_invite_id_idx ON co_proposer_invites(co_proposer_invite_id);
 
     END;
   END IF;
--- a/apps/backend/db_patches/0168_AddCoProposerClaims.sql
+++ b/apps/backend/db_patches/0168_AddCoProposerClaims.sql
@@ -1,0 +1,17 @@
+DO
+$$
+BEGIN
+  IF register_patch('0168_CreateCoProposerInvite', 'Jekabs Karklins', 'Adding co-proposer claims', '2025-01-10') THEN
+    BEGIN
+      CREATE TABLE co_proposer_invites (
+        role_invite_id SERIAL PRIMARY KEY,
+        invite_code_id INT NOT NULL REFERENCES invite_codes(invite_code_id) ON DELETE CASCADE,
+        role_id INT NOT NULL REFERENCES roles(role_id) ON DELETE CASCADE
+      );
+      CREATE INDEX co_proposer_invites_invite_code_id_idx ON co_proposer_invites(invite_code_id);
+
+    END;
+  END IF;
+END;
+$$
+LANGUAGE plpgsql;
--- a/apps/backend/db_patches/0168_AddCoProposerClaims.sql
+++ b/apps/backend/db_patches/0168_AddCoProposerClaims.sql
@@ -3,12 +3,12 @@ $$
 BEGIN
   IF register_patch('0168_CreateCoProposerInvite', 'Jekabs Karklins', 'Adding co-proposer claims', '2025-01-10') THEN
     BEGIN
-      CREATE TABLE co_proposer_invites (
+      CREATE TABLE IF NOT EXISTS co_proposer_invites (
         co_proposer_invite_id SERIAL PRIMARY KEY,
         invite_code_id INT NOT NULL REFERENCES invite_codes(invite_code_id) ON DELETE CASCADE,
         proposal_pk INT NOT NULL REFERENCES proposals(proposal_pk) ON DELETE CASCADE
       );
-      CREATE INDEX co_proposer_co_proposer_invite_id_idx ON co_proposer_invites(co_proposer_invite_id);
+      CREATE INDEX IF NOT EXISTS co_proposer_co_proposer_invite_id_idx ON co_proposer_invites(co_proposer_invite_id);
 
     END;
   END IF;
--- a/apps/backend/db_patches/0168_AddCoProposerClaims.sql
+++ b/apps/backend/db_patches/0168_AddCoProposerClaims.sql
@@ -4,12 +4,9 @@ BEGIN
   IF register_patch('0168_CreateCoProposerInvite', 'Jekabs Karklins', 'Adding co-proposer claims', '2025-01-10') THEN
     BEGIN
       CREATE TABLE IF NOT EXISTS co_proposer_invites (
-        co_proposer_invite_id SERIAL PRIMARY KEY,
         invite_code_id INT NOT NULL REFERENCES invite_codes(invite_code_id) ON DELETE CASCADE,
-        proposal_pk INT NOT NULL REFERENCES proposals(proposal_pk) ON DELETE CASCADE
-      );
-      CREATE INDEX IF NOT EXISTS co_proposer_co_proposer_invite_id_idx ON co_proposer_invites(co_proposer_invite_id);
-
+        proposal_pk INT NOT NULL REFERENCES proposals(proposal_pk) ON DELETE CASCADE,
+        PRIMARY KEY (invite_code_id, proposal_pk));
     END;
   END IF;
 END;
--- a/apps/backend/src/auth/InviteAuthorizer.ts
+++ b/apps/backend/src/auth/InviteAuthorizer.ts
@@ -14,12 +14,13 @@ export class InviteAuthorization {
     agent: UserWithRole | null,
     roleIds?: number[]
   ) => {
-    // If no roleIds are provided, the invite is authorized
+    // If no roleIds are provided, the invite is considered as authorized
     if (roleIds === undefined || roleIds.length === 0) return true;
 
-    const onlyUserRole = roleIds.length === 1 && roleIds[0] === UserRole.USER;
-    const isUserOfficer = this.userAuth.isUserOfficer(agent);
+    if (this.userAuth.isUserOfficer(agent)) return true;
 
-    return isUserOfficer || onlyUserRole;
+    const onlyUserRole = roleIds.length === 1 && roleIds[0] === UserRole.USER;
+
+    return onlyUserRole;
   };
 }
--- a/apps/backend/src/auth/InviteAuthorizer.ts
+++ b/apps/backend/src/auth/InviteAuthorizer.ts
@@ -14,6 +14,7 @@ export class InviteAuthorization {
     agent: UserWithRole | null,
     roleIds?: number[]
   ) => {
+    // If no roleIds are provided, the invite is authorized
     if (roleIds === undefined || roleIds.length === 0) return true;
 
     const onlyUserRole = roleIds.length === 1 && roleIds[0] === UserRole.USER;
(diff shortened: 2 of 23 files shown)


apps/backend/db_patches/0168_AddCoProposerClaims.sql
```diff
@@ -0,0 +1,17 @@
+DO
+$$
+BEGIN
+  IF register_patch('0168_CreateCoProposerInvite', 'Jekabs Karklins', 'Adding co-proposer claims', '2025-01-10') THEN
+    BEGIN
+      CREATE TABLE co_proposer_invites (
```
gnyiri, line 6: You can maybe add an "IF NOT EXIST" parameter as usual

apps/backend/src/mutations/InviteMutations.spec.ts
```diff
@@ -30,7 +40,7 @@ describe('Test Invite Mutations', () => {
         email: email,
         note: 'Test note',
         claims: {
-          roleIds: [1],
+          roleIds: [UserRole.USER],
```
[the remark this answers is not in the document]
jekabs-karklins: Yes, thanks

apps/backend/src/auth/InviteAuthorizer.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor(
+    @inject(Tokens.UserAuthorization) private userAuth: UserAuthorization
+  ) {}
+
+  public isRoleInviteAuthorized = async (
+    agent: UserWithRole | null,
+    roleIds?: number[]
+  ) => {
+    if (roleIds === undefined || roleIds.length === 0) return true;
+
```
mutambaraf, line 18: Here if there are no roleIds do we want to authorise.

jekabs-karklins: Yes, it is logically ok to add 0 or no roles in an invite for everyone

mutambaraf (approved):
Looks good to me.

apps/backend/db_patches/0168_AddCoProposerClaims.sql
```diff
@@ -0,0 +1,17 @@
+DO
+$$
+BEGIN
+  IF register_patch('0168_CreateCoProposerInvite', 'Jekabs Karklins', 'Adding co-proposer claims', '2025-01-10') THEN
+    BEGIN
+      CREATE TABLE IF NOT EXISTS co_proposer_invites (
+        co_proposer_invite_id SERIAL PRIMARY KEY,
+        invite_code_id INT NOT NULL REFERENCES invite_codes(invite_code_id) ON DELETE CASCADE,
```
yoganandaness, line 8: Can an Invite Code be used for multiple Proposal? If not, should we create an Unique Key for (invite_code_id, proposal_pk)

apps/backend/src/datasources/CoProposerInviteDataSource.ts
```diff
@@ -0,0 +1,6 @@
+import { CoProposerInvite } from '../models/CoProposerInvite';
+
+export interface CoProposerInviteDataSource {
+  create(inviteCodeId: number, proposalPk: number): Promise<CoProposerInvite>;
+  findByInviteCodeId(inviteCodeId: number): Promise<CoProposerInvite[]>;
```
yoganandaness, line 5: Can a Invite Code used for More than one Proposal? If not, should this function return only one CoProposerInvite instead of multiple?

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -87,6 +162,12 @@ export default class InviteMutations {
     }
   }
 
+  private async setCoProposerInvites(invCodeId: number, proposalPk?: number) {
```
yoganandaness, line 165:
```suggestion
  private async setCoProposerInvites(invCodeId: number, proposalPk: number) {

```

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -87,6 +162,12 @@ export default class InviteMutations {
     }
   }
 
+  private async setCoProposerInvites(invCodeId: number, proposalPk?: number) {
+    if (!proposalPk) return;
```
yoganandaness, line 166: We should not be needing this, as this function should have the proposalPk as the argument mandatorily.

yoganandaness: Probably we can check this condition, where we call this function.

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -67,14 +113,43 @@ export default class InviteMutations {
...
+    if (isCoProposerClaimAuthorized === false) {
+      return rejection(
+        'User is not authorized to update invites to this proposal',
+        { userId: agent?.id, proposalPk: coProposerProposalPk }
+      );
+    }
+
+    const updatedInvite = await this.inviteDataSource.update(args);
+    await this.setRoleInvites(updatedInvite.id, args.claims?.roleIds);
+    await this.setCoProposerInvites(
```
yoganandaness, line 144: I think, This could potentially create duplicated value for the CoProposerInvites. And it could be a good idea to create a unique constraint for the Code and Proposal Pk pair in the DB layer. 

InviteCode1 Proposal1
InviteCode1 Proposal1

apps/backend/src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -232,6 +232,12 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
       });
   }
 
+  async addProposalUser(proposalPk: number, userId: number): Promise<void> {
```
yoganandaness, line 235: Looks like this function has the potential to insert duplicated records of the proposalPk, userId pair, which we should not allow.

apps/backend/db_patches/0168_AddCoProposerClaims.sql
```diff
@@ -0,0 +1,17 @@
+DO
+$$
+BEGIN
+  IF register_patch('0168_CreateCoProposerInvite', 'Jekabs Karklins', 'Adding co-proposer claims', '2025-01-10') THEN
+    BEGIN
+      CREATE TABLE IF NOT EXISTS co_proposer_invites (
+        co_proposer_invite_id SERIAL PRIMARY KEY,
```
yoganandaness, line 7: Do we need this Primary key. If not, we can make the `invite_code_id` and `proposal_pk` pair as the Composite key

yoganandaness (changes requested):
Overall Nice work. I think, i have covered the points that i felt like. Let me know if you find something irrelevant, if so, please comment on it. @jekabs-karklins

yoganandaness (changes requested):
...

apps/backend/db_patches/0168_AddCoProposerClaims.sql
jekabs-karklins: Although I see this primary key as not being used right now, I think that for siplicity and consistance we should keep it, and address it with the other comment about adding a UNIQUE constraing on invite_code_id, proposal_pk

apps/backend/db_patches/0168_AddCoProposerClaims.sql
jekabs-karklins: Thank you, this is a good idea

apps/backend/src/auth/InviteAuthorizer.ts
jekabs-karklins: If by "use case" you mean if scenarios where role_ids are undefined or empty will happen then yes, as you might create an invite without a user role and in that scenario you would not want the Authorizer to fail.

apps/backend/db_patches/0168_AddCoProposerClaims.sql
jekabs-karklins: Okay, I have added it, but I will also need to change the role_invites table for the concistency.

apps/backend/db_patches/0168_AddCoProposerClaims.sql
jekabs-karklins: Since I have added a composite primary key the uniquenes is inherited from it, so there is no need to add the constraint. Thanks

apps/backend/src/datasources/CoProposerInviteDataSource.ts
jekabs-karklins: This just allows to join two proposals with one code. 
I see that the method name expects to return one, so I think you are right, lets restrict it to one, and if we need such invites in the future it will be easy to adjust

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -1,45 +1,91 @@
...
+      roleIds
+    );
+    if (isRoleClaimAuthorized === false) {
+      return rejection(
+        'User is not authorized to create invites to this user type',
+        { userId: agent?.id, roleIds }
+      );
+    }
+
+    const isCoProposerClaimAuthorized =
```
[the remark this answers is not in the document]
jekabs-karklins: Yes, that is the point. As a coproposer you can modify a proposal or invite new coroposers, and the role for them is going to be USER

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -1,45 +1,91 @@
...
+      coProposerProposalPk === undefined ||
+      this.userAuth.isUserOfficer(agent) ||
```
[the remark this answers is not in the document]
jekabs-karklins: Is the emphasis in the statement on the word "should" or on "here" :). 
If it is on "here" then yes, I think we normally perform these kind of validations in Mutation layer. Do you have any speciffic concerns?

apps/backend/src/mutations/InviteMutations.ts
jekabs-karklins: Yes thank you, I have added it

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -87,6 +162,12 @@ export default class InviteMutations {
     }
```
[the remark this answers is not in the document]
jekabs-karklins: Sounds good

apps/backend/src/datasources/postgres/ProposalDataSource.ts
jekabs-karklins: If a duplicate record is added, an exception will be thrown. The check for existing records is performed at the Mutation layer, as it seems more appropriate to handle it there. The database layer is intentionally kept simple and minimal in logic.

apps/backend/src/resolvers/mutations/CreateInviteMutation.ts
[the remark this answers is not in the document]
jekabs-karklins: Thank you for the suggestion! The current design allows an invite to include both role AND co-proposer claims within a single invite. This is to allow with one invite code to configure the user to the desired state.

apps/backend/src/mutations/InviteMutations.ts
jekabs-karklins: Okay, I move the check out to where the functions are called.

apps/backend/src/auth/InviteAuthorizer.ts
```diff
@@ -0,0 +1,25 @@
...
+  public isRoleInviteAuthorized = async (
+    agent: UserWithRole | null,
+    roleIds?: number[]
+  ) => {
+    // If no roleIds are provided, the invite is authorized
+    if (roleIds === undefined || roleIds.length === 0) return true;
+
+    const onlyUserRole = roleIds.length === 1 && roleIds[0] === UserRole.USER;
+    const isUserOfficer = this.userAuth.isUserOfficer(agent);
```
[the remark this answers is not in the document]
jekabs-karklins: This function in plain english means
Is user allowed to invite someone else to have a role (indicated by roleIds)

User officer - any role
Any other user - only USER role

apps/backend/src/mutations/InviteMutations.ts
jekabs-karklins: Done, thanks

yoganandaness (approved):
Looks great :).

PR #918, opened 2025-01-22
fix: handle the deletion of subtemplate answers

deepaksftc:
## Description
This PR introduces a fix to clear the answers of dependent generic template questions when their conditions are not satisfied.

## Motivation and Context
The change is required to maintain the integrity of the data and avoid any confusion or misinformation due to the persistence of answers that are no longer relevant or correct. This improves the user experience and ensures the accuracy of the information presented to the users.

## Changes
1. Introduced new functionality to delete old answers satisfying certain conditions.
2. Extended the `deleteOldAnswers` function to include `answers` and `agent` parameters to better handle the deletion process.
3. Added `deleteSubTemplatesAnswers` function to handle the deletion of subtemplate answers when their parent question's conditions are not met.

## Fixes Issue

https://github.com/UserOfficeProject/issue-tracker/issues/1261


apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -53,9 +66,109 @@ export default class QuestionaryMutations {
...
+            questionId: questionId,
+            proposalPk: proposal?.primaryKey,
+          },
+        },
+        agent
+      );
+
+    if (!genericTemplates || !genericTemplateId) {
+      return;
+    }
```
mutambaraf, line 138: We can put the check for `!genericTemplateId` higher up before any queries since we relay of it for deletion which is the main function of this.

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -53,9 +66,109 @@ export default class QuestionaryMutations {
...
+
+    if (genericTemplateStepQuestions === undefined) {
+      logger.logError('Expected to find sub template step, but was not found', {
+        genericTemplateId,
+        questionaryId,
+      });
+      throw new GraphQLError(
+        'Expected to find sub template step, but was not found'
+      );
+    }
```
mutambaraf, line 154: Maybe since this is a clean up if there is nothing to remove we can just return without throwing an error. But if it is really necessary maybe we can just log it.

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -53,9 +66,109 @@ export default class QuestionaryMutations {
...
+      );
+
+      unsatisfiedDependencyQues.forEach((subTemplateQues) => {
+        const config = subTemplateQues.config as SubTemplateConfig;
+        this.deleteSubTemplatesAnswers(
+          subTemplateQues.question.id,
+          config.templateId,
+          questionaryId,
+          agent
+        );
```
mutambaraf, line 106: We may need to await this deletion before deleting answers l think.

mutambaraf (commented):
Looks good except a few comments.

mutambaraf (approved):

yoganandaness (changes requested):
Looks fine. But can we add an e2e.

yoganandaness (changes requested):
Looks fine. But can we add an e2e.

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -53,9 +66,107 @@ export default class QuestionaryMutations {
...
+
+          if (notSatisfiedQuestions.length != 0) {
+            return true;
+          }
+        }
+      );
+
+      unsatisfiedDependencyQues.forEach(async (subTemplateQues) => {
+        const config = subTemplateQues.config as SubTemplateConfig;
+        await this.deleteSubTemplatesAnswers(
```
jekabs-karklins, line 101: This will not be awaited, because `forEach` does not await the function body statements.
Maybe use `Promise.all` with map or a `for ... of`
This might now show as a problem if you do not expect immediate update latter in the code.

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -53,9 +66,107 @@ export default class QuestionaryMutations {
...
+    }
+
+    const genericTemplateQuestionIds: string[] | undefined =
+      genericTemplateStepQuestions?.map((question) => question.question.id);
+
+    if (!genericTemplateQuestionIds) {
+      return;
+    }
+
+    genericTemplates.forEach(async (genericTemplate) => {
```
jekabs-karklins, line 161: The same issue with the await is here

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -53,9 +66,107 @@ export default class QuestionaryMutations {
...
+    const genericTemplateStepQuestions = templateSteps.flatMap(
+      (step) => step.fields
+    );
+
+    if (!genericTemplateStepQuestions) {
+      return;
+    }
+
+    const genericTemplateQuestionIds: string[] | undefined =
+      genericTemplateStepQuestions?.map((question) => question.question.id);
```
jekabs-karklins, line 155: You are already narrowing the datatype by checkinf if `genericTemplateStepQuestions` is defined in :150, so this can be simplified
```suggestion
    const genericTemplateQuestionIds = genericTemplateStepQuestions.map(
      (question) => question.question.id
    );
```

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -53,9 +66,107 @@ export default class QuestionaryMutations {
...
+              if (answer) {
+                const { value } = JSON.parse(answer.value);
+                if (value[0] !== dependency.condition.params) {
+                  return true;
+                }
+              }
+            }
+          );
+
+          if (notSatisfiedQuestions.length != 0) {
```
jekabs-karklins, line 93: Can we do a strict equals for consistency?
```suggestion
          if (notSatisfiedQuestions.length !== 0) {
```

jekabs-karklins (approved):
LGTM

mutambaraf (approved):

PR #924, opened 2025-01-27
fix: pagination issue with proposal table

deepaksftc:
## Description
This PR addresses a pagination issue with the proposal table by modifying the `onPageChange` function in the `ProposalTableOfficer` component.

## Motivation and Context
The pagination problem in the proposal table was causing data loading and user interface issues. The current implementation was not correctly updating the page size when changing pages, which led to incorrect data being displayed.

## Changes
- Removed the `pageSize` parameter from the `onPageChange` function within the `ProposalTableOfficer` component. This change ensures that the page size remains consistent when navigating through different pages of the proposal table.
- The `onPageChange` function now only updates the `page` parameter in the `setSearchParams` function. This adjustment ensures that the correct data is loaded when the page is changed.

## Fixes Jira Issue

### DIFF
--- a/apps/backend/package-lock.json
+++ b/apps/backend/package-lock.json
@@ -14,7 +14,7 @@
         "@graphql-tools/utils": "^10.1.0",
         "@isaacs/ttlcache": "^1.4.1",
         "@user-office-software/duo-localisation": "^1.2.0",
-        "@user-office-software/duo-logger": "^2.1.1",
+        "@user-office-software/duo-logger": "^2.3.1",
         "@user-office-software/duo-message-broker": "^1.6.0",
         "@user-office-software/duo-validation": "^5.1.11",
         "@user-office-software/openid": "^1.4.0",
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -887,10 +887,8 @@ const ProposalTableOfficer = ({
           Toolbar: ToolbarWithSelectAllPrefetched,
         }}
         onPageChange={(page, pageSize) => {
-          setSearchParams({
-            page: page.toString(),
-            pageSize: pageSize.toString(),
-          });
+          searchParams.set('search', page.toString());
```
yoganandaness, line 890: Can we use the `setSearchParams` instead of mutating the `searchParams` directly. Just following the recommended approach. Correct me if I am wrong.

yoganandaness (changes requested):

yoganandaness (approved):

mutambaraf (approved):

PR #932, opened 2025-02-03
perf: use batch loading for BasicUserDetails

ACLay:
## Description

Expands the user batch loading from #469 to apply to BasicUserDetails as well, since the basic variant is quite widely used, and thus causes a lot of calls to the slow external user service at STFC. By combining multiple external calls into one we'll have to pay the time penalty of making network calls fewer times, and when that's ~100ms for each call in our prod environment, it'll add up to quite a bit of user-data-heavy requests.

## How Has This Been Tested

I made 3 local page loads of access panel 11's assignments, recording the time taken by the main 3 calls when starting with an empty users cache. Due to react's dev setup making calls twice, this results in 6 calls for each. For those runs I also tracked the number of UOWS calls made. All times are in milliseconds

Method | Call time (current client) |   |   |   |   |   | Call time (with batching) |   |   |   |   |   | Average time saved | Percentage improvement
---|---|---|---|---|---|---|---|---|---|---|---|---|---|---
getFap | 87 | 89 | 83 | 83 | 104 | 125 | 66 | 68 | 54 | 55 | 73 | 77 | 29.666 | 31%
getFapMembers | 716 | 735 | 704 | 721 | 674 | 675 | 446 | 564 | 589 | 577 | 622 | 642 | 130.833 | 18%
getFapProposals | 711 | 741 | 702 | 725 | 645 | 674 | 538 | 562 | 593 | 581 | 612 | 644 | 111.333 | 16 %

Attempt | UOWS calls - no batching | UOWS calls - batching
---|---|---
1 | 71 | 38
2 | 86 | 48
3 | 72 | 49

## Fixes

Closes https://github.com/UserOfficeProject/issue-tracker/issues/1197

## Depends on

https://github.com/UserOfficeProject/mock-uows/pull/19

While the test suite as a whole is passing here, some tests in isolation would have failed without this update to the mock uows, which has already merged. The initial passing state was I believe due to requests for individual users being made first in other tests and then cached in the backend. Without that, any test that first requests multiple users with batch loading would have failed as the mock server would only give one user's details.


.vscode/launch.json
```diff
@@ -12,13 +12,13 @@
...
-            "name": "Attach to Windows Docker container",
+            "name": "Attach to Docker container",
             "type": "node",
             "request": "attach",
             "port": 9229,
             "protocol": "inspector",
-            "localRoot": "${workspaceFolder}",
-            "remoteRoot": "C:\\app"
+            "localRoot": "${workspaceFolder}/apps/backend",
+            "remoteRoot": "/app"
```
ACLay, line 21: This gets rid of our old windows container debug profile and replaces it with the newer one that our debug instructions on confluence advise putting here. Once this merges in, that section will need to be updated https://user-office-project.atlassian.net/wiki/spaces/UOP/pages/19136712/STFC+development+workflow#Debugging-the-backend

TCMeldrum (approved):
Looks good to me. Out of interest if the number of calls to the UOWS is cut down by a half does that mean only 2 people are being batch loaded at a time?

ACLay:
Seems like it. I guess the batching might be done at a more local level within each query, like if we're loading proposals on the panel one by one, it's probably not the fastest to batch up and only request the reviewers at the very end. Plus there's likely going to be a lot of repeats in the reviewers requested in the panel, so a lot of requests are already be dealt with and reduced in the initial unbatched case by the caching.

mehta-pooja123 (approved):

janosbabik (approved):

ACLay:
I'd been hoping for more impact to this, but since it would only really affect queries when user cache is empty, I suppose little movement in the bulk of the stats should be expected. Still, looks like a nice improvement for the slowest running queries.

Latency charts from dev for the 2 week window either side of this change going in: 
image

PR #940, opened 2025-02-06
test: split template tests to speed up github actions

ACLay:
The runtime of templatesBasic.cy.ts is ridiculous to the point that it's almost solely responsible for half the runtime of our test suite in github actions, only completing the STFC mode tests in the same time that every other test segment takes to run both.

This PR fixes that by splitting the template deletion & archival tests out to a separate file which we run as a separate segment in the github actions. While doing this, I also moved vast quantities of duplicate setup/context code that was shared between the Basic and Advanced template test files to a single location, as I didn't want to have to make a 3rd copy.

## Fixes

Closes https://github.com/UserOfficeProject/issue-tracker/issues/1330


simonfernandes (approved):

jekabs-karklins (approved):
I like the simple approach here 💯 

Lets merge it in and adjust the merge test requirements to have the new tests as a requirement and delete the old one

PR #947, opened 2025-02-12
fix: refactor event logging to ensure resultKey and event is defined before usage

jekabs-karklins:
## Description
This PR is a tweak to make sure the `resultKey` and `event` is not defined if it is a test environment , because it is not used in test.

## Changes
1. Moved the definition of `resultKey` inside the check for `NODE_ENV !== 'test'`, ensuring it is only defined when it is needed.
2. Adjusted the creation of the `event` object to be inside the check for `NODE_ENV !== 'test'`, as it now relies on the `resultKey` that is defined inside this check.


Scott-James-Hurley (approved):

janosbabik (approved):
nice catch!

PR #953, opened 2025-02-17
feat: experiment safety review schema refactoring

yoganandaness:
## Description

The New Experiment Safety Review process is in progress and the first step towards is to change the DB schema to adapt to the changes and ensure, the existing workflows around the Experiment works as usual. 

Attached with [Database Schema_7630ead669eb4e9e92346d4af197a5f1-120325-1615-20.pdf](https://github.com/user-attachments/files/19212683/Database.Schema_7630ead669eb4e9e92346d4af197a5f1-120325-1615-20.pdf) on the detailed specification of the proposed changes

## Motivation and Context

Implement New Experiment Safety Review process

## How Has This Been Tested

Manually. 

## Changes

This significantly involves DB migration changes and various layers including graphql, datasources and Frontend. 


apps/backend/src/mutations/ExperimentMutation.ts
```diff
@@ -0,0 +1,335 @@
...
+      return rejection('No experiment safety found');
+    }
+
+    const hasAccessRights = await this.experimentSafetyAuth.hasWriteRights(
+      user,
+      experimentSafety.experimentSafetyPk
+    );
+    if (hasAccessRights === false) {
+      return rejection(
+        'User does not have permission to remove samples from the experimemnt',
```
jekabs-karklins, line 233: fix: a typo
```suggestion
        'User does not have permission to remove samples from the experiment',
```

apps/backend/src/mutations/ExperimentMutation.ts
```diff
@@ -0,0 +1,335 @@
...
+        {
+          args,
+        }
+      );
+    }
+
+    return this.dataSource.updateExperimentSample(
+      args.experimentPk,
+      args.sampleId,
+      args.isSubmitted!
```
jekabs-karklins, line 281: suggestion: we should handle the undefined case as well
```suggestion
      args.isSubmitted ?? false
```

apps/backend/src/queries/ExperimentQueries.ts
```diff
@@ -0,0 +1,81 @@
...
+import { ExperimentHasSample } from '../resolvers/types/ExperimentHasSample';
+import { ExperimentSafety } from '../resolvers/types/ExperimentSafety';
+
+@injectable()
+export default class ExperimentQueries {
+  constructor(
+    @inject(Tokens.ExperimentDataSource) public dataSource: ExperimentDataSource
+  ) {}
+
+  @Authorized(Roles.USER)
```
jekabs-karklins, line 20: question: Should we also add User Officer?

apps/backend/src/resolvers/types/ExperimentSafety.ts
```diff
@@ -0,0 +1,117 @@
...
+import { Tokens } from '../../config/Tokens';
+import { ResolverContext } from '../../context';
+import { ExperimentDataSource } from '../../datasources/ExperimentDataSource';
+import { ExperimentSafety as ExperimentSafetyOrigin } from '../../models/Experiment';
+import { ExperimentHasSample } from './ExperimentHasSample';
+import { Proposal } from './Proposal';
+import { Questionary } from './Questionary';
+
+@ObjectType()
+@Directive('@key(fields: "experimentSafetyPk")')
```
jekabs-karklins, line 22: question: This means that we also need to update Scheduler, do we need to make a PR on scheduler as well before this gets merged?

apps/backend/index.ts
```diff
@@ -32,7 +32,7 @@ async function bootstrap() {
     .use(healthCheck())
     .use(readinessCheck())
     .use(exceptionHandler())
-    .use(express.json({ limit: '5mb' }));
+    .use(express.json({ limit: '5mb' })); //todo: Is it redundant?
```
jekabs-karklins, line 35: fix: We can remove this, as it is already done above

jekabs-karklins (changes requested):
I like that this PR reduces proposal and sample safety to just experiment safety. I did left some comments so please look at them and re-request the review. I only checked the code for the frontend, without running it locally, but with next review I will focuss more on the frontend.
Cheers

apps/backend/db_patches/0176_DropOldSampleSafetySchema.sql
```diff
@@ -0,0 +1,11 @@
+DO
+$$
+BEGIN
+  IF register_patch('0176_DropOldSampleSafetySchema', 'Yoganandan Pandiyan', 'Dropping tables involved in old Sample Safety system', '2025-03-08') THEN
+    DROP TABLE IF EXISTS "sample_experiment_safety_inputs";
+    DROP TABLE IF EXISTS "experiment_safety_inputs";
+    DROP TABLE IF EXISTS "scheduled_events";
```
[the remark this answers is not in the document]
yoganandaness: @jekabs-karklins here is the minutes of our meeting
1. We will delete the old tables and add new one. We will not migrate data, bcoz this feature is not stable and is not ready for production. So considering the fact that it is not stable, we believe this is not being used by any other facility. Hence we decide not to do migrations
2. Clear all the records in the references table like feedbacks, visits, shipments
3. We will not rename the referenced column(scheduled_event_id to experiment_pk) but create brand new column

apps/backend/src/datasources/mockups/CallDataSource.ts
```diff
@@ -44,7 +44,7 @@ export const dummyCallFactory = (values?: Partial<Call>) => {
     values?.callReviewEnded || false,
     values?.callFapReviewEnded || false,
     values?.templateId || 1,
-    values?.esiTemplateId || 2,
+    values?.esiTemplateId,
```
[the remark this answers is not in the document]
yoganandaness: Yes. This was to test the scenario 'User should not be able create Experiment Safety for an Experiment, whose Call does not have a Experiment Safety Form Questionary template'. 

But i see your point. I will revert this back and pass undefined to the argument(esiTemplateId)

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
```diff
@@ -0,0 +1,476 @@
...
+            experiment,
+          });
+          throw new GraphQLError('Failed to update experiment');
+        }
+
+        return createExperimentObject(records[0]);
+      });
+  }
+
+  async deleteByScheduledEventId(
```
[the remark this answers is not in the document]
yoganandaness: This is needed. This is not the scheduledEventId in the user office, but the one from scheduler App. 

Context: We have only removed the scheduled events table and replaced it with experiments table. However, the scheduled events table in the scheduler remains same. `Experiments` table in the UO db will be referencing to the `Scheduled Events` table in the scheduler DB

Experiments table in UO will contain experiments
Scheduled Events table in Scheduler will contain all the events including experiments, downtimes, maintenance etc.,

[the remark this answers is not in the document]
yoganandaness: So when a scheduled Experiment is deleted in the Scheduler, a message( with Scheduled Event ID) will be passed to RabbitMQ and we will delete the corresponding experiments using the Scheduled Event ID

apps/backend/db_patches/0173_VisitsUpdate.sql
```diff
@@ -0,0 +1,14 @@
+DO
+$$
+BEGIN
+  IF register_patch('0173_VisitsUpdate', 'Yoganandan Pandiyan', 'Updating the Visits to refer to the new Experiments table', '2025-03-08') THEN
+    DELETE FROM visits;  -- Cleaning up old data
```
jekabs-karklins, line 5: Suggestion: Didn't we want to TRUNCATE?

jekabs-karklins (approved):
Great PR Yoga!

yoganandaness:
> Great PR Yoga!

Thank you for the valuable suggestions.

TCMeldrum (approved):
We are not using this feature at the moment so haven't done a very in depth look, but look at the DB and config changes they should be fine for us as far as I can tell. 

Checking it out locally seems to work fine in stfc mode

PR #958, opened 2025-02-20
1305: caching only when token validation request returns true. To tes…

mehta-pooja123:
To test further. This is done in relation with issue https://github.com/UserOfficeProject/issue-tracker/issues/1305


ACLay (commented):
The thinking behind cacheing the request while it's in progress is we ensure that at any time we're only checking a token against the user office web service once. Say a user tries to load the home page after a long time inactive and their token is expired, we don't want every single graphql request on the homepage to try and independently verify the token at the same time with separate UOWS requests. If we only add the entry to the cache after we have a result then I think that's what would happen.

mutambaraf:
Is this still looked into

PR #967, opened 2025-02-28
fix: handle empty files

mutambaraf:
## Description
This PR introduces a new feature that prevents the uploading and saving of empty files (0 bytes) to the system.

## Motivation and Context
The change is required to enhance the error handling mechanism of our file upload process. Previously, the system allowed empty files to be uploaded and saved, which could lead to data inconsistency and potential errors down the line.

## Changes
1. Added a condition to check the file size in the backend (FileMutations.ts). If the file size is 0 bytes, the system will reject the file and return an error message.
2. Updated the frontend (FileUploadComponent.tsx) to handle the new rejection case. If a user tries to upload an empty file, they will see a rejection message stating "File size is 0 bytes".
3. Updated the error message displayed to the user to include issues with file size, in addition to the existing messages about file type and extension.

## How Has This Been Tested?
Manual testing

## Fixes Jira Issue

Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1326

### DIFF
--- a/apps/backend/src/mutations/FileMutations.ts
+++ b/apps/backend/src/mutations/FileMutations.ts
@@ -19,7 +19,7 @@ export default class FileMutations {
     path: string
   ): Promise<FileMetadata | Rejection> {
     if (sizeImBytes === 0) {
-      return rejection('Could not save file because its 0 bytes', {
+      return rejection('Could not save file because it's 0 bytes', {
         fileName,
         path,
       });
--- a/apps/backend/src/mutations/FileMutations.ts
+++ b/apps/backend/src/mutations/FileMutations.ts
@@ -19,7 +19,7 @@ export default class FileMutations {
     path: string
   ): Promise<FileMetadata | Rejection> {
     if (sizeImBytes === 0) {
-      return rejection('Could not save file because it's 0 bytes', {
+      return rejection('Could not save file because it is 0 bytes', {
         fileName,
         path,
       });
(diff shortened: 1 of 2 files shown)


Scott-James-Hurley (approved):

apps/backend/src/mutations/FileMutations.ts
```diff
@@ -18,6 +18,13 @@ export default class FileMutations {
     sizeImBytes: number,
     path: string
   ): Promise<FileMetadata | Rejection> {
+    if (sizeImBytes === 0) {
+      return rejection('Could not save file because its 0 bytes', {
```
ACLay, line 22:
```suggestion
      return rejection('Could not save file because it's 0 bytes', {
```

apps/frontend/src/components/common/FileUploadComponent.tsx
```diff
@@ -265,8 +275,8 @@ export function NewFileEntry(props: {
             </Avatar>
           </ListItemAvatar>
           <ListItemText
-            primary="Incorrect file type"
-            secondary="Ensure that the file is of the correct type and has an extension."
+            primary={rejectionMessage}
+            secondary="Ensure that the file is of the correct type, size and has an extension."
```
mehta-pooja123, line 279: Do you think we can utilize this opportunity to improve on user experience while they see this msg: 
For example: "Please make sure your file is non-zero size and is a PDF. Make sure the file has .pdf as extension"

jekabs-karklins (approved):
👍🏻

PR #968, opened 2025-03-02
feat: Adding approval for visit registrations

jekabs-karklins:


apps/frontend/src/components/visit/CreateVisitRegistration.tsx
```diff
@@ -18,7 +23,8 @@ function createRegistrationStub(
   return {
     userId: userId,
     registrationQuestionaryId: 0,
-    isRegistrationSubmitted: false,
+    status: VisitRegistrationStatus.DRAFTED,
+    trainingStatus: TrainingStatus.NONE,
```
jekabs-karklins, line 27: @janosbabik i checked why this line was added. This is because we use `trainingStatus` value to render Experiments table.
image

This value was and is obtained from getScheduledEventsCore query
https://github.com/UserOfficeProject/user-office-core/blob/d85906c594dfd98914b38f6248e64b51043d3f8e/apps/frontend/src/graphql/scheduledEvent/getScheduledEventsCore.graphql#L25

Since we now update the VisitRegistration the updated value from server is used to update the state. If the updated value does not contain the same properties, that makes a type error.

I think in the future we might write a resolver for this field and retrieve with our specific TrainingDataSource where we get it from the external system

janosbabik (approved):

apps/e2e/cypress/support/utils.ts
```diff
@@ -205,6 +205,13 @@ const setDatePickerValue = (selector: string, value: string) =>
 const getTinyMceContent = (tinyMceId: string) => {
   cy.get(`#${tinyMceId}`).should('exist');
 
+  cy.window().should('have.property', 'tinymce'); // wait for tinyMCE
```
jekabs-karklins, line 208: Note: this part is not related to changes but was the cause for flaky test

TCMeldrum (approved):

PR #975, opened 2025-03-04
feat: user invites frontend

jekabs-karklins:
## Description
This PR introduces the feature to invite users to collaborate on a proposal, right from the frontend. 
The feature needs to be explicitly enabled in the application settings.

Peek 2025-03-04 22-18

## Motivation and Context
This feature is necessary to enhance the collaboration experience among users while working on proposals. It provides an easy way to invite other users to join and contribute to a proposal.

## Changes
- Added the ability to invite users in the Proposal file.
- Created new end-to-end tests for the invite feature in Cypress.
- Integrated the invite feature to the frontend for a seamless user experience.

## How Has This Been Tested?

Added E2E tests

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4368](https://jira.esss.lu.se/browse/SWAP-4368)


jekabs-karklins:
@UserOfficeProject/stfc-user-programme Could you please take a look at this PR?

apps/frontend/src/components/proposal/ProposalParticipants.tsx
```diff
@@ -37,39 +46,73 @@ const Participants = ({
...
-    const newUsers = users.filter((u) => u.id !== user.id);
-    setUsers(newUsers);
+  const handleAddParticipants = (props: {
+    users: BasicUserDetails[];
+    invites: Invite[];
+  }) => {
+    setUsers([...users, ...props.users]);
+    setInvites([...invites, ...props.invites]);
+    setOpen(false);
   };
```
mutambaraf, line 74: Are these two function not doing the same or almost similar function the difference is maybe on invites .Maybe using one function for both legacy and the new flow.

apps/frontend/src/components/proposal/ProposalParticipants.tsx
```diff
@@ -79,41 +122,78 @@ const Participants = ({
...
+                  color="secondary"
+                  icon={invite.isEmailSent ? <SendIcon /> : <ScheduleSend />}
+                  size="small"
+                  label={invite.email}
+                  key={invite.email}
+                  onDelete={() => handleDeleteInvite(invite)}
+                />
+              ))}
+            </Box>
+          ) : null}
```
mutambaraf, line 179: Maybe we can use the short hand && rather than returning null.

mutambaraf (commented):
Looks good accept some few comments.

apps/frontend/src/components/proposal/ProposalParticipants.tsx
jekabs-karklins: Hi, Thanks, I'm keeping it DRY, now. I made addUsers to reference handleParticipants with empty initializer for invites. If we are happy with tne new flow, the fleanup PR will remove the clean up the addUsers()

apps/frontend/src/components/proposal/ProposalParticipants.tsx
jekabs-karklins: Thanks

apps/e2e/cypress/e2e/invites.cy.ts
```diff
@@ -0,0 +1,168 @@
...
+      cy.login('user1', initialDBData.roles.user);
+      cy.visit('/');
+      cy.contains('New Proposal').click();
+      cy.get('[data-cy=call-list]').find('li:first-child').click();
+
+      cy.get('[data-cy="add-participant-button"]').click();
+
+      cy.get('[data-cy="invite-user-autocomplete"]').type(email);
+      cy.get('[role=menuitem]').contains(lastName);
+    });
```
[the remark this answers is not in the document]
jekabs-karklins: I think I shoud have add some more info to the PR description, but here it goes.
The add proposer window gets proposer info in 3 ways depending on what is in the query (text input field)
1. If query is empty it will get latest collaborators
2. If partial user information is entered it will call "getUsers", and return results, which might vary depending on how facility has implemented that endpoint.
3. If email is entered, it will try to get user by exact email if found it will show the user, if not it will show the "invite by email" button if the invite by email is enabled, otherwise it will so that no results found

This test tests the case 3. Inviter should be able to invite user knowing his exact email. So this test is related to the invites.

PR #990, opened 2025-03-11
refactor: keep fap proposals loaded between tab changes

TCMeldrum:
Close https://github.com/UserOfficeProject/issue-tracker/issues/1319

## Description
FAP secretaries reported that they navigated between the tabs regularly, and reloading the proposals data can be long (we have panels of up to 130 proposals). This change keeps the proposals loaded to make navigating between tabs more seamless. 

## Motivation and Context
This change was required to improve the user experience by preventing the reloading of proposals every time a user navigates away and then returns to the FAP Proposals page. This not only saves time but also reduces unnecessary network requests, thus improving the overall performance of the application.

## Changes
1. Introduced `useFapProposalsData` hook to manage the state of loaded proposals in the `FapPage` component.
2. Updated `FapProposalsAndAssignmentsView` component to accept `fapProposals` as a prop.
3. Replaced the direct use of `useFapProposalsData` hook with the `fapProposals` prop in `FapProposalsAndAssignmentsTable` component.
4. Refactored `useFapProposalsData` hook to return an object of type `FapProposals` which includes loading state, data, and a setter function for the loaded proposals.

## Fixes Jira Issue


Scott-James-Hurley:
Proposals no longer need to be reloaded which is great but I wonder if there's a way to keep the page length and number the user had last selected on the `Proposals and Assignments` table between tab switching. I think this would go a long way to helping users find where they were again.

yoganandaness (approved):
Looks good. 

I am not sure if this is covered with e2e. If not, could you plan on that?

TCMeldrum:
> Looks good.
> 
> I am not sure if this is covered with e2e. If not, could you plan on that?

@yoganandaness This change is to make less graphql requests when navigating between tabs, I think there could be a way of testing this but I can't find one that works. I think you might be able to do it with `cy.intercept` but I can't figure it out. Any suggestions?

yoganandaness (approved):
Looks good to me. E2E looks fine as well. I dont think we need to test the data loading part, as this seems beyond the scope of e2e to me. 

Good work.

apps/e2e/cypress/e2e/FAPs.cy.ts
```diff
@@ -454,6 +454,28 @@ context('Fap reviews tests', () => {
...
+
+      cy.contains('Documents').click();
+
+      cy.contains('Proposals and Assignments').click();
+
+      cy.contains(proposal1.title)
+        .parent()
+        .find('[type="checkbox"]')
+        .should('be.checked');
+    });
```
Scott-James-Hurley, line 477: Thanks for taking on my suggestion. Could you add checks that the page number and length are saved as well?

Scott-James-Hurley (approved):
Thanks for amending the test, much appreciated. Just one wee thing I noticed.

PR #1006, opened 2025-03-24
fix: add null check on fields which can be null for some legacy proposals

Bhaswati1148:
## Description
This PR adds a null check for date fields - `endCall`, `endCallInternal`, `endFapReview`, `endReview` when executing the async job 
`checkAllCallsEnded`, `checkCallsReviewEndedJob` and `checkCallsFAPReviewEndedJob`

## Motivation and Context
It was found that ISIS_Direct 2024_2 did not move automatically on the proposal workflow from FAP_REVIEW to Submitted Locked status because it could not find the date to end the FAP review period for some of the calls. These calls were set up for legacy proposals, so were added with minimal data. 
image

## How Has This Been Tested
Manually tested the changes in local.

## Fixes

Closes https://github.com/UserOfficeProject/issue-tracker/issues/1102
## Changes

- apps/backend/src/asyncJobs/jobs/checkAllCallsEnded.ts
- apps/backend/src/asyncJobs/jobs/checkCallsFAPReviewEnded.ts
- apps/backend/src/asyncJobs/jobs/checkCallsReviewEnded.ts

### DIFF
--- a/apps/backend/src/asyncJobs/jobs/checkAllCallsEnded.ts
+++ b/apps/backend/src/asyncJobs/jobs/checkAllCallsEnded.ts
@@ -16,7 +16,7 @@ const checkCallsEnded = async (dataSource: CallDataSource) => {
     const currentDate = new Date();
 
     const callsThatShouldEnd = notEndedCalls.filter(
-      (notEndedCall) => notEndedCall.endCall.getTime() <= currentDate.getTime()
+      (notEndedCall) => notEndedCall.endCall?.getTime() <= currentDate.getTime()
     );
 
     for (const callThatShouldEnd of callsThatShouldEnd) {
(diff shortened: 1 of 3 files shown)


apps/backend/src/asyncJobs/jobs/checkAllCallsEnded.ts
```diff
@@ -16,7 +16,7 @@ const checkCallsEnded = async (dataSource: CallDataSource) => {
     const currentDate = new Date();
 
     const callsThatShouldEnd = notEndedCalls.filter(
-      (notEndedCall) => notEndedCall.endCall.getTime() <= currentDate.getTime()
+      (notEndedCall) => notEndedCall.endCall?.getTime() <= currentDate.getTime()
```
jekabs-karklins, line 19: Question: I have a question. How can it be that the endCall in your database is null. When I look at the DB schema I see that endCall field can not be null.
```
CREATE TABLE public."call" (
	call_id serial4 NOT NULL,
	call_short_code varchar(20) NOT NULL,
	start_call timestamptz NOT NULL,
	end_call timestamptz NOT NULL,
	...
```

Bhaswati1148: Added `?` for an extra check but it seems to be an overhead, as the DB schema already handles null check. I shall revert the changes.

jekabs-karklins (approved):

TCMeldrum (approved):

PR #1011, opened 2025-03-27
chore: remove un-need npm dependency in frontend

TCMeldrum:
Looks like at some point npm got added to the dependencies part of the frontend package.json when I believe it should just be in the engines section like in the backend and e2e packages? 

When building locally, this npm version was out date and causing failures.


jekabs-karklins (approved):

jekabs-karklins:
💯

Bhaswati1148 (approved):

PR #1017, opened 2025-04-02
refactor: update RabbitMQ configuration initialization for dev environment

jekabs-karklins:
## Description
This PR refactors the RabbitMQ configuration initialization for the development and production environment.

## Motivation and Context
The change is needed to make the RabbitMQ queue, exchange and binding definitions more clear.
Previously, queues and their bindings were defined in the code "messageBroker.tsx" file whenever the application was started, which made the user office core knew about the scheduler.
Now the setup happens outside, keeping the core cleaner.

## Changes
1. RabbitMQ configuration initialization has been updated in both `docker-compose-stfc.e2e.yml` and `docker-compose.yml` files to include `rabbitmq-definitions.json` and a command to import definitions.

2. The `example.development.env` file has been pruned to remove redundant RabbitMQ environment variables.

3. A new `rabbitmq-definitions.json` file has been created to define RabbitMQ queues, exchanges and bindings.

4. The `messageBroker.ts` file in the backend app has been cleaned up.

5. Updated the `configuration.md` documentation file to reflect the changes in RabbitMQ configuration.

### Required changes in production
If you rely on RabbitMQ 
- make sure that the queue with name specified in env variable `RABBITMQ_CORE_QUEUE_NAME` exists and is bound to exchange that produces messages intended for core
- make sure  that the exchange with the name specified in env variable `RABBITMQ_CORE_EXCHANGE_NAME` exists and is bound to the queues that are intended to receive messages from core

## How Has This Been Tested?

Tested by running the application and making sure the queues, exchanges and bindings are properly configured. 

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4540](https://jira.esss.lu.se/browse/SWAP-4540)


janosbabik (changes requested):
If I understand correctly  based on these changes, this does not only affect the dev. environment, it is also a breaking change due to the environment name changes. Am I right?

I am not sure about the environment variable names changing. it seemed clearer that this is related to the scheduler messages.

Also, without binding the queue to the exchange, messages will not be delivered.

Could you update the description with the steps required to upgrade to this version?

Thanks!

jekabs-karklins:
> If I understand correctly based on these changes, this does not only affect the dev. environment, it is also a breaking change due to the environment name changes. Am I right?
> 
> I am not sure about the environment variable names changing. it seemed clearer that this is related to the scheduler messages.
> 
> Also, without binding the queue to the exchange, messages will not be delivered.
> 
> Could you update the description with the steps required to upgrade to this version?
> 
> Thanks!

The current setup is quiet misleading and there seems to be confusion between queues and exchanges.
See this line
` const EVENT_SCHEDULING_QUEUE_NAME = process.env.RABBITMQ_SCHEDULER_EXCHANGE_NAME as Queue;
`
This results that on staging we have queue with extension ".fanout" which itself is connected to a actual fanout with he same name.

Additionally, the environment parameter `RABBITMQ_EVENT_SCHEDULING_QUEUE_NAME` appears only in the documentation and isn’t used in the code.

Furthermore, while RabbitMQ supports service decoupling, we aren’t taking full advantage of it. The scheduler queue name is hardcoded in messagebroker.ts, which causes the queues to be created even if the scheduler isn’t used. Additionally only some bindings—primarily those used in ESS—are specified, leading to further confusion about the remaining bindings.

Scott-James-Hurley:
Please correct me if I'm wrong, STFC, but I don't think we need to adjust our environment to accommodate these changes.

Scott-James-Hurley:
Could the `breaking:change` label please be added to this PR to raise awareness of it?

jekabs-karklins:
> If I understand correctly based on these changes, this does not only affect the dev. environment, it is also a breaking change due to the environment name changes. Am I right?
> 
> I am not sure about the environment variable names changing. it seemed clearer that this is related to the scheduler messages.
> 
> Also, without binding the queue to the exchange, messages will not be delivered.
> 
> Could you update the description with the steps required to upgrade to this version?
> 
> Thanks!

@janosbabik 
When we merge this we need to 
1. add a new ENV to user office backend called 
`RABBITMQ_CORE_QUEUE_NAME = user_office_backend.queue` (optional as matches the default value)

2. Add new binding 
 user_office_scheduler_backend.fanout => user_office_backend.queue (mandatory)
 
 Side note i noticed that current bindings defined by duo-backend has no effect so this has low risk of breaking anything new
```js
const EVENT_SCHEDULING_QUEUE_NAME = process.env
    .RABBITMQ_SCHEDULER_EXCHANGE_NAME as Queue;
  const SCHEDULER_EXCHANGE_NAME = process.env.RABBITMQ_SCHEDULER_EXCHANGE_NAME;

  const rabbitMQ = await getRabbitMQMessageBroker();
  rabbitMQ.addQueueToExchangeBinding(
    EVENT_SCHEDULING_QUEUE_NAME,
    SCHEDULER_EXCHANGE_NAME
  );

```

this code evaluates to
```js
  rabbitMQ.addQueueToExchangeBinding(
   "user_office_scheduler_backend.fanout",
    "user_office_scheduler_backend.fanout"
  );
```

jekabs-karklins:
> I see that messages are still posted to the queue our downstream software uses so I'm satisfied I but I'll double check with others.

@alexyvanot yes, the downstream queue stays untouched 👍🏻

janosbabik (approved):

PR #1019, opened 2025-04-03
fix: bugfix for mistakenly invoking  rabbitMQ status action when emailAction is selected

jekabs-karklins:
## Description
This PR refactors the existing email action handler and prevents it from publishing messages to RabbitMQ.

## Motivation and Context
This change is required to address the bug.

## Changes
1. Created a new LoggingHandler and added it to the dependency configuration.
2. Refactored the email action handler to call the new LoggingHandler and the existing EmailEventHandler directly, instead of publishing messages to RabbitMQ.
3. Refactored the event handlers to use the new LoggingHandler and removed the now-unused createPostToQueueHandler function.

Previously, the EmailActionHandler was sending an event tothe  event bus to reach the logging handler and add some logs, but as a side-effect it reached postToQueueHandler which was set to listen to the same event
image

The logic has been changed so that EmailActionHandler does not post a message into the event bus but sends it directly to emailHandler and loggingHandler. This should preserve the current behavior while making sure the message is not sent to the postToQueue Handler.
image

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4612](https://jira.esss.lu.se/browse/SWAP-4612)

### DIFF
--- a/apps/backend/src/config/Tokens.ts
+++ b/apps/backend/src/config/Tokens.ts
@@ -16,6 +16,7 @@ export const Tokens = {
   InviteDataSource: Symbol('InviteDataSource'),
   TechniqueDataSource: Symbol('TechniqueDataSource'),
   ListenToMessageQueue: Symbol('ListenToMessageQueue'),
+  LoggingHandler: Symbol('LoggingHandler'),
   MailService: Symbol('MailService'),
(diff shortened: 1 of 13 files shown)


janosbabik (approved):
👍

TCMeldrum (approved):

PR #1024, opened 2025-04-10
feat: allow proposals cloned on call with different template

jekabs-karklins:
## Description
This PR introduces the feature to allow proposals to be cloned on a call with a different template.

## Motivation and Context
This change was required to enhance the flexibility of the proposal cloning process. Previously, it was not possible to clone a proposal to a call whose template was different. This limitation has been removed with this PR, enabling more efficient and diverse use of the proposal cloning feature.

## Changes
1. The `clone` method in the `QuestionaryDataSource` interface now accepts an optional `targetTemplateId` parameter. This allows the clone operation to target a different template.
2. The `clone` method in the `ProposalMutations` class no longer throws an error when the `call.templateId` is different from the `sourceQuestionary.templateId`. It now passes the `call.templateId` to the `clone` method of `questionaryDataSource`.
3. The `CallSelectModalOnProposalsClone` component no longer filters calls based on the templateId of the proposal to be cloned. This change allows the user to select a call with a different template when cloning a proposal.

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-2254](https://jira.esss.lu.se/browse/SWAP-2254)


apps/backend/src/datasources/postgres/GenericTemplateDataSource.ts
```diff
@@ -44,6 +44,7 @@ export default class PostgresGenericTemplateDataSource
 
     const newQuestionary = await this.questionaryDataSource.clone(
       sourceGenericTemplate.questionaryId,
+      undefined,
```
mutambaraf, line 47: Maybe we can swap these in clone() function or pass them as an object to avoid passing undefined.

mutambaraf: That is great.

yoganandaness (approved):

PR #978, opened 2025-03-06
feat: allow for bulk reassignement of tech reviews

TCMeldrum:

PR #1028, opened 2025-04-15
fix: deleted the outdated patch, that led to error

yoganandaness:
deleted the outdated patch, that led to error, since it is depending on the table that does not exist. replaced the outdated one with newer one with latest run number

## Description

After merging the PR https://github.com/UserOfficeProject/user-office-core/pull/978, the system started to fail while running db patches. Apparently, this is bcoz of the db patch file, that became outdated due to increased commit activities. As it became outdated, the patch has gone unaware of the table rename(proposal_statuses to statuses) and hence the deployment of systems started failing. As an effort of quick fix, the patch has been renamed to make it latest in the chronological order. As a long fix, we will discuss further on how to tackle this situation. 

## How Has This Been Tested

Manually

### DIFF
--- a/apps/backend/db_patches/0177_AddInstrumentIdToTechRevInPropsalTableView.sql
+++ b/apps/backend/db_patches/0177_AddInstrumentIdToTechRevInPropsalTableView.sql
@@ -1,7 +1,7 @@
 DO
 $$
 BEGIN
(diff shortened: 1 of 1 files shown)


janosbabik (approved):

jekabs-karklins (approved):
LGTM

PR #1030, opened 2025-04-15
feat: add DB constraint instrumentID not null for technical_review

jekabs-karklins:
## Description

It should not be possible to set `instrument_id` to `null` in `technical_review` table. When `technical_review` is crated it does not allow null values, and backend model `TechnicalReview` expects this value to not be `null`. If it is `null` the graphql exception will be thrown.

This PR makes sure that inconsistent data can not be stored in the database. 

Note: while investigating I found out that there is unnecessary optional chaining in the frontend `apps/frontend/src/components/review/TechnicalReviewSummary.tsx` and given the small scope of this PR i took a liberty to address it with this PR.

## Motivation and Context

Improve data integrity

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4582](https://jira.esss.lu.se/browse/SWAP-4582)

### DIFF
--- a/apps/backend/db_patches/0177_AddTechnicalReviewInstrumentNotNull.sql
+++ b/apps/backend/db_patches/0177_AddTechnicalReviewInstrumentNotNull.sql
@@ -1,0 +1,13 @@
+DO
+$$
+BEGIN
+  -- Check if there are any rows with NULL instrument_id
+  IF EXISTS (SELECT 1 FROM technical_review WHERE instrument_id IS NULL) THEN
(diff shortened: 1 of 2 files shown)


Scott-James-Hurley (approved):

jekabs-karklins:
> Remeber this coming in #978 should we also make instrument non nullable in apps/backend/src/resolvers/types/ProposalView.ts

This change was about making sure that instrument ID is not null for technical review
As for the proposal, i think not every proposal has an instrument. As far as I see we add the entry in instrument_has_proposals once proposal is assigned an instrument.

TCMeldrum (approved):

PR #1034, opened 2025-04-23
fix: add more validation on instrument picker on the frontend

mutambaraf:
## Description
This PR introduces additional validation for the instrument picker on the frontend.

## Motivation and Context
This change is required to ensure that the selected instrument in the instrument picker exists in the configuration which has been mainly caused by cloned proposals where the call the proposal is being cloned doesn't have the previous selected instrument. . This also prevents the possibility of selecting an instrument that doesn't exist or has been removed, thereby improving the reliability and robustness of the instrument picker.

## Changes
1. Added validation in `InstrumentPickerDefinition.tsx` that checks if the selected instrument exists in the configuration.
2. Introduced similar validation in `QuestionaryComponentInstrumentPicker.tsx` and ensured that the component's state is updated correctly based on the validation.
3. Updated the `handleOnChange` and `handleTimeInput` functions in `QuestionaryComponentInstrumentPicker.tsx` to incorporate the new validation.

## Fixes Jira Issue

Part of: https://github.com/UserOfficeProject/issue-tracker/issues/1192


apps/frontend/src/components/questionary/questionaryComponents/InstrumentPicker/InstrumentPickerDefinition.tsx
```diff
@@ -27,7 +27,28 @@ export const instrumentPickerDefinition: QuestionaryComponentDefinition = {
     questionRenderer: defaultRenderer.questionRenderer,
   },
   createYupValidationSchema: instrumentPickerValidationSchema,
-  getYupInitialValue: ({ answer }) => answer.value || null,
+  getYupInitialValue: ({ answer }) => {
+    if (answer && answer.config) {
+      const config = answer.config as InstrumentPickerConfig;
+      if (Array.isArray(answer.value)) {
```
Scott-James-Hurley, line 33: Is there a way to avoid repeating this code?

mutambaraf:
> Would it be possible to add a test that tries to select an instrument that's not in the configuration? I could see that test being very verbose, though.

I can use a test that use a proposal clone to a call with out the instrument and see if it throws an error.

yoganandaness (approved):
Nice work.

mutambaraf:
> I cant understand, what would be the error message user sees?

We are just removing the instruments is they are not available in the proposal call 
image
which will then cause a validation error 
image
With the current setup the use was able to save and continue even if the instrument was not available on the call.

Scott-James-Hurley (approved):

PR #1035, opened 2025-04-24
fix: disable DataLoader internal caching

deepaksftc:
Hotfix release:
The user details were also being cached internally by a dependency DataLoader, apart from our normal caches. And these record seems to be staying there for the entire time the particular pod/app is up. This PR is to disable this internal caching for now, as this will not be reflecting any user details changes (first name, lastname etc) until a pod/app is restarted.


yoganandaness (approved):
Seems fine

mutambaraf (approved):

PR #1037, opened 2025-04-24
feat: add 'cancelled' state to action buttons and update related tests

jekabs-karklins:
## Description
This PR introduces the 'cancelled' state to the action buttons and uses it for the visit registration.

## Motivation and Context
The change is required to provide a visual representation of a 'cancelled' state in the visit registration, as it wasn't very clear why is the icon disabled.

Additionally the hovering the icon now shows the tool tip, which was not visible before.

image
image

## How Has This Been Tested?

Manually and updated e2e tests.

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4649](https://jira.esss.lu.se/browse/SWAP-4649)


Scott-James-Hurley (approved):
We don't use visits functionality but lgtm

PR #1040, opened 2025-04-28
fix: api keys now checked before use of authorizers

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1103

## Description
This PR improves the security and efficiency of our application by ensuring API keys are verified before the use of authorizers.

I was unsure weather to put these api checks in the autherisers or straight into the query/mutation. I went with the latter as the authorisers could be used else where in the system and I wanted the api key check to have only occered when the `@Authorized()` decorater (and therefore verifyied the api key) was also run. 

## Motivation and Context

This is to ensure that when we assign a permission to a api token it definatly has access to that endpoint. Sometime apikey would not have access as the function would do a check with the authorisers which the api key would fail as for instance the api is not a user officer. 

## Changes
1. Removed redundant ProposalAuthorization import from the proposal.ts file.
2. Removed updateCallValidationSchemas import from CallMutations.ts as it was no longer needed.
3. In the ExperimentMutation.ts file, the user parameter was replaced with the agent parameter. This change ensures that the API key is verified before the authorizer is used. This change was propagated throughout the file, affecting various methods such as createOrGetExperimentSafety, submitExperimentSafety, addSampleToExperiment, etc.

## Fixes Jira Issue


yoganandaness:
@TCMeldrum This looks good. As i understood the changes(Correct me if i am wrong), we have added the API Token Access Check to ensure that any request with API access token with proper permission should be able to do the operation completely. Currently, even though we have Authorizer and that allows the API Access Token Requests, the logical layer(Queries, Mutations) still could terminate the operation as the Write Access or Read Access gets failed. It make sense to exempt the API Access Token Requests to undergo the Read/Write checks, same as User Officer. 

Is my understanding correct?

Looking at the Motivation section, it emphasize on the Security threat, which i am a bit confused on where we had the encounter. 

The Motivation section of this PR says that

The necessity of this change arose from the discovery that API keys were not being checked prior to the use of authorizers, **potentially leading to unauthorized access and inefficiencies**. This PR addresses this issue and enhances the overall security and performance of the application.

I would like to understand that did we have any situations that lead to Unauthorized access before for the API Access Token requests? If so, could you cite them?

TCMeldrum:
@yoganandaness Sorry about that I used the AI PR description generorater then forgot to update it. You are correct this is not to do with security but fix that miss match between the Authorizer and the logic layer. Sorry about that

yoganandaness:
> @yoganandaness Sorry about that I used the AI PR description generorater then forgot to update it. You are correct this is not to do with security but fix that miss match between the Authorizer and the logic layer. Sorry about that

No worries. Thanks for the clarification.

apps/backend/src/queries/TechnicalReviewQueries.ts
```diff
@@ -59,24 +64,25 @@ export default class TechnicalReviewQueries {
...
           technicalReviews.splice(index, 1);
         }
       })
     );
 
-    const isReviewerOfProposal = await this.proposalAuth.isReviewerOfProposal(
-      agent,
-      proposalPk
-    );
+    const isReviewerOfProposal =
```
Scott-James-Hurley, line 83:
```suggestion
    const hasRights =
```

Scott-James-Hurley (approved):

PR #1043, opened 2025-04-29
feat: implement visit registration cancellation and approval features

jekabs-karklins:
## Description
This PR introduces visit registration cancellation.

## Motivation and Context
This change was necessary to give users the ability to cancel their visit registrations.
This allows user to cancel their registration and let the facility know that the user is not coming.
image

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4651](https://jira.esss.lu.se/browse/SWAP-4651)


jekabs-karklins:
> The cancel visit button is accessed via the ellipses, right? Is it possible a user might miss that and assume they can't cancel their visit?
Good point, and we thought about it. In practice we do not expect this to be the main use case. Showing a red button during every step might be too inviting for user to cancel the visit from a design perspective. We should probably run some focus groups if user can find the cancel, but ellipses is sort of following external design consistencies with other platforms.

janosbabik (approved):

PR #1052, opened 2025-05-08
feat: add open telemetry on requests and logs

mutambaraf:


mutambaraf:
> Did some reading up and I think this looks good, but where is the `OTEL_EXPORTER_OTLP_ENDPOINT` variable populated?

We will pass them as envs

jekabs-karklins (approved):
LGTM

PR #1054, opened 2025-05-08
POC:  Dynamic role creation

bolmsten:
## Description
This PR introduces dynamic role creation in the system, it is very much a proof of concept but the basic functionality works. A new role can be created and a certain set of data access and permissions can be attached. The permissions are the same as are set for the API token and data access is based upon instruments. One could imagine to have a broader set of data access such as techniques or call. 

## Motivation and Context
The purpose of this change is to provide a more flexible and scalable user permissions management system. This will allow roles to be created dynamically, along with their associated permissions and data access rules, thereby enhancing the system's ability to adapt to changing requirements.

## Changes
1. Modified the 'roles' table in the database to include 'permissions' and 'data_access' fields.
2. Updated the UserDataSource interface and its mock implementation to include methods for creating, updating, and deleting roles.
3. Implemented checks for data access permissions in the ProposalAuthorization class. This includes retrieving the instruments associated with a proposal and checking if the user's current role has data access rights for those instruments.
4. Updated the Role model and its usage across various files to include the new 'permissions' and 'data_access' fields.

## Screenshots

Screenshot 2025-05-08 at 15 15 06
Screenshot 2025-05-08 at 15 14 50
Screenshot 2025-05-08 at 15 14 36
Screenshot 2025-05-08 at 15 14 20

### DIFF
--- a/apps/backend/db_patches/0180_CreateUserRoles
+++ b/apps/backend/db_patches/0180_CreateUserRoles
@@ -1,0 +1,17 @@
+DO
+$$
+BEGIN
+	IF register_patch('0180_RenameXpressFeature.sql', 'Fredrik Bolmsten', 'Create dynamic user roles', '2025-05-06') THEN
+	BEGIN
+
+    ALTER TABLE roles
+    ADD COLUMN permissions TEXT[] NOT NULL DEFAULT '{}';
+
+    ALTER TABLE roles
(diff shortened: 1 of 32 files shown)


yoganandaness:
@bolmsten @janosbabik @jekabs-karklins 

Closing this, as we are likely to go in different direction. This PR still could be referenced, if we are reverting back to implementing RBAC

PR #1060, opened 2025-05-13
fix: amend code to allow fap specific proposal count

RasmiaKulan:
## Description
Closes [#1398](https://github.com/UserOfficeProject/issue-tracker/issues/1398)
## Description

This change updates the logic of how proposal counts are displayed for FAP members. Previously, the proposal count was being aggregated across multiple FAPs, causing the same member to show the combined proposal count across all FAPs they belong to. With this update, the proposal count is now FAP-specific, ensuring that each FAP's proposal count is correctly displayed for the respective member.

image

## Motivation and Context

This change is required because the previous logic was incorrectly summing proposal counts across multiple FAPs. For example, if a user is assigned 2 proposals in FAP1 and 3 proposals in FAP2, the system was showing 5 proposals in both FAP1 and FAP2, which was inaccurate. The new logic ensures the proposal count is FAP-specific, reflecting the correct number of proposals for each FAP a user is part of.

### DIFF
--- a/apps/backend/src/datasources/FapDataSource.ts
+++ b/apps/backend/src/datasources/FapDataSource.ts
@@ -83,7 +83,10 @@ export interface FapDataSource {
   getFapProposalCount(fapId: number): Promise<number>;
   getCurrentFapProposalCount(fapId: number): Promise<number>;
   getFapReviewerProposalCount(reviewerId: number): Promise<number>;
-  getCurrentFapReviewerProposalCount(reviewerId: number): Promise<number>;
+  getCurrentFapReviewerProposalCount(
+    reviewerId: number,
+    fapId: number
+  ): Promise<number>;
   getFapProposal(
     fapId: number,
     proposalPk: number,
--- a/apps/backend/src/datasources/mockups/FapDataSource.ts
+++ b/apps/backend/src/datasources/mockups/FapDataSource.ts
@@ -358,7 +358,7 @@ export class FapDataSourceMock implements FapDataSource {
     return dummyFapProposals.length;
   }
 
-  async getCurrentFapReviewerProposalCount(reviewerId: number) {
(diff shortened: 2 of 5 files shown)


mutambaraf (commented):
Is this for API only or the count was already included in the frontend.

Scott-James-Hurley:
> Is this for API only or the count was already included in the frontend.

The count is shown on the members tab of the fap page, in the reviewers table.

RasmiaKulan:
@mutambaraf From my understanding the frontend was already set up to display `proposalsCountByCall` per FAP member, but the backend was returning a cumulative count across all FAPs for that user. After updating the API to filter by fapId, the frontend reflected the correct, FAP-specific count.

RasmiaKulan:
@Scott-James-Hurley These changes were requested by Emma Gozzard. She found this bug while setting up the FAP panel for the round next week.

mutambaraf:
> @mutambaraf From my understanding the frontend was already set up to display `proposalsCountByCall` per FAP member, but the backend was returning a cumulative count across all FAPs for that user. After updating the API to filter by fapId, the frontend reflected the correct, FAP-specific count.
There maybe a mismatch of naming on the fields of this end point.

mutambaraf (approved):

RasmiaKulan:
> There maybe a mismatch of naming on the fields of this end point.

Yes youre right - the parameter is called `reviewerId`, but it's actually the same as the user's `userId` in this context. I think the name is set for clarity on the frontend side, since we refer to them as reviewers there.

Scott-James-Hurley:
Cool. I think the original intent behind the count was that it would inform those assigning reviewers the current workload of a reviewer across all calls currently in the review process.

Scott-James-Hurley (approved):

PR #1066, opened 2025-05-16
feat: replace non-printable characters within the proposal's abstract field. #1370

bashanlam:
Closes [https://github.com/UserOfficeProject/issue-tracker/issues/1370](https://github.com/UserOfficeProject/issue-tracker/issues/1370)

## Description
Replace any non-printable characters found within the proposal's abstract field with a space character. 

## Motivation and Context
Non-printable characters in the abstract field will be replaced to prevent issues with interacting systems.  

## How Has This Been Tested
Manually tested the change. 

## Changes

Update the input page to replace any non-printable characters found within the proposal's abstract field with a space character.


Scott-James-Hurley:
I think it might be helpful to log what what's being changed, just in case we get a support query about why an abstract is different from what the user provided.

bashanlam:
> I think it might be helpful to log what what's being changed, just in case we get a support query about why an abstract is different from what the user provided.

Revert logging changes.

apps/frontend/src/components/questionary/questionaryComponents/ProposalBasis/QuestionaryComponentProposalBasis.tsx
```diff
@@ -116,8 +116,16 @@ function QuestionaryComponentProposalBasis(props: BasicComponentProps) {
           name={`${id}.abstract`}
           label="Proposal Abstract"
           inputProps={{
-            onChange: (event: ChangeEvent<HTMLInputElement>) =>
-              setLocalAbstract(event.target.value),
+            onChange: (event: ChangeEvent<HTMLInputElement>) => {
```
jekabs-karklins, line 119: question: Do you think using onPaste would be more efficient here? Assuming that user can not enter these invalid characters by keyboard
```
            onPaste: (e: React.ClipboardEvent<HTMLInputElement>) => {
              e.preventDefault();
              const raw = e.clipboardData.getData('text');
              const cleaned = raw.replace(/[^a-z0-9 ]/gi, '');
              setLocalTitle(cleaned);
            },
```

PR #1067, opened 2025-05-19
fix: backfill the submitted date field in the proposals table

ACLay:
## Description

Adds a database patch to fill in missing values from the submitted_date column in the proposals table using values from the event log.

## Motivation and Context

It ensures that our data is consistent, simplifying any queries we perform so that they don't have to be aware of historic ways to store/query submission dates.

## How Has This Been Tested

I've run the patch against my local install, using a combination of manually inserted dummy events, and complete copies of our production database.

## Fixes

Closes https://github.com/UserOfficeProject/issue-tracker/issues/1353

### DIFF
--- a/apps/backend/db_patches/0181_BackfillSubmittedDate.sql
+++ b/apps/backend/db_patches/0181_BackfillSubmittedDate.sql
@@ -1,0 +1,22 @@
+DO
+$$
+BEGIN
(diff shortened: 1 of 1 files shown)


mutambaraf (approved):

jekabs-karklins (approved):
👍🏻

PR #1074, opened 2025-05-23
feat: add proposal download status action

simonfernandes:
## Description
This adds a new "proposal download" status action. When the status action is executed, it will download the PDF from the Factory (generated as normal) and store it as a Postgres large object with a reference to the file in the Files table. A later PR will add an option when downloading a PDF to check Postgres storage and prefer the stored version, instead of generating it via the Factory.

The files are stored in `primaryKey.pdf` format, and there a new foreign key has been added to the proposals table for its `file_id`. When `files.file_id` is deleted, the `proposals.file_id` is set to null, and when `files.file_id` is updated, the value refelects in `proposals.file_id`.

A new page has been added to the status action logs to show successful/failed downloads with an option to replay a single status action or all failed status actions in a call.

Where the status action payload contains multiple proposals, the proposals are processed one-by-one to avoid heavy load.

image
image

## Motivation and Context
At STFC, we need a snapshot of the proposal users' details at the time the proposal was submitted or when the call ends. Generating the PDF at a later date can often mean the PDF is innacurate where user data has changed (see https://github.com/UserOfficeProject/issue-tracker/issues/1161). We currently store our legacy PDFs in Postgres, but they have been imported via a script and we have a need to automate this process.

## Changes
- Added a new status action for proposal download in the SQL database. This new action is configured to download PDFs from the factory and store them in the database.
- Modified the GenericTemplateQuestionaryAuthorizer and GenericTemplateDataSource to remove the agent parameter from getGenericTemplates function. This simplifies the function and removes unnecessary complexity.
- Updated the FileDataSource interface and its implementations to include an optional internalUse parameter. This parameter is used to flag files for internal use only, providing a way to distinguish between files that should be publicly available and those that should not.
- Made corresponding changes to the mock implementations of these data sources to reflect the changes in the interfaces.
- Updated the PostgresFileDataSource to handle the new internalUse parameter and to unlink the source file after storing its blob in the database. This is done to free up storage space.

## How Has This Been Tested?

This PR is just for storing the PDFs in Postgres and there isn't yet a mechanism to download them through the interface. So currently there is limited testing that can be done until that feature is added and it wouldn't be recommended to use the status action yet.

- e2e for status action logs
- tested that correct proposal metadata is being stored in the Files table
- spot checked some proposal PDFs to ensure that the correct PDF is stored as large objects
- tested replaying the logs
- tested single and multiple proposal PDF download (800 proposals)

## Fixes Jira Issue

Part of https://github.com/UserOfficeProject/issue-tracker/issues/1161

## Depends On

https://github.com/UserOfficeProject/user-office-lib/pull/214

### DIFF
--- a/apps/frontend/src/components/PageLayout.tsx
+++ b/apps/frontend/src/components/PageLayout.tsx
@@ -71,14 +71,12 @@ const PageLayout = ({
   };
   const drawerOpen = () => ({
     width: drawerWidth,
-    overflowX: 'hidden',
     transition: theme.transitions.create('width', {
       easing: theme.transitions.easing.sharp,
       duration: theme.transitions.duration.enteringScreen,
     }),
   });
   const drawerClose = () => ({
-    overflowX: 'hidden',
     transition: theme.transitions.create('width', {
       easing: theme.transitions.easing.sharp,
       duration: theme.transitions.duration.leavingScreen,
@@ -164,7 +162,7 @@ const PageLayout = ({
             </IconButton>
           </Box>
           <Divider />
-          <List disablePadding>
+          <List disablePadding sx={{ overflowX: 'hidden' }}>
             <MenuItems currentRole={currentRole} />
           </List>
           <Divider />
--- a/apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
+++ b/apps/frontend/src/components/fap/Proposals/FapProposalsAndAssignmentsTable.tsx
@@ -603,13 +603,12 @@ const FapProposalsAndAssignmentsTable = ({
       onClick: handleAssignMembersToFapProposals,
       position: 'toolbarOnSelect',
     });
-  hasRightToAssignReviewers &&
-    tableActions.push({
-      icon: () => <GetAppIcon data-cy="download-fap-proposals" />,
-      tooltip: 'Download proposals',
-      onClick: handleBulkDownloadClick,
-      position: 'toolbarOnSelect',
-    });
+  tableActions.push({
+    icon: () => <GetAppIcon data-cy="download-fap-proposals" />,
+    tooltip: 'Download proposals',
+    onClick: handleBulkDownloadClick,
+    position: 'toolbarOnSelect',
+  });
   hasRightToRemoveAssignedProposal &&
     tableActions.push({
       icon: () => <DeleteOutline data-cy="remove-assigned-fap-proposal" />,


Scott-James-Hurley (approved):

apps/backend/src/statusActionEngine/proposalDownloadActionHandler.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  const successMessage = !!options?.statusActionsLogId
+    ? 'Proposal successfully downloaded on status action replay'
+    : 'Proposal successfully downloaded';
+  const failMessage = !!options?.statusActionsLogId
+    ? 'Proposal failed to download on status action replay'
+    : 'Proposal failed to download';
+
+  const submittedProposals = proposals.filter(
+    (prop) => (prop.submitted = true)
```
Scott-James-Hurley, line 45:
```suggestion
    (prop) => (prop.submitted)
```

apps/backend/src/statusActionEngine/proposalDownloadActionHandler.ts
```diff
@@ -0,0 +1,236 @@
...
+    };
+
+    logger.logInfo(
+      `Downloading proposal PDF for proposal ${proposal.proposalId} via status action download`,
+      {
+        ...logContext,
+      }
+    );
+
+    const { collectProposalPDFDataTokenAccess } = await import(
```
Scott-James-Hurley, line 71: Why not hoist this import to the top with the rest?

Scott-James-Hurley (commented):
Thanks for taking on my feedback

Scott-James-Hurley (approved):

apps/backend/src/datasources/postgres/FileDataSource.ts
```diff
@@ -24,54 +24,73 @@ export default class PostgresFileDataSource implements FileDataSource {
...
+    internalUse?: boolean
+  ): Promise<FileMetadata>;
+  public async put(
+    fileName: string,
+    mimeType: string,
+    sizeInBytes: number | undefined,
+    readableStream: NodeJS.ReadableStream,
+    internalUse?: boolean
+  ): Promise<FileMetadata>;
+  public async put(
```
yoganandaness, line 64: Is it possible to just go with the third put method and discard the put 1 and put 2?

apps/backend/src/middlewares/files.ts
```diff
@@ -44,7 +44,9 @@ const files = () => {
     try {
       const fileId = req.params.file_id;
       const path = await baseContext.mutations.file.prepare(fileId);
-      const metaData = await baseContext.queries.file.getFileMetadata(fileId);
+      const metaData = (
+        await baseContext.queries.file.getFileMetadata(fileId)
+      )[0];
```
yoganandaness, line 49: Possible chance of Array index error.

yoganandaness (changes requested):
Looks fine to me. Just a couple of comments

apps/backend/src/datasources/mockups/FileDataSource.ts
```diff
@@ -2,52 +2,76 @@ import { ReadStream } from 'pg-large-object';
...
-  async getMetadata(param: unknown): Promise<FileMetadata | FileMetadata[]> {
-    if (param instanceof FilesMetadataFilter) {
-      const filter = param as FilesMetadataFilter;
-
+  async getMetadata(
+    fileIds?: string[],
+    filenames?: string[],
+    internalUse?: boolean
+  ): Promise<FileMetadata[]> {
+    if (fileIds || filenames) {
```
mutambaraf, line 16: Do we have a dependence on these variables which is outside of this function because the check on filenames is not being used unless l missed it.

apps/backend/src/datasources/mockups/FileDataSource.ts
```diff
@@ -2,52 +2,76 @@ import { ReadStream } from 'pg-large-object';
...
       return Promise.resolve(
-        filter.fileIds?.map(
-          (id) => new FileMetadata(id, 1, 'name', 'text/xml', 1, new Date())
+        fileIds?.map(
```
mutambaraf, line 18: We can remove this check since we have an if just above.

apps/backend/src/datasources/mockups/FileDataSource.ts
```diff
@@ -2,52 +2,76 @@ import { ReadStream } from 'pg-large-object';
...
+            )
         ) ?? []
       );
     }
 
-    if (typeof param === 'string') {
-      const fileId = param as string;
-
-      return new FileMetadata(fileId, 1, 'name', 'text/xml', 100, new Date());
+    if (internalUse != null && internalUse !== undefined) {
```
mutambaraf, line 33: A simple if maybe able to check both states for us here.

apps/backend/src/datasources/postgres/FileDataSource.ts
```diff
@@ -24,54 +24,59 @@ export default class PostgresFileDataSource implements FileDataSource {
...
+    return database('files')
+      .select('*')
+      .modify((queryBuilder) => {
+        if (fileIds && fileIds.length > 0) {
+          queryBuilder.whereIn('file_id', fileIds);
+        }
+        if (filenames && filenames.length > 0) {
+          queryBuilder.whereIn('file_name', filenames);
+        }
+        if (internalUse === true) {
```
mutambaraf, line 41: If internalUse is a boolean a simple if might be able to work here.

apps/backend/src/factory/pdf/genericTemplates.ts
```diff
@@ -81,3 +86,70 @@ export async function collectGenericTemplatePDFData(
 
   return out;
 }
+
+export async function collectGenericTemplatePDFDataTokenAccess(
+  genericTemplateId: number,
+  notify?: CallableFunction,
+  newGenericTemplate?: GenericTemplate,
+  newQuestionary?: Questionary,
+  newQuestionarySteps?: QuestionaryStep[]
```
mutambaraf, line 95: Maybe lets make this into an object

apps/backend/src/datasources/postgres/FileDataSource.ts
```diff
@@ -133,6 +138,84 @@ export default class PostgresFileDataSource implements FileDataSource {
       });
   }
 
+  private async storeBlobFromStream(
+    readableStream: NodeJS.ReadableStream
+  ): Promise<{ oid: number; sizeInBytes: number }> {
```
mutambaraf, line 143: Was there a suggestion to move this part to the factory to allow handling of  multiple proposals download.

mutambaraf (approved):

PR #1078, opened 2025-05-30
fix: issue with default value of abstract field after sanitisation

bashanlam:
## Description
Fix issue with default value of abstract field after sanitisation update.

## Motivation and Context
This change is necessary to address the default value issue with the abstract field introduced after the sanitisation update was merged via https://github.com/UserOfficeProject/user-office-core/pull/1066 .

## Changes
- Fix the issue

Manually tested the change.


bashanlam:
> Maybe we need a test for this change.
I have added tests to cover my changes.

PR #1081, opened 2025-06-02
refactor: move stfc role assignments into proposal backend

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1411

## Description
This PR moves the STFC FAP role assignment into the User Office Project (UOP).

## Motivation and Context
This change is required to centralize the FAP role assignment process into the UOP, making it easier to manage and maintain. It solves the issue of having to manage FAP role assignments on the STFC platform separately.

## Changes
Added a call into the fap reviewer/sec/chair/internalreviewer assignements to stfc external users service to give the user the new role. 
Takes away the role when removed from the panels. 

## Fixes Jira Issue


apps/backend/src/datasources/stfc/StfcFapDataSource.ts
```diff
@@ -59,18 +84,71 @@ export default class StfcFapDataSource
...
+  stfcRoleMap = new Map<UserRole, STFCRolesMap>([
+    [UserRole.FAP_CHAIR, STFCRolesMap.FAP_CHAIR],
+    [UserRole.FAP_SECRETARY, STFCRolesMap.FAP_SECRETARY],
+    [UserRole.FAP_REVIEWER, STFCRolesMap.FAP_REVIEWER],
+  ]);
+
+  async removeMemberFromFap(args: UpdateMemberFapArgs): Promise<Fap> {
+    const role = UserRoleShortCodeMap[args.roleId];
+    const faps = await this.getUserFaps(args.memberId, role);
+
```
Scott-James-Hurley, line 134: Could you add a comment here explaining that if the user's last position in a FAP is removed, then the accompanying role is also revoked?

apps/backend/src/datasources/mockups/FapDataSource.ts
```diff
@@ -173,6 +173,12 @@ export class FapDataSourceMock implements FapDataSource {
   async assignChairOrSecretaryToFap(
     args: AssignChairOrSecretaryToFapInput
   ): Promise<Fap> {
+    if (args.userId === 5 || args.userId === 6) {
```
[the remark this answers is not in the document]
TCMeldrum: This is a mock implemention for some tests where users 5 and 6 don't have the FAP reviewer role.

yoganandaness (dismissed):

apps/backend/src/datasources/mockups/FapDataSource.ts
```diff
@@ -173,6 +173,12 @@ export class FapDataSourceMock implements FapDataSource {
...
+      throw new Error(
+        'Can not assign to Fap, because only users with fap reviewer role can be chair or secretary'
+      );
+    }
```
jekabs-karklins, line 180: suggestion: I have tried running the tests with these lines commented out and the tests are still passing. Are you sure we need these lines?

apps/backend/src/auth/StfcUserAuthorization.ts
```diff
@@ -385,4 +385,8 @@ export class StfcUserAuthorization extends UserAuthorization {
       (await this.userDataSource.isSearchableUser(id))
     );
   }
+
+  async canBeAssignedToFap(userId: number): Promise<boolean> {
```
jekabs-karklins, line 389: Question: I see you always return `true`. Can you explain a little more. From the first glance at the code it looks like we allow anyone (including people with just USER role) to become a FAP reviewer.

TCMeldrum: Yeah thats right for us that isnt really a issue.

For us we get our roles from a external users service and map them onto roles within the UOP. Before this PR we would have to go and add the fap_reviewer (or any fap role) role to the users (about 100 of them) in a external service then add them to the panel within UOP. Now we would like it so that we can cut out that intermeuary step on the external service and just be able to add them here when the user is added to the panel. This means we are adding the roles when they are added to the panel not before so we dont need to check to see if they have the right role yet as we are just going to give it to them. 

We do the role assignment in the STFCFapDataSource so when this check executes we have not given the reviewer role yet so this check just returns true. However we need to do the check as when the softwear is run in non-stfc mode we to check the user has the corret role.

Let me know if you have nay questions.

jekabs-karklins (approved):

PR #1084, opened 2025-06-04
feat: add daterange picker for selecting call start and end date

bolmsten:
## Description
This PR introduces a daterange picker for selecting call start and end date.

Screenshot 2025-06-04 at 15 04 09

## Motivation and Context
The change was required to make the selection of the call start and end date more intuitive and user-friendly.

## Changes
1. Added the "react-day-picker" package to the project, which provides the daterange picker functionality.
2. Updated the CallGeneralInfo component to include the new daterange picker and replaced the previously separate date pickers for start and end date.
3. Adjusted the handling of start and end date in the form context to accommodate the daterange picker.

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4727](https://jira.esss.lu.se/browse/SWAP-4727)


bolmsten:
> Why are we not use the MUI date range picker? https://mui.com/x/react-date-pickers/date-range-picker/
> 
> Since it is a internal use only I think it is fine but I was just wondering

My understanding was that this is using  the pro package and looking at the licensing structure this would not work for us. But if that is not the case I would much rather use that.

jekabs-karklins (approved):
Minor comments but looks good

PR #1086, opened 2025-06-05
fix: add impersonation logs

RasmiaKulan:
## Description
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1406

This PR ensures that the event log correctly reflects actions taken by an impersonating user, rather than attributing those actions to the impersonated user. It also explicitly mentions impersonated actions in the event logs which provides accurate accountability.

image

## Motivation and Context

Previously, the event logs would misleadingly show the impersonated user as the actor for actions, obscuring the identity of the actual user taking those actions. 
## How Has This Been Tested
This PR adds a new Cypress test to verify that the correct user (impersonator) is recorded in the event logs when actions are taken during an impersonation session.

## Changes

1. Introduced an `impersonatingUserId` field in the `EventLogsDataSource` and implemented logic to detect impersonation, ensuring the event log descriptions accurately reflect the impersonating user when applicable.
2. Updated the `EventBus` to include `impersonatingUserId` when creating an `ApplicationEvent`. This ensures that events triggered through the bus correctly capture impersonation context for improved traceability.
3. Updated the `logging` to include impersonatingUserId in all relevent event logging calls.
4. Updated the `EventLog` model to add  `impersonatingUserId`.
5. Added a test in the `eventLogs.cy.ts`.
6. Fixed and added cy tags for improved understanding of test implementation.
7. Updated `UserMutations` with backend logs when `Connect as this user...` is clicked:

image

### DIFF
--- a/apps/backend/src/datasources/postgres/EventLogsDataSource.ts
+++ b/apps/backend/src/datasources/postgres/EventLogsDataSource.ts
@@ -23,15 +23,25 @@ export default class PostgresEventLogsDataSource
     eventType: string,
     rowData: string,
     changedObjectId: string,
-    description?: string
+    description?: string,
+    impersonatingUserId?: number
   ) {
+    const isImpersonating =
+      impersonatingUserId != null && impersonatingUserId !== changedBy;
+
+    const actingUserId = impersonatingUserId ?? changedBy;
+
+    const updatedDescription = isImpersonating
+      ? `${description || ''} (${impersonatingUserId} impersonating ${changedBy})`
+      : description;
+
     return database
       .insert({
-        changed_by: changedBy,
+        changed_by: actingUserId,
         event_type: eventType,
         row_data: rowData,
         changed_object_id: changedObjectId,
-        description: description,
+        description: updatedDescription,
       })
       .returning('*')
       .into('event_logs')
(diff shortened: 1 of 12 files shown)


apps/backend/src/datasources/postgres/EventLogsDataSource.ts
```diff
@@ -23,15 +23,25 @@ export default class PostgresEventLogsDataSource
...
+
     return database
       .insert({
-        changed_by: changedBy,
+        changed_by: actingUserId,
         event_type: eventType,
         row_data: rowData,
         changed_object_id: changedObjectId,
-        description: description,
+        description: updatedDescription,
```
jekabs-karklins, line 44: suggestion: Do you think we can store the `impersonatingUserId` as a column in the `event_logs`? This would be very nice and would allow us to associate the actual user object. We can then query and display this information more flexible. I see you already have the value in the method, it would be just adding a new db_patch with column and setting it.

apps/backend/src/datasources/postgres/EventLogsDataSource.ts
```diff
@@ -23,15 +23,25 @@ export default class PostgresEventLogsDataSource
...
-    description?: string
+    description?: string,
+    impersonatingUserId?: number
   ) {
+    const isImpersonating =
+      impersonatingUserId != null && impersonatingUserId !== changedBy;
+
+    const actingUserId = impersonatingUserId ?? changedBy;
+
+    const updatedDescription = isImpersonating
```
jekabs-karklins, line 34: suggestion: Do you think we could maybe add `userId` into the description, that would make it more readable. so `(userId: 2 impersonating userId:1)`
image

jekabs-karklins (commented):
Hi, I just looked a bit more and noticed 2 things, let me know WDYT.

jekabs-karklins (approved):
Looks good

PR #1087, opened 2025-06-09
fix: fix stfc fap export for LSF

TCMeldrum:


ACLay (commented):
Does it matter for the sheets that we're putting the number of weeks into a variable, presumably which goes into a column header, that's called days?

ACLay (approved):

jekabs-karklins (approved):

PR #1089, opened 2025-06-10
feat: order calls by cycle comment

ellen-wright:
Closes: https://github.com/isisbusapps/proposal-launchpad/issues/160 
## Description

Orders the calls in the select calls list by cycle comment alphabetically. 

This can be used while we do not have facility implementation to order calls, the cycle comment can start with the facility name and they will appear in order. 
## Motivation and Context

As part of the decommissioning of the Proposal launchpad we want the option to order the calls by ISIS/CLF 
## How Has This Been Tested
Tested manually 


bolmsten:
Can we do this with an explicit order instead? Where user officers can move calls up or down in the call table with up and down arrows?

PR #1108, opened 2025-06-26
1333 allow multiple reviewers for one technical review for hpl

Bhaswati1148:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1333
## Description
This PR introduces the ability to mark an instrument for multiple tech reviews. 

## Motivation and Context
This change enables instrument scientists to add and also submit their review comments in addition to those provided by the assigned technical reviewer, when the instrument is marked as allowing multiple technical reviews. Previously, only the designated technical reviewer could submit comments.

The HPL facility in STFC requested this enhancement to ensure that instrument scientists — even when not formally assigned as technical reviewers — can contribute their expertise to proposal evaluations.

## How Has This Been Tested
Added e2e test 

## Changes

1. A new column `multiple_tech_reviews_enabled` has been added in instruments table.
2. `proposal_table_view` has been updated to include the new column `multiple_tech_reviews_enabled` from instruments table
3.  In addition to the existing  where clause, a new where clause has been added in  `getInstrumentScientistProposals` method in  `StfcProposalDataSource.ts`  when `ReviewerFilter.ME` is set to true. The where clause gets executed when the logged in user is not the assigned technical reviewer. Further it also checks if the logged in user is an instrument scientist where the instrument is  marked as enabled for multiple technical reviews, then show those proposals under filter `ReviewerFilter.ME`  for the user.
4. Added a new check box in `CreateUpdateInstrument.tsx` to enable multiple technical reviews for that instrument.
5. GraphQL types, queries, and mutations have been updated to include a new field reflecting the `multiple_tech_reviews_enabled` column in the instrument table.
6.  Added a new e2e test in `instruments.cy.ts`


gnyiri:
At ELI we would definitely use this feature, however, the PR title and the configuration is a bit misleading. From what I understand, this PR would only allow the Instrument Scientist to update the Tech Review even if he/she is not the assigned Tech Reviewer. That is, there will be a single review but more than one user will be able to change it. If so, I would change the option 'Allow mutiple reviews for a technical review for this instrument' to a clearer description.

Scott-James-Hurley (approved):
I was being a numpty, thanks.

jekabs-karklins:
Additionally, as an instrument scientist no assigned to do a review, for proposal where multiple reviewers are allowed  I was not able to comment or review the proposal

The only think I can do is see the proposal.
image
Do you know what might be going on there?

jekabs-karklins:
> @jekabs-karklins
> 
> 1. We have asked '_would it make sense to have this as an only and default behavior, i.e. allow all instrument scientists to contribute their opinion_' to our User-Officer Emma Gozzard and her reply is -
>    '_No, I think it would be good to be flexible here and have the on/off toggle.  I can see that some lasers/instruments may want this feature, but some will definitely not_'
> 2. As of now both instrument scientists and the primary reviewer can submit reviews; there are no restrictions in place However, I will update the logic to restrict submission to the primary reviewer only.
> 3. I would update the logic in https://github.com/UserOfficeProject/user-office-core/blob/729b602fcb4ad315624bc4a5b4e187ec23b95ca1/apps/backend/src/datasources/postgres/ProposalDataSource.ts#L589
>     so that it reflects properly on ESS side and hopefully solve the issue [here](https://github.com/UserOfficeProject/user-office-core/pull/1108#issuecomment-3044503236)

@Bhaswati1148  Than you for followinf up on it. This is good information and if you see value in the feature toggle, it will probably be relevant for us too.

When you say "However, I will update the logic to restrict submission to the primary reviewer only." do you plan to do it in this PR or will create a new one?

Thanks

jekabs-karklins:
@Bhaswati1148 Thank you. That sounds perfect 👍🏻

jekabs-karklins:
@Bhaswati1148 I am reviewing again could you help me understand how this feature differ from Internal Reviewers we already have in UOS, added here 
https://github.com/UserOfficeProject/user-office-core/pull/258

Aren't both features trying to solve the same problem?

TCMeldrum:
I think the only differance who the tech reviewer/internal reviewer is. To edit the tech reivew you must be on the instrument the proposal is assigned to but a internal reviewer does not have to be a instrument sci and on the proposal. At STFC we it has been suggested the internal reviews come from people like the duteration lab or from another instrument.

PR #1118, opened 2025-07-04
feat: Add read permissions to questions 

TCMeldrum:


jekabs-karklins:
Hi @TCMeldrum I will try to look at this today

jekabs-karklins:
question: I think there is something that does not work intuitively
So say I create a Boolean question with read permission set to FAP_CHAIR and leave USER role unchecked

image
As a user I can still read the question

apps/backend/src/datasources/QuestionaryDataSource.ts
```diff
@@ -14,9 +14,18 @@ export interface QuestionaryDataSource {
   ): Promise<AnswerBasic | null>;
   delete(questionary_id: number): Promise<Questionary>;
   getQuestionary(questionary_id: number): Promise<Questionary | null>;
-  getQuestionarySteps(questionaryId: number): Promise<QuestionaryStep[]>;
-  getBlankQuestionarySteps(templateId: number): Promise<QuestionaryStep[]>;
-  getBlankQuestionaryStepsByCallId(callId: number): Promise<QuestionaryStep[]>;
+  getQuestionarySteps(
+    questionaryId: number,
+    currentUserRole: string | null
```
jekabs-karklins, line 19: question: this is more a question if maybe a better approach is do the logic of filtering in the authorization layer, instead of sending it down to the database layer.

pros: 
- keeping db function signature simple and function body DRY 
- keeping authorization related code in the authorization layer
cons:
- slower to execute

TCMeldrum:
> question: I think there is something that does not work intuitively So say I create a Boolean question with read permission set to FAP_CHAIR and leave USER role unchecked
> image As a user I can still read the question

@jekabs-karklins The user role is a interesting case, like if a user is retriving a questionary is there any time it would not be a questionary that they own or are a part of (like a co I)? So would read permissions ever be needed for the user role? 

Also I think I might need to refactor slightly so people are able to fully see questionarys they have themselves filled in.

TCMeldrum:
I've been looking at some edge cases and I'm not quite sure what too do. For instance lets say a person is looking at questionary they filled in, should the read permissions still apply? The data base structure makes it very hard to get from a questionary to the person who filled in the questionary (get template of questionary -> get template catorgory -> template group -> base table of that group e.g. proposals or fap reviews). There are also other situations like Co-Is or the new Data Access users. 

If anyone has any ideas on the best way to go about it then that would be helpfull!

jekabs-karklins:
> I've been looking at some edge cases and I'm not quite sure what too do. For instance lets say a person is looking at questionary they filled in, should the read permissions still apply? The data base structure makes it very hard to get from a questionary to the person who filled in the questionary (get template of questionary -> get template catorgory -> template group -> base table of that group e.g. proposals or fap reviews). There are also other situations like Co-Is or the new Data Access users.
> 
> If anyone has any ideas on the best way to go about it then that would be helpfull!

Hi @TCMeldrum,

I think we have a bit of conundrum with Robe Based Access Control vs Object-Level Authorization here.
We might want all users with INSTRUMENT_SCIENTIST role to be able to read the question, at the same time we want only a subset of users with USER_ROLE to be able to read..

We already have similar problem solved with endpoint authorization. While decorators specify roles who can access the endpoint, Authorizers do object level permission checks.

If we want to follow the same patter I'd suggest the following design assumption - 
if no readPermission role has been set - every role can read.
if role(s) has been set only the specified roles can read the question

So let's say if you want question to be readable by INSTRUMENT_SCIENTIST you would set USER,USER_OFFICER,INSTRUMENT_SCIENTIST.

Luckily you do not have to deal with object level authorization here for USER, because that is handled already in the object above (proposal or questionnaire) . 

I'd try to look at the analogy we have right now with the endpoint auth, and by the way, what happens when the request with the API_ACCESS_TOKEN comes in, how are the permissions handled with question level reads there?

TCMeldrum:
Cool thanks! I think I have implemented as you discribe with role level permissions not object. For API keys I thought they should just have access no matter what the readPermissions are if that makes sense. So in the check I check if it a api key or the user is null and therefore comes from a internal api token function in apps/backend/src/queries/QuestionaryQueries.ts

jekabs-karklins (approved):

PR #1120, opened 2025-07-04
fix: add access route to stfc excel export

TCMeldrum:


apps/backend/src/factory/xlsx/stfc/StfcFapDataRow.ts
```diff
@@ -55,6 +55,14 @@ export async function getStfcDataRow(
         answer.question.naturalKey === 'Weeks_Requested'
     )?.value.value;
 
+  const accessRoute = proposalAnswers
+    ?.flatMap((step) => step.fields)
+    .find(
+      (answer) =>
+        answer.question.naturalKey === 'Proposed_Route' ||
```
Scott-James-Hurley, line 62: Is there a way to make it more obvious these values need to be updated if the keys change? Maybe extracting them to constants or adding some comments?

TCMeldrum: I'll add some comments in code but I dont think that we will be able to add anything in the fronend ui unforunatly

jekabs-karklins (approved):

Scott-James-Hurley (approved):

PR #1126, opened 2025-07-14
feat: remove gender, telephone, username, position, placeholder, birthdate …

bolmsten:


apps/backend/db_patches/0183_RemoveUserColulmns2.sql
```diff
@@ -0,0 +1,24 @@
+-- 0181_RemoveUserColumns.sql
+DO
+$$
+BEGIN
+    IF register_patch(
+        '0182_RemoveUserColumns2.sql',
```
jekabs-karklins, line 6: suggestion: No need for the numbers, they might go easily out of sync with the file name
```suggestion
        'RemoveUserColumns2.sql',
```

apps/backend/src/mutations/UserMutations.ts
```diff
@@ -99,7 +99,7 @@ export default class UserMutations {
     }
     // Check if email exist in database and if user has been invited before
     const user = await this.dataSource.getByEmail(args.email);
-    if (user && user.placeholder) {
+    if (user) {
```
jekabs-karklins, line 102: issue: I think this is fine, only now the`else if (user)`clause is will never be called, we should probably clean it up

apps/frontend/src/components/user/UpdateUserInformation.tsx
```diff
@@ -310,15 +179,7 @@ export default function UpdateUserInformation(
...
-              <Field
-                name="telephone"
-                label="Telephone"
-                id="telephone-input"
-                component={TextField}
-                type="text"
-                data-cy="telephone"
-                required
-              />
+              {/* Remove telephone field */}
```
jekabs-karklins, line 182:
```suggestion
```
suggestion: Some leftover comment, we can clean it up

gnyiri:
How about removing preferred name as well?

bolmsten:
> How about removing preferred name as well?

I will raise it again with our user officer, the last time she identified this as a field she wants. Can STFC comment here on this?

simonfernandes (commented):
@bolmsten We have moved over to the new particapant selector modal now, so we are all good on removing the `EMAIL_INVITE_LEGACY` feature 🙂

simonfernandes:
Hi @bolmsten, I've had a look at the tests. I can see in this PR that the only failing file is proposals.cy.ts, and locally in STFC config there is only one failing test within it. It was failing because it was still looking for the old PI selector instead of the new one. I've updated it and it's passing locally now

PR #1133, opened 2025-07-21
fix: escape special characters in email and filter queries

jekabs-karklins:
## Description
This PR introduces a fix to escape special characters in email and filter queries.
Issue raised here
https://github.com/UserOfficeProject/issue-tracker/issues/1441

## Motivation and Context
This change is required to address the issue where special characters in email and filter queries were causing unexpected search results and potential security vulnerabilities.

## Changes
1. Introduced a new `escapeLike` function to escape special characters in a given string.
2. Used `whereILikeEscaped`, and `orWhereILikeEscaped` function to sanitize email input in `PostgresUserDataSource`.
3. Applied the new function in the affected functions

This fix enhances the robustness of our search functionality and ensures that special characters in email and filter queries are handled properly.

## Fixes Jira Issue

https://github.com/UserOfficeProject/issue-tracker/issues/1441

### DIFF
--- a/apps/backend/src/datasources/mockups/UserDataSource.ts
+++ b/apps/backend/src/datasources/mockups/UserDataSource.ts
@@ -399,7 +399,7 @@ export class UserDataSourceMock implements UserDataSource {
     return dummyUser;
   }
 
-  async getUser(id: number) {
+  async getUser(id: number): Promise<User | null> {
     if (id === dummyPlaceHolderUser.id) {
       return dummyPlaceHolderUser;
     }
--- a/apps/backend/src/eventHandlers/email/eliEmailHandler.spec.ts
+++ b/apps/backend/src/eventHandlers/email/eliEmailHandler.spec.ts
@@ -7,9 +7,7 @@ import { CoProposerClaimDataSourceMock } from '../../datasources/mockups/CoPropo
 import { RoleClaimDataSourceMock } from '../../datasources/mockups/RoleClaimDataSource';
 import { ApplicationEvent } from '../../events/applicationEvents';
 import { Event } from '../../events/event.enum';
-import { UserRole } from '../../models/User';
 import { eliEmailHandler } from './eliEmailHandler';
-import { getTemplateIdForInvite } from './essEmailHandler';
 
 // Mock MailService
 const mockMailService = {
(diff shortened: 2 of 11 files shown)


apps/backend/src/datasources/postgres/UserDataSource.ts
```diff
@@ -488,11 +493,12 @@ export default class PostgresUserDataSource implements UserDataSource {
...
+          const escapedFilter = escapeLike(filter.trim());
           query.andWhere((qb) => {
-            qb.where('institution', 'ilike', `%${filter}%`)
-              .orWhere('firstname', 'ilike', `%${filter}%`)
-              .orWhere('preferredname', 'ilike', `%${filter}%`)
-              .orWhere('lastname', 'ilike', `%${filter}%`);
+            qb.where('institution', 'ilike', `%${escapedFilter}%`)
+              .orWhere('firstname', 'ilike', `%${escapedFilter}%`)
+              .orWhere('preferredname', 'ilike', `%${escapedFilter}%`)
+              .orWhere('lastname', 'ilike', `%${escapedFilter}%`);
```
yoganandaness, line 501: This is great. There are few other instances, where we do `ilike`. Should we change them as well. 

Also, since this needs to be updated in several places, it is a good idea to consider writing utils. 

If we can write this on the Knex prototype, we can use it in many places.I just chatGPTed and got this. 

// Add custom method
knex.QueryBuilder.extend('whereLikeEscaped', function (column: string, userInput: string) {
  const escapedInput = escapeLike(userInput);
  return this.whereRaw(
    `${column} LIKE ? ESCAPE '\\'`,
    [`%${escapedInput}%`]
  );
});

// Usage:
knex('my_table')
  .whereLikeEscaped('my_column', userInput)
  .then(...);


Let me know your thoughts.

yoganandaness (changes requested):

apps/backend/src/datasources/postgres/UserDataSource.ts
jekabs-karklins: This is a good idea, although it expands the scope a bit, now we have more generic and easier to read solution

jekabs-karklins:
@yoganandaness could you please look at this when you get some time?

yoganandaness (approved):
Awesome @jekabs-karklins

Scott-James-Hurley:
Can you give an example of using `%` to search a user's email?

PR #1134, opened 2025-07-22
feat: download pregenerated PDFs

simonfernandes:
## Description
This PR follows up on https://github.com/UserOfficeProject/issue-tracker/issues/1168 (storing PDFs via status actions) to add a way to download them. When the `PREGENERATED_PROPOSAL_PDF` feature is enabled and a PDF is downloaded, it will download the pregenerated PDF for each proposal where possible, and will generate one as normal where one doesn't exist. It supports single PDF, zip and merged PDF downloads. As there is now a reference to a proposal's file in the proposals table, the STFC download service has been removed along with dependency injection for download services.

The `PREGENERATED_PROPOSAL_PDF` feature also hides the ability to use the Proposal Download status action, and view the associated status action logs.

There is also a change to the GitHub `test-build` workflow as part of this. When creating this PR I found that the GitHub Actions e2e tests were running against the develop Factory image, instead of the associated Factory image for this feature. It was due to the `test-build` workflow not handling pagination in the Factory tags request. This change solves the problem by making use of the GitHub API to make the request instead of cURL.

## Motivation and Context
The current system generates a PDF for every proposal download request, which is time-consuming and can slow down the server performance. By storing pre-generated PDFs and offering them for download, we can enhance the speed and performance of the system and improve user experience.

## Changes
- Introduced a new feature switch `PREFER_PREGENERATED_PROPOSAL_DOWNLOAD` to enable or disable the pre-generated PDF download functionality.
- Modified the `ProposalPDFData` type to include `FullProposalPDFData` and `PregeneratedProposalPDFData` to handle both full and pre-generated proposal data.
- Added `collectProposalPregeneratedPdfData` and `collectProposalPregeneratedPdfDataTokenAccess` functions to fetch the pre-generated PDF data if available.
- If pre-generated PDF is not available, the system will fall back to the existing functionality of generating a PDF.

## How Has This Been Tested?

- Added e2e testing with PDF content checking
- Manual testing of different download types

## Fixes Jira Issue

Refs https://github.com/UserOfficeProject/issue-tracker/issues/1396

## Depends On

Factory PR: https://github.com/UserOfficeProject/user-office-factory/pull/656

Blocked from merging until https://github.com/UserOfficeProject/issue-tracker/issues/1443 is complete

### DIFF
--- a/apps/backend/package-lock.json
+++ b/apps/backend/package-lock.json
@@ -4215,6 +4215,21 @@
       "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
       "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k="
     },
+    "node_modules/available-typed-arrays": {
+      "version": "1.0.7",
+      "resolved": "https://registry.npmjs.org/available-typed-arrays/-/available-typed-arrays-1.0.7.tgz",
+      "integrity": "sha512-wvUjBtSGN7+7SjNpq/9M2Tg350UZD3q62IFZLbRAR1bSMlCo1ZaeW+BJ+D090e4hIIZLBcTDWe4Mh4jvUDajzQ==",
+      "license": "MIT",
+      "dependencies": {
+        "possible-typed-array-names": "^1.0.0"
+      },
+      "engines": {
+        "node": ">= 0.4"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/ljharb"
+      }
+    },
     "node_modules/await-to-js": {
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/await-to-js/-/await-to-js-2.1.1.tgz",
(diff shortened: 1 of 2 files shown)


yoganandaness:
@simonfernandes Requesting to resolve the merge conflicts

deepaksftc (approved):

.github/workflows/test-build.yml
yoganandaness, line 1: @simonfernandes  Thanks for the explanation on this. Now, i understand this much more better than before. 

I see a major drawbacks to this. Correct me if i am wrong

I understand that, the e2e test runs agains the factory develop branch. You have changed this to the matching branch of that of this repository. The potential problem could be that, when factory PR is still in review and the User office PR gets approved, this PR gets to be merged. However, later, when the factory PR undergoes changes, then it would be difficult to change the merged PR in User office core and need to create a new PR based on the change in the factory PR. 

This is why, it is important for the Factory PR to be reviewed and merged into develop first, before merging the User office core. This way, the User office e2e tests always pull from factory develop code, which contains approved code. 

Let me know in case of any concerns.

yoganandaness (changes requested):

yoganandaness (approved):

PR #1138, opened 2025-07-23
fix: masking sensitive data from traces and logs

deepaksftc:
## Description
This PR introduces a mechanism to mask sensitive data in our trace logs.

## Motivation and Context
The need for this change arises from the necessity to ensure the security of sensitive data, such as tokens, which were previously being exposed in our trace logs.

## Changes
1. A new `AttributeFilterProcessor` class has been added which implements the `SpanProcessor` interface from the OpenTelemetry SDK. This class masks specified attributes in the span of a trace.
2. This new processor is added to the `spanProcessors` array in the `NodeSDK` configuration, applying it to all spans.
3. A `masklist` is defined which currently includes 'graphql.variables.token' and 'graphql.variables.externalToken'. Any attributes in this list will be masked by the `AttributeFilterProcessor`.

## How Has This Been Tested?

Manually through Grafana


ACLay:
Just to check, is this purely about traces, or also about logs? They're two different data types in opentelemetry, but you mention them together as "trace logs" a couple of times. 

I don't think these changes would cover logs, where we also see tokens getting recorded as part of failed calls to the external users service ([example](https://logs.devs.facilities.rl.ac.uk/app/discover#/doc/7e1733e0-d283-11ea-856e-3dc8c807c0b0/kubernetes-bisappslogs-7.13.2-2025.07.25?id=YV7wQJgBPVF-QYYgRNfQ)).

deepaksftc:
@ACLay, I have added logic for masking the token in logs. Could you please have a look.

deepaksftc:
> Changes look okay, but as a side note it seems like your branch is getting recreated and force pushed each time you make changes, and that makes it harder to see what's changing between reviews in the github UI as it's different commits every time.

It should be because I rebased with develop and force pushed. Will keep in mind next time.

PR #1153, opened 2025-08-11
feat: more powerful workflows

jekabs-karklins:
## Description
This PR introduces more powerful workflow UI editor based on React Flow.

image

## Motivation and Context
The older workflow UI was based on Beautiful DND and while a good library, it wasnt quiet suitable for the graph like structure workflows are. ReactFlow enables treat and edit workflows easier.

This is first part of the Workflow Overhaul which keeps backend pretty much intact, at least when it comes to core functioning principles. This means that the PR does not introduce "multi-parent structure", cycles or conditions which will follow latter.

This PR is mainly moving the UI to ReactFlow and keeping the existing functionality unchanged

## Changes
1. In the backend, new columns `pos_x`, `pos_y`, and `prev_connection_id` have been added to `workflow_connections` table to capture the position and relationship of workflow connections.
2. Unused columns `droppable_group_id` and `parent_droppable_group_id` have been made nullable and will be dropped later as they are not used anymore.
3. Data sources `StatusDataSource` and `WorkflowDataSource` have been updated to reflect the changes in the database schema and to handle the new workflow functionality.
4. ReactFlow library in the UI has been added

## How Has This Been Tested?

Manually and updated E2E tests

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4869](https://jira.esss.lu.se/browse/SWAP-4869)

## Depends On

### DIFF
--- a/apps/e2e/cypress/e2e/settings.cy.ts
+++ b/apps/e2e/cypress/e2e/settings.cy.ts
@@ -1808,13 +1808,13 @@ context('Settings tests', () => {
       cy.get(`[aria-label="Edge from ESF_IS_REVIEW to ESF_REJECTED"]`).should(
         'exist'
       );
-    });
 
-    cy.connectReactFlowNodes(awaitingEsf, esfIsReview);
-    cy.finishedLoading();
-    cy.get(`[aria-label="Edge from AWAITING_ESF to ESF_IS_REVIEW"]`).should(
-      'exist'
-    );
+      cy.connectReactFlowNodes(awaitingEsf, esfIsReview);
+      cy.finishedLoading();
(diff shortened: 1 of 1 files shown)


Scott-James-Hurley (changes requested):
This looks tremendous, well done. When testing this I noticed:

- scrolling to the bottom or top of the status list can cut part of it off
- the draft status isn't in the status list and isn't the default starting status so I don't know how to access it
- the statuses in existing workflows all have the same coordinates, so they all overlap - it's probably a bigger headache to solve this programatically than to just move them manually 

Some additional thoughts I had whilst messing around with workflows:
- should it be possible to change an edge's parent/target
- should saving be manual in case the user mucks up a workflow?

Scott-James-Hurley (approved):
Apologies, I wasn't aware this was an existing issue. Thank you for resolving it.

PR #1154, opened 2025-08-12
fix: assign instruments on proposal clone

TCMeldrum:
Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1462

## Description
This PR introduces a fix to assign the same instruments to a cloned proposal as the source proposal.

## Motivation and Context
This change is required to ensure that cloned proposals maintain the same instrument assignments as their source proposal. This solves the problem where cloned proposals were not correctly inheriting the instrument assignments from their source proposals.

## Changes
1. Added InstrumentMutations to ProposalMutations.
2. Implemented functionality to fetch the instrument ID from the source proposal and assign it to the cloned proposal.
3. Added error handling to log a warning if the instrument assignment fails for the cloned proposal. 

These changes ensure that the instrument assignments are correctly copied over when a proposal is cloned, maintaining consistency and accuracy in our data.

## Fixes Jira Issue


deepaksftc (approved):

yoganandaness:
Also, another question that is popping on my mind is that

When we clone a Proposal, we get to select a Call. Here there is a Potential chance of selecting a Call, which doesn't have the Instrument that the Original Proposal has. Do we get to handle this? Let me know your thoughts.

yoganandaness:
> > Also, another question that is popping on my mind is that
> > When we clone a Proposal, we get to select a Call. Here there is a Potential chance of selecting a Call, which doesn't have the Instrument that the Original Proposal has. Do we get to handle this? Let me know your thoughts.
> > @TCMeldrum
> 
> In that senario the assign instrument method in instrument mutations will fail and no instrument will be assigned. The instrument picker answer will also remove any instruments not on the call.

Great point. I can see that in the code. Thanks a lot @TCMeldrum

yoganandaness (changes requested):

yoganandaness (approved):

PR #1155, opened 2025-08-13
hotfix: api token download for pdfs

mutambaraf:
## Description
This fix the issue were api tokens can not down load proposal pdfs 

## Motivation and Context

Issue with pdf not able to de downloaded using api token
`{"name":"Error","message":"Questionary steps for Questionary ID '39668' not found, or the user has insufficient rights","stack":"Error: Questionary steps for Questionary ID '39668' not found, or the user has insufficient rights\n    at /home/node/app/build/src/factory/pdf/proposal.js:42:15\n `

### DIFF
--- a/apps/backend/src/factory/pdf/proposal.ts
+++ b/apps/backend/src/factory/pdf/proposal.ts
@@ -79,18 +79,13 @@ const getTechnicalReviewHumanReadableStatus = (
 };
 
 const getSampleQuestionarySteps = async (
-  questionaryId: number,
-  user: UserWithRole | null
+  questionaryId: number
 ): Promise<QuestionaryStep[]> => {
   const questionaryDataSource = container.resolve<QuestionaryDataSource>(
     Tokens.QuestionaryDataSource
   );
-  const questionarySteps = user
-    ? await baseContext.queries.questionary.getQuestionarySteps(
-        user,
-        questionaryId
-      )
-    : await questionaryDataSource.getQuestionarySteps(questionaryId);
+  const questionarySteps =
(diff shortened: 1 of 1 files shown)


deepaksftc (approved):

apps/backend/src/factory/pdf/proposal.ts
```diff
@@ -82,11 +82,15 @@ const getSampleQuestionarySteps = async (
...
-    );
+  const questionaryDataSource = container.resolve<QuestionaryDataSource>(
+    Tokens.QuestionaryDataSource
+  );
+  const questionarySteps = user
+    ? await baseContext.queries.questionary.getQuestionarySteps(
+        user,
+        questionaryId
+      )
+    : await questionaryDataSource.getQuestionarySteps(questionaryId);
```
yoganandaness, line 93: `baseContext.queries.questionary.getQuestionarySteps` has the code logic to handle the authorization stuff. 

https://github.com/UserOfficeProject/user-office-core/blob/a37d254d92892c93d6f317860acb9164bee89ff0/apps/backend/src/queries/QuestionaryQueries.ts#L88-L95

Should do the appropriate fix in the Query? Let me know your thoughts.

yoganandaness (changes requested):

apps/backend/src/factory/pdf/proposal.ts
mutambaraf: We are not passing agent so it will fail here `this.userAuth.isApiToken(agent)` and there are places where we do not have the agent at all .Like `await collectProposalPDFDataTokenAccess(
        proposal.primaryKey
      );` in status actions handler.

PR #1161, opened 2025-08-19
refactor: adapt the InviteUser co-proposer dialog to work with other features

simonfernandes:
## Description

This changes the InviteUser component to respect the `EMAIL_SEARCH` and `EMAIL_INVITE` features. Doing so means the component can be used with invites disabled and by restricting user lookups to only knowing someone's full email address. This  component is only used when adding co-proposers to a proposal currently.

## Motivation and Context

STFC looked into whether the ESS could go ahead with removing the `EMAIL_INVITE_LEGACY` feature and we found that doing so would result in the InviteUser component being used instead of the PeopleTable component. We liked the look of it and wanted to use the new component without invites and with only the email search enabled.

## How Has This Been Tested

Added e2e tests and manual testing.

## Fixes

Refs https://github.com/UserOfficeProject/issue-tracker/issues/1446

## Changes

- Display different messages/placeholders when the EMAIL_SEARCH feature is enabled.
- Disable being able to invite people when EMAIL_INVITE is disabled.
- Disable partial user lookups when EMAIL_SEARCH is enabled

### DIFF
--- a/apps/e2e/cypress/e2e/invites.cy.ts
+++ b/apps/e2e/cypress/e2e/invites.cy.ts
@@ -185,6 +185,32 @@ context('Invites tests', () => {
         .should('not.exist');
     });
 
+    it.only('Should not be able to invite the email of the current user', function () {
+      if (!featureFlags.getEnabledFeatures().get(FeatureId.EMAIL_INVITE)) {
+        this.skip();
+      }
+
+      const currentUserEmail = initialDBData.users.user1.email;
+
+      cy.get('[data-cy=add-participant-button]').click();
+
+      cy.get('[data-cy="invite-user-autocomplete"]').type(
+        `${currentUserEmail}`
+      );
+      cy.contains('Loading...').should('not.exist');
+      cy.contains(`${currentUserEmail} has already been invited`).should(
+        'exist'
+      );
+
+      cy.get('[data-cy="invite-user-autocomplete"]').type('{enter}');
+      cy.contains(`${currentUserEmail} has already been invited`).should(
+        'exist'
+      );
+      cy.get('[data-cy="invite-user-autocomplete"]')
+        .find('.MuiChip-label')
+        .should('not.exist');
+    });
+
     it('Should not be able to delete added users with backspace', function () {
       const name = initialDBData.users.user2.lastName;
       const email = initialDBData.users.user2.email;
--- a/apps/frontend/src/components/common/UserManagementTable.tsx
+++ b/apps/frontend/src/components/common/UserManagementTable.tsx
@@ -12,6 +12,7 @@ import React, { useContext, useState } from 'react';
 import { ActionButtonContainer } from 'components/common/ActionButtonContainer';
 import PeopleTable from 'components/user/PeopleTable';
 import { FeatureContext } from 'context/FeatureContextProvider';
+import { UserContext } from 'context/UserContextProvider';
 import { BasicUserDetails, FeatureId, Invite, UserRole } from 'generated/sdk';
 
 import InviteUser from '../proposal/InviteUser';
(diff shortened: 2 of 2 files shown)


Scott-James-Hurley (approved):

Scott-James-Hurley (approved):

jekabs-karklins:
> @jekabs-karklins I noticed a potential issue:
> 
> When adding a user by email, we are only checking the emails of previously invited users (via the `excludeEmails` list). This means that you can still invite a user who is already added as a co-proposer or PI. This isn't a problem for adding non-invite users. I think the fix is for us to additionally add the emails of the all of the users already on the proposal [when creating the excludeEmails array](https://github.com/UserOfficeProject/user-office-core/blob/dd58481449ea94cc1180f93e331ac15401ed0a9b/apps/frontend/src/components/common/UserManagementTable.tsx#L93). I've tried to do this but I'm struggling to figure out why the email property in each BasicUserDetails user object is coming back null.
> 
> If this is a bug and not by design then let me know and I will take another look via this PR, and reach out if I get really stuck. For now I've temporarily excluded the duplication tests from running with invites enabled via ([52f6c9e](https://github.com/UserOfficeProject/user-office-core/pull/1161/commits/52f6c9e6583e976162d85279012d7d10b0aa9851)).

Hi @simonfernandes to me it looks like you have fixed it, correct?

jekabs-karklins (approved):

jekabs-karklins:
The email field is guarded by role-based authorization. Unless the requester has one of these roles — USER_OFFICER, INSTRUMENT_SCIENTIST, or INTERNAL_REVIEWER — the GraphQL API will return null for privacy:

[https://github.com/UserOfficeProject/user-office-core/blob/532117ac10dee66782ba4dc0a507d7c130d308f1/apps/backend/src/resolvers/types/BasicUserDetails.ts#L31](https://github.com/UserOfficeProject/user-office-core/blob/532117ac10dee66782ba4dc0a507d7c130d308f1/apps/backend/src/resolvers/types/BasicUserDetails.ts?utm_source=chatgpt.com#L31)
```
@Authorized([Roles.USER_OFFICER, Roles.INSTRUMENT_SCIENTIST, Roles.INTERNAL_REVIEWER])
@Field({ nullable: true })
public email: string;
```

Edge cases to align on
- If a user is invited as a co-proposer but is already a co-proposer, accepting the invite will be a no-op.
- I haven’t checked whether we already prevent someone from being both PI and co-proposer; if not, we should add a guard/constraint.

jekabs-karklins:
> @jekabs-karklins I noticed a handful of issues around using the enter key and not waiting for results to finish loading which are fixed now.
> 
> The invite-specific issues mentioned above are still present but I understand that you're fine for this to be merged without them as it's either handled by invite logic or a bigger issue due to the privacy level on the email field resolver.
> 
> Do you want me to create an issue for the outstanding issues? Or if we are fine to lower the privacy level I'm happy to do it as part of this PR?

I tested and it works.
I am fine with that you can send an invite to earlier added co-proposer. because I think users would not invite co-proposers that are already on the list, and if they do the invite when accepted will not change anything.

One thing though, i noticed is that you can add your email to the invite list. I think this, too, is not a common usecase but the self-email is easy to fix as we do have that in the client. Let me know WDYT

apps/e2e/cypress/e2e/invites.cy.ts
```diff
@@ -185,6 +185,32 @@ context('Invites tests', () => {
         .should('not.exist');
     });
 
+    it.only('Should not be able to invite the email of the current user', function () {
```
jekabs-karklins, line 188: praise: Great you have added the test

jekabs-karklins (approved):

apps/frontend/src/components/common/UserManagementTable.tsx
```diff
@@ -54,6 +55,7 @@ const UserManagementTable = ({
   const isLegacyInviteFlow = featuresMap.get(
     FeatureId.EMAIL_INVITE_LEGACY
   )?.isEnabled;
+  const currentUser = useContext(UserContext)?.user;
```
jekabs-karklins, line 58: question/suggestion: Isn't the useContext(UserContext) always defined? So we could simply have
```suggestion
  const currentUser = useContext(UserContext).user;
````

jekabs-karklins:
Sorry, my bad I noticed only some minor issues after approved. You were too quick. the comments were minor. Thanks!

PR #1163, opened 2025-08-26
fix: workflow editor template route optimised for experiment safety and proposal

yoganandaness:
Since both the Experiment Safety workflow and the Proposal workflow uses same table and the datasources fetches the workflow by its id, the route differentiation in the frontend - /ExperimentWorkflowEditor/34 and /ProposalWorkflowEditor/35 does not validate the returned data, resulting in Proposal workflow content being loaded on the Experiment workflow route.

## Motivation and Context

Improve Security

## How Has This Been Tested

Manually

## Changes

In the getExperiment graphql, an additional parameter called entityType has been added to cross verify the data. If the workflow ID and the Entity type mismatches, then the graphql throws error and the same will be shown in the frontend. 


apps/backend/src/queries/WorkflowQueries.ts
```diff
@@ -22,9 +22,17 @@ export default class WorkflowQueries {
...
-  async getWorkflow(agent: UserWithRole | null, id: number) {
+  async getWorkflow(
+    agent: UserWithRole | null,
+    id: number,
+    entityType: WorkflowType
+  ) {
     const workflow = await this.dataSource.getWorkflow(id);
 
+    if (workflow?.entityType !== entityType) {
+      return null;
```
jekabs-karklins, line 33: suggestion: maybe we can add a log here as this is quiet unexpected.

jekabs-karklins (approved):
Just a small suggestion

mutambaraf (approved):

PR #1164, opened 2025-08-26
fix: restricting instrument scientists to access only the experiments of the assigned instruments

yoganandaness:
## Description

Restricting Instrument Scientists to access only the experiments of the assigned instruments

## Motivation and Context

Currently, in the experiments table of the Instrument Scientist's dashboard, all the experiments are being shown, which is undesirable. The instrument scientist should only have access to the experiments, that are belonging to the Instruments assigned to them

## How Has This Been Tested

Attached e2e test and tested manually as well. 

## Fixes

Instrument scientist able to access all experiments. 

## Changes

Experiment Data source layer, where instrument is being checked before retrieving the experiments. This way, the sql query itself will not return all the experiments. 

- [*] I have added tests to cover my changes.

### DIFF
--- a/apps/backend/src/mutations/ProposalMutations.ts
+++ b/apps/backend/src/mutations/ProposalMutations.ts
@@ -38,6 +38,7 @@ import { ChangeProposalsStatusInput } from '../resolvers/mutations/ChangeProposa
 import { CloneProposalsInput } from '../resolvers/mutations/CloneProposalMutation';
 import { CreateProposalScientistCommentArgs } from '../resolvers/mutations/CreateProposalScientistCommentMutation';
 import { ImportProposalArgs } from '../resolvers/mutations/ImportProposalMutation';
+import { NotifyProposalArgs } from '../resolvers/mutations/NotifyProposalMutation';
 import { UpdateProposalArgs } from '../resolvers/mutations/UpdateProposalMutation';
 import { UpdateProposalScientistCommentArgs } from '../resolvers/mutations/UpdateProposalScientistCommentMutation';
 import { ProposalScientistComment } from '../resolvers/types/ProposalView';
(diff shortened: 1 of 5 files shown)


apps/backend/db_patches/db_seeds/0001_ProposalForScheduling.sql
```diff
@@ -61,27 +61,64 @@ BEGIN
     )
     VALUES 
     (
-       'Test proposal'    -- title
```
jekabs-karklins, line 64: questions: what is the consideration for removing the comments?

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
```diff
@@ -672,60 +691,6 @@ export default class PostgresExperimentDataSource
       });
   }
 
-  async getExperiments({ filter }: ExperimentsArgs): Promise<Experiment[]> {
```
jekabs-karklins, line 675: suggestion: since this method name is free, do you think we could use `getExperiments` instead of `getAllExperiments`. It is shorter and more consistant with other data sources like `async getCalls(filter?: CallsFilter): Promise<Call[]>`

apps/backend/src/queries/ExperimentQueries.spec.ts
```diff
@@ -107,12 +126,15 @@ describe('ExperimentQueries', () => {
...
-      const result = await queries.getExperiment(null, experimentPk);
+      const result = await queries.getExperiment(
+        dummyUserOfficerWithRole,
+        experimentPk
+      );
       expect(result).toEqual(expected);
     });
 
     test('returns null when experiment not found', async () => {
-      const result = await queries.getExperiment(null, 999);
```
jekabs-karklins, line 115: 👍🏻

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
```diff
@@ -573,25 +572,45 @@ export default class PostgresExperimentDataSource
     offset?: number,
     sortField?: string,
     sortDirection?: string,
-    searchText?: string
```
jekabs-karklins, line 576: suggestion: I think we should move `searchText` and `instrumentScientistUserId` inside the filter, wdyt?

apps/backend/db_patches/db_seeds/0001_ProposalForScheduling.sql
yoganandaness: I felt it was quite obvious from the Values mapping with the Column name.

yoganandaness:
> Hi @yoganandaness, I just wanted to check... this seems to only affect the Experiments page for scientists - is that correct? We don't have the Experiments feature enabled and so I don't see any problem we would have with this at STFC. I'm still happy to review it once updated though.

@simonfernandes This is now ready for review

apps/backend/src/queries/ExperimentQueries.ts
```diff
@@ -1,28 +1,36 @@
...
 
 @injectable()
 export default class ExperimentQueries {
   constructor(
-    @inject(Tokens.ExperimentDataSource) public dataSource: ExperimentDataSource
+    @inject(Tokens.ExperimentDataSource)
+    public dataSource: ExperimentDataSource,
+    @inject(Tokens.UserAuthorization) private userAuth: UserAuthorization,
+    @inject(Tokens.InstrumentDataSource)
+    private instrumentDataSource: InstrumentDataSource
```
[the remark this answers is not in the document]
yoganandaness: Noted. Removed

PR #1173, opened 2025-09-08
feat: add configurable email templates

gnyiri:
## Description

Configure email templates from UI.  The PR introduces a new table for storing email templates (name, description, subject, body). The logic of assigning email templates to status actions will be the same as before: the user selects the template name from the list of available templates (populated from DB). 

Important notes: 
- email handler classes will still be able to use the original approach (file based templates). If the email template of the given name don't exist in the DB the email service will try to look for a template file with the given name and use it as a template
- the migration will only be partial and a one-time manual step will be needed from STFC after deployment: the body/subject of the email templates have to be manually copied from the physical files so that the email templates will be stored in the DB (since I only have the template names and not the content). 

## Motivation and Context

At ELI we frequently change the email templates and found it uncomfortable to write physical files on the host of the backend. 

## How Has This Been Tested

Manual tests + backend test. 

## Fixes

N/A

## Changes

See above

## Depends on

N/A

Yes


apps/backend/src/resolvers/types/EmailTemplate.ts
```diff
@@ -0,0 +1,24 @@
+import { Field, Int, ObjectType } from 'type-graphql';
+
+import { EmailTemplate as EmailTemplateOrigin } from '../../models/EmailTemplate';
+
+@ObjectType()
+export class EmailTemplate implements Partial<EmailTemplateOrigin> {
```
yoganandaness: Should this be Partial? Bcoz, i seems to me that, we are retrieving back all the fields. Correct me if i am wrong

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -441,13 +459,30 @@ const sendMail = async (
...
+  }
   try {
     const mailServiceResponse = await Promise.all(
       recipientsWithData.map(async (recipientWithData) => {
         try {
           const res = await mailService.sendMail({
             content: {
-              template_id: recipientWithData.template,
+              template_id: '',
+              db_template_id: recipientWithData.templateId,
```
yoganandaness: I think the old template_id can hold both parties. 

ie., 
for spark post, template_id = spark_post_template_1
for smtp, template_id="CLF PI Co-I Submission Email"

It is the correponding Mailservices like SparkPost or SMTP, that knows how to resolve the template_id. SparkpostMailService knows how to process the templateId and the SMTPService will try to pull it from db and get the contents. 

If above is going to work, then we wouldn't be needing db_template_id

Correct me if i am wrong.

yoganandaness (changes requested):
@gnyiri 

Apologies for the late review. 

Good work this. However, there are few areas that i felt could be improved and have commented here. Feel free to mail me or msg me, in case you need a call and we can talk. 

And there are few conflicts in the PR

Thanks

apps/backend/db_patches/0202_CreateEmailTemplatesTable.sql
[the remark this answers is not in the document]
gnyiri: Updated, thanks.

apps/backend/src/datasources/postgres/EmailTemplateDataSource.ts
```diff
@@ -0,0 +1,132 @@
...
+          description: description,
+          subject: subject,
+          body: body,
+        },
+        ['*']
+      )
+      .from('email_templates')
+      .where('email_templates.email_template_id', emailTemplateId)
+      .then((resultSet: EmailTemplateRecord[]) => {
+        return createEmailTemplateObject(resultSet[0]);
```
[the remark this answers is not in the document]
gnyiri: Fixed, thanks.

apps/backend/src/datasources/postgres/StatusActionsDataSource.ts
```diff
@@ -130,6 +131,18 @@ export default class PostgresStatusActionsDataSource
     return this.createConnectionStatusActionObject(statusActionRecord);
   }
 
+  async hasEmailTemplateIdConnectionStatusAction(
```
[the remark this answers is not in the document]
gnyiri: See comment above.

apps/backend/src/resolvers/mutations/CreateEmailTemplateMutation.ts
```diff
@@ -0,0 +1,45 @@
...
+
+@InputType()
+export class CreateEmailTemplateInput {
+  @Field(() => Int)
+  public createdByUserId: number;
+
+  @Field()
+  public name: string;
+
+  @Field({ nullable: true })
```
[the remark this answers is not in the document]
gnyiri: Fixed, thanks.

jekabs-karklins:
question: You specify template as an input text field. 
image
I wonder if  you will specify there `PUG` syntax, e.g.
```
doctype html
html(lang="en")
  head
    title= pageTitle
    script(type='text/javascript').
      if (foo) bar(1 + 5)
  body
    h1 Pug - node template engine
    #container.col
      if youAreUsingPug
        p You are amazing
      else
        p Get on it!
      p.
        Pug is a terse and simple templating language with a
        strong focus on performance and powerful features.
``` 
Maybe we need to improve the template editor in one of the next iterations. Instead of using modal maybe use full screen and syntax highlighting like (https://www.npmjs.com/package/ace-builds)

jekabs-karklins:
> > suggestion: I suggest not putting the template names in the PR, as the content of the templates appears to be placeholders and will change dynamically. I suggest putting one or two placeholder template names like "Template 1"
> 
> Where should I apply these template names in fact? Are you referring to the sql script?

Hi, yes sorry - I'm reffering to the SQL patch 
I think we can defined new templates with the Create new template 
image . Would that work?

jekabs-karklins (approved):
👍🏻

mutambaraf (commented):
Looks good just the few comments.

gnyiri:
> Just one last comment , I tried to add a template where the subject expect some dynamic variables e.g `SRA Request Xpress Proposal: #{proposals[0].proposalId}` , l was not able to make it work.

I made two changes:
- the subject will also be interpolated with the substitution variables, use it like this (with pipe): | SRA Request Xpress Proposal: #{proposals[0].proposalId} to not render html tags
- the mail service will try to load the template from file with the specified name and only if the file is not found will the template be loaded from the DB. This way, you can enforce using a template file instead of the DB template.

mutambaraf:
I think having hidden config will make this confusing for the users in case a name is set without knowing there is a file with the same template name to the one they are setting in the database. It will be nice to have an explicit flag for where to get the template.

gnyiri:
> I think having hidden config will make this confusing for the users in case a name is set without knowing there is a file with the same template name to the one they are setting in the database. It will be nice to have an explicit flag for where to get the template.

So the subject + body would either come from the DB or template files and a UI flag would allow the user to switch between?

mutambaraf:
> > I think having hidden config will make this confusing for the users in case a name is set without knowing there is a file with the same template name to the one they are setting in the database. It will be nice to have an explicit flag for where to get the template.
> 
> So the subject + body would either come from the DB or template files and a UI flag would allow the user to switch between?

Yes we can have the subject and the the body .

gnyiri:
> > > I think having hidden config will make this confusing for the users in case a name is set without knowing there is a file with the same template name to the one they are setting in the database. It will be nice to have an explicit flag for where to get the template.
> > 
> > 
> > So the subject + body would either come from the DB or template files and a UI flag would allow the user to switch between?
> 
> Yes we can have the subject and the the body .

with recent commit there will be an explicit flag to use template file with the specified template name and read the subject and body from the file and not from DB

mutambaraf (approved):
Looks good thanks

gnyiri:
> Looks good thanks

thanks for your approval. After the merge, pls do not forget to create the email templates in your DB according to your current email configuration.

gnyiri:
@yoganandaness : I tried to fix most of your findings except for those with the foreign keys. From what I understand this part of the DB will be refactored from @jekabs-karklins.

jekabs-karklins:
I have just tested this PR with a copy of our test data and it's a Go from our side as it does not break anything for us. Once merge we will look into adopting this.

yoganandaness:
Sorry, mistakenly closed. Reopened now.

apps/backend/src/queries/EmailTemplateQueries.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  @Authorized([Roles.USER_OFFICER])
+  async get(agent: UserWithRole | null, id: number) {
+    const emailTemplate = await this.dataSource.getEmailTemplate(id);
+
+    if (!emailTemplate) {
+      return null;
+    }
+
+    if (this.userAuth.isApiToken(agent) || this.userAuth.isUserOfficer(agent)) {
```
yoganandaness: I think, this is redundant. @Authorized will do the same already.

apps/backend/src/queries/EmailTemplateQueries.ts
```diff
@@ -0,0 +1,44 @@
...
+    } else {
+      return null;
+    }
+  }
+
+  @Authorized([Roles.USER_OFFICER])
+  async getAll(agent: UserWithRole | null, filter: EmailTemplatesFilter) {
+    const emailTemplates = this.dataSource.getEmailTemplates(filter);
+
+    if (this.userAuth.isApiToken(agent) || this.userAuth.isUserOfficer(agent)) {
```
yoganandaness: I think, this is redundant. @Authorized will do the same already.

apps/backend/src/datasources/postgres/EmailTemplateDataSource.ts
```diff
@@ -0,0 +1,146 @@
...
+        };
+      });
+  }
+
+  async create(
+    createdByUserId: number,
+    name: string,
+    description: string,
+    useTemplateFile: boolean,
+    subject: string,
```
yoganandaness: Contradicting with base class

The base class allows undefined for subject and body, but the implementation class does not allow. 

Looks like an edge case typescript issue, but high probability on run time error.

yoganandaness:
```suggestion
    subject?: string,
```

apps/backend/src/datasources/postgres/EmailTemplateDataSource.ts
```diff
@@ -0,0 +1,146 @@
...
+    body: string
```
yoganandaness:
```suggestion
    body?: string
```

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -372,6 +373,16 @@ export interface CallRecord {
   readonly experiment_workflow_id: number;
 }
 
+export interface EmailTemplateRecord {
+  readonly email_template_id: number;
+  readonly created_by: number;
+  readonly name: string;
+  readonly description: string;
+  readonly subject: string;
```
yoganandaness:
```suggestion
  readonly subject?: string | nullable;
```

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -372,6 +373,16 @@ export interface CallRecord {
...
+  readonly body: string;
```
yoganandaness:
```suggestion
  readonly body?: string | nullable;
```

apps/backend/src/models/EmailTemplate.ts
```diff
@@ -0,0 +1,11 @@
+export class EmailTemplate {
+  constructor(
+    public id: number,
+    public createdByUserId: number,
+    public name: string,
+    public description: string,
+    public useTemplateFile: boolean,
+    public subject: string,
```
yoganandaness:
```suggestion
    public subject?: string | null,
```

apps/backend/src/models/EmailTemplate.ts
```diff
@@ -0,0 +1,11 @@
...
+    public body: string
```
yoganandaness:
```suggestion
    public body?: string | null
```

apps/backend/src/models/EmailTemplate.ts
yoganandaness: aligning with db schema

apps/backend/src/models/EmailTemplate.ts
yoganandaness: aligning with db schema

apps/backend/src/mutations/EmailTemplateMutations.ts
```diff
@@ -0,0 +1,121 @@
...
+      return rejection('Email template not found', { emailTemplateId });
+    }
+
+    try {
+      const result = await this.dataSource.delete(emailTemplateId);
+
+      return result;
+    } catch (error) {
+      // NOTE: We are explicitly casting error to { code: string } type because it is the easiest solution for now and because it's type is a bit difficult to determine because of knexjs not returning typed error message.
+      if ((error as { code: string }).code === '23503') {
```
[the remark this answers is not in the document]
yoganandaness: Residual code from copy/paste. Fixed.

apps/backend/src/resolvers/mutations/CreateEmailTemplateMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+  public name: string;
+
+  @Field(() => String)
+  public description: string;
+
+  @Field(() => Boolean)
+  public useTemplateFile: boolean;
+
+  @Field(() => String, { nullable: true })
+  public subject?: string;
```
yoganandaness, line 18: This is good aligning with the database schema. It would be good to have the same alignment down in the layer of business layer and datasource layer

apps/backend/src/resolvers/types/EmailTemplate.ts
```diff
@@ -0,0 +1,27 @@
...
+  public name: string;
+
+  @Field(() => String, { nullable: true })
+  public description: string;
+
+  @Field(() => Boolean)
+  public useTemplateFile: boolean;
+
+  @Field(() => String, { nullable: true })
+  public subject: string;
```
yoganandaness, line 23:
```suggestion
  public subject?: string;
```

apps/backend/src/resolvers/types/EmailTemplate.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  @Field(() => String, { nullable: true })
+  public body: string;
```
yoganandaness, line 26:
```suggestion
  public body?: string;
```

apps/backend/src/eventHandlers/email/eliEmailHandler.ts
```diff
@@ -357,20 +328,37 @@ export async function eliEmailHandler(event: ApplicationEvent) {
   }
 }
 
+function gettemplateNameForRole(role: UserRole): string {
```
yoganandaness:
```suggestion
function getTemplateNameForRole(role: UserRole): string {
```

yoganandaness: I think this is not used anywhere

apps/frontend/src/components/template/EmailTemplateTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <CreateUpdateEmailTemplate
+      emailTemplate={editEmailTemplate}
+      close={(emailTemplate: EmailTemplateFragment | null) =>
+        !!editEmailTemplate ? onUpdate(emailTemplate) : onCreate(emailTemplate)
+      }
+    />
+  );
+
+  return (
+    <>
```
yoganandaness: Not needed

apps/frontend/src/components/template/CreateUpdateEmailTemplate.tsx
```diff
@@ -0,0 +1,148 @@
...
+  emailTemplate: EmailTemplateFragment | null;
+};
+
+const CreateUpdateEmailTemplate = ({
+  close,
+  emailTemplate,
+}: CreateUpdateEmailTemplateProps) => {
+  const { t } = useTranslation();
+  const { api, isExecutingCall } = useDataApiWithFeedback();
+  // const [useTemplateFile, setUseTemplateFile] = useState(
```
yoganandaness: Can be removed if not needed

apps/frontend/src/components/template/CreateUpdateEmailTemplate.tsx
```diff
@@ -0,0 +1,148 @@
...
+    subject: emailTemplate?.subject || '',
+    body: emailTemplate?.body || '',
+    createdByUserId:
+      emailTemplate?.createdByUserId ||
+      (getCurrentUser()?.user.id as number) ||
+      0,
+  };
+
+  return (
+    <Formik
```
yoganandaness, line 41: We need to have validation schema here aligning with the db structure. Like in other places.

apps/frontend/src/components/template/CreateUpdateEmailTemplate.tsx
```diff
@@ -0,0 +1,148 @@
...
+            data-cy="name"
+            disabled={isExecutingCall}
+            required
+          />
+          <Field
+            name="description"
+            id="description"
+            label="Description"
+            type="text"
+            component={TextField}
```
yoganandaness, line 94: Should this be a multiline?

apps/frontend/src/components/template/CreateUpdateEmailTemplate.tsx
yoganandaness: And e2e for the same

apps/backend/src/queries/StatusActionQueries.ts
```diff
@@ -59,15 +63,18 @@ export default class StatusActionQueries {
...
-          id: item.id,
-          name: item.name,
-        }));
+        logger.logInfo('Email templates', { emailTemplatesResult });
 
-        return new EmailActionDefaultConfig(allEmailRecipients, emailTemplates);
+        return new EmailActionDefaultConfig(
+          allEmailRecipients,
+          emailTemplatesResult.results.map((e) => ({
+            id: e.name,
```
yoganandaness:
```suggestion
            id: e.id,
```

yoganandaness: It needs to be id. Else it would break for ESS, as id and name  of a template are different in Sparkpost .

apps/frontend/src/components/settings/workflow/EmailActionConfig.tsx
```diff
@@ -238,7 +254,7 @@ const EmailActionConfig = ({
                         options={emailTemplates || []}
                         getOptionLabel={(option) => option.name}
                         isOptionEqualToValue={(option, value) =>
-                          option.id === value.id
+                          option.name === value.name
```
yoganandaness: Can we continue using id. For ESS, id is the unique identifier. Name is not.

apps/frontend/src/components/settings/workflow/EmailActionConfig.tsx
```diff
@@ -171,6 +172,21 @@ const EmailActionConfig = ({
     }
   };
 
+  const getEmailTemplate = (foundRecipientWithEmailTemplateIndex: number) => {
+    if (foundRecipientWithEmailTemplateIndex !== -1) {
+      return (
+        emailTemplates.find(
+          (template) =>
+            template.name ===
```
yoganandaness: This will also break for ESS. In Sparkpost, id is unique. Name is not.

apps/backend/src/datasources/postgres/records.ts
yoganandaness: subject can be null as per the db

apps/backend/src/datasources/postgres/records.ts
yoganandaness: body can be null as per the db

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -1400,7 +1425,7 @@ export interface InviteRecord {
   readonly claimed_at: Date | null;
   readonly is_email_sent: boolean;
   readonly expires_at: Date | null;
-  readonly template_id: EmailTemplateId | null;
+  readonly template_id: EmailTemplateName | null;
```
yoganandaness: Can we use id instead of name. For Sparkpost, id is unique indentifier.

apps/backend/src/datasources/postgres/StatusActionsDataSource.ts
```diff
@@ -130,6 +131,18 @@ export default class PostgresStatusActionsDataSource
     return this.createConnectionStatusActionObject(statusActionRecord);
   }
 
+  async hasEmailTemplateIdConnectionStatusAction(
+    emailTemplateName: string
+  ): Promise<boolean> {
+    const fromClause = "config->'recipientsWithEmailTemplate'";
+    const pattern = `\'[{"emailTemplate": {"name": "${emailTemplateName}"}}]\'`;
```
yoganandaness: should be id. as the function name suggests

yoganandaness:
```suggestion
    const pattern = `\'[{"emailTemplate": {"id": "${emailTemplateId}"}}]\'`;
```

apps/backend/src/datasources/StatusActionsDataSource.ts
```diff
@@ -13,6 +13,10 @@ export interface StatusActionsDataSource {
     workflowConnectionId: number,
     statusActionId: number
   ): Promise<ConnectionHasStatusAction>;
+  hasEmailTemplateIdConnectionStatusAction(
+    emailTemplateName: string
```
yoganandaness:
```suggestion
    emailTemplateId: string
```

apps/backend/src/eventHandlers/MailService/MailService.ts
yoganandaness: Can we remove this? as this is not being used anywhere

apps/backend/src/eventHandlers/MailService/MailService.ts
```diff
@@ -28,7 +28,7 @@ export type STFCEmailTemplate = {
 };
 
 export type ELIEmailTemplate = {
```
yoganandaness:
```suggestion
export type SMTPEmailTemplate = {
```

apps/backend/src/eventHandlers/email/emailTemplateName.ts
```diff
@@ -0,0 +1,20 @@
+export enum EmailTemplateName {
```
yoganandaness:
```suggestion
export enum EmailTemplateId {
```

apps/backend/src/eventHandlers/email/emailTemplateId.ts
yoganandaness, line 1: apps/backend/src/eventHandlers/email/emailTemplateId.ts

apps/backend/src/datasources/mockups/InviteDataSource.ts
```diff
@@ -1,7 +1,7 @@
 import { inject, injectable } from 'tsyringe';
 
 import { Tokens } from '../../config/Tokens';
-import { EmailTemplateId } from '../../eventHandlers/email/essEmailHandler';
+import { EmailTemplateName } from '../../eventHandlers/email/emailTemplateName';
```
yoganandaness:
```suggestion
import { EmailTemplateId } from '../../eventHandlers/email/emailTemplateId';
```

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -90,7 +91,7 @@ export const emailStatusActionRecipient = async (
   loggedInUserId?: number | null
 ) => {
   const proposalPks = proposals.map((proposal) => proposal.primaryKey);
-  const templateMessage = recipientWithTemplate.emailTemplate.id;
+  const emailTemplateName = recipientWithTemplate.emailTemplate.name;
```
yoganandaness:
```suggestion
  const emailTemplateId = recipientWithTemplate.emailTemplate.id;
```

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -443,13 +448,30 @@ const sendMail = async (
...
+  );
+
   try {
     const mailServiceResponse = await Promise.all(
       recipientsWithData.map(async (recipientWithData) => {
         try {
           const res = await mailService.sendMail({
             content: {
-              template_id: recipientWithData.template,
+              template: emailTemplate?.name || '',
```
yoganandaness:
```suggestion
              template: emailTemplate?.id || '',
```

apps/backend/src/statusActionEngine/emailActionHandler.ts
```diff
@@ -443,13 +448,30 @@ const sendMail = async (
...
+    { recipientsWithData }
+  );
+
   try {
     const mailServiceResponse = await Promise.all(
       recipientsWithData.map(async (recipientWithData) => {
         try {
           const res = await mailService.sendMail({
             content: {
-              template_id: recipientWithData.template,
```
yoganandaness, line 452: If we are not going to use recipientWithData.template, we can remove this key `template` from `EmailReadyType`

apps/backend/src/statusActionEngine/statusActionUtils.ts
```diff
@@ -214,7 +214,7 @@ export const getEmailReadyArrayOfUsersAndProposals = async (
         emailReadyUsersWithProposals.push({
           id: recipientsWithEmailTemplate.recipient.name,
           proposals: [proposal],
-          template: recipientsWithEmailTemplate.emailTemplate.id,
+          template: recipientsWithEmailTemplate.emailTemplate.name,
```
yoganandaness: Do we need this key, as i can see, it is not being used anywhere.

apps/backend/src/models/Invite.ts
```diff
@@ -11,6 +11,6 @@ export class Invite {
     public claimedByUserId: number | null,
     public isEmailSent: boolean,
     public expiresAt: Date | null,
-    public templateId: EmailTemplateId | null
+    public templateId: EmailTemplateName | null
```
yoganandaness:
```suggestion
    public templateId: EmailTemplateId | null
```

apps/backend/src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -76,6 +107,57 @@ export class SMTPMailService extends MailService {
     );
   }
 
+  private async getEmailTemplate(options: EmailSettings): Promise<{
```
yoganandaness:
```suggestion
  private async compileEmailTemplate(options: EmailSettings): Promise<{
```

yoganandaness: Just a naming suggestion, so that it does not resemble same as getEmailTemplates

package.json
```diff
@@ -50,9 +50,12 @@
   ],
   "dependencies": {
     "concurrently": "^9.0.1",
+    "ejs": "^3.1.10",
     "wait-on": "^9.0.1"
   },
   "devDependencies": {
+    "@types/pug": "^2.0.10",
```
yoganandaness: Do we need types pug here? I can see it is only used in backend

apps/frontend/src/components/template/CreateUpdateEmailTemplate.tsx
```diff
@@ -0,0 +1,148 @@
...
+  emailTemplate,
+}: CreateUpdateEmailTemplateProps) => {
+  const { t } = useTranslation();
+  const { api, isExecutingCall } = useDataApiWithFeedback();
+  // const [useTemplateFile, setUseTemplateFile] = useState(
+  //   emailTemplate?.useTemplateFile || false
+  // );
+
+  const initialValues = {
+    id: emailTemplate?.id || 0,
```
yoganandaness, line 32: do we need to initialize id, createdByUserId as they are not the part of the form

apps/backend/src/mutations/EmailTemplateMutations.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  @EventBus(Event.EMAIL_TEMPLATE_CREATED)
+  @Authorized([Roles.USER_OFFICER])
+  async create(
+    agent: UserWithRole | null,
+    args: CreateEmailTemplateInput
+  ): Promise<EmailTemplate | Rejection> {
+    try {
+      const createdEmailTemplate = await this.dataSource.create(
+        args.createdByUserId,
```
yoganandaness:
```suggestion
        agent.id,
```

apps/backend/src/resolvers/mutations/CreateEmailTemplateMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+  Resolver,
+} from 'type-graphql';
+
+import { ResolverContext } from '../../context';
+import { EmailTemplate } from '../types/EmailTemplate';
+
+@InputType()
+export class CreateEmailTemplateInput {
+  @Field(() => Int)
+  public createdByUserId: number;
```
yoganandaness: This should not be passed as the parameter due to potential security conflict. This should be filled by the backend based on which user has signed in.

apps/backend/package.json
```diff
@@ -92,6 +92,7 @@
     "@types/content-disposition": "^0.5.4",
     "@types/cors": "^2.8.13",
     "@types/cron": "^1.7.3",
+    "@types/ejs": "^3.1.5",
```
yoganandaness, line 97: Are we using ejs?

package.json
```diff
@@ -50,9 +50,12 @@
   ],
   "dependencies": {
     "concurrently": "^9.0.1",
+    "ejs": "^3.1.10",
```
yoganandaness: Are we using ejs?

package.json
```diff
@@ -50,9 +50,12 @@
...
     "wait-on": "^9.0.1"
   },
   "devDependencies": {
+    "@types/pug": "^2.0.10",
+    "@types/react-dom": "^19.2.2",
```
yoganandaness, line 56: Do we need this here? It is needed only in backend.

yoganandaness (changes requested):
I can repeatedly see that you have used template `name` instead of template `email_template_id`. As for as Sparkpost is concerned, `id` is important over `name`. Passing `name` will not send email for Sparkpost and hence will break for ESS. Is it going to be difficult using email template id from the database for the purpose of SMTP? If yes, we need to find a generic solution that satisfies both SMTP and Sparkpost.  @mutambaraf Feel free to add your thoughts, as i understand STFC also uses SMTP. 

Sorry for having this review very late. After continuous refinement, i now have multiple comments, that needs to be addressed. I could be wrong in some case, for which kindly clarify me.

package.json
gnyiri: Fixed

yoganandaness (changes requested):
I can repeatedly see that you have used template `name` instead of template `email_template_id`. As for as Sparkpost is concerned, `id` is important over `name`. Passing `name` will not send email for Sparkpost and hence will break for ESS. Is it going to be difficult using email template id from the database for the purpose of SMTP? If yes, we need to find a generic solution that satisfies both SMTP and Sparkpost.  

Sorry for having this review very late. After continuous refinement, i now have multiple comments, that needs to be addressed. I could be wrong in some case, for which kindly clarify me.

apps/backend/src/datasources/postgres/records.ts
gnyiri: You mean the id as a string?

apps/frontend/src/components/template/CreateUpdateEmailTemplate.tsx
gnyiri: We need it when updating instead of creating.

apps/frontend/src/components/template/CreateUpdateEmailTemplate.tsx
gnyiri: Added validation shcemes and e2e tests.

yoganandaness (approved):

jekabs-karklins:
❤️

PR #1182, opened 2025-09-12
feat: add option to force email notification

gnyiri:
## Description

At ELI we would sometimes send multiple notification messages for the same proposals. This is only possible if we force the notification flag set when the proposal is notified. 

## Motivation and Context

See above

## How Has This Been Tested

Manual tests. 

## Fixes

Change notification dialog and put a new checkbox to force the notification (do not block the trigger if the proposal was already notified). 

## Changes

Frontend + backend. DB not affected. 

## Depends on

N/A

### DIFF
--- a/apps/backend/src/mutations/ShipmentMutations.ts
+++ b/apps/backend/src/mutations/ShipmentMutations.ts
@@ -9,8 +9,7 @@ import { ProposalDataSource } from '../datasources/ProposalDataSource';
 import { QuestionaryDataSource } from '../datasources/QuestionaryDataSource';
 import { ShipmentDataSource } from '../datasources/ShipmentDataSource';
 import { TemplateDataSource } from '../datasources/TemplateDataSource';
-import { Authorized, EventBus } from '../decorators';
-import { Event } from '../events/event.enum';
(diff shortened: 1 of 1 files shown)


apps/backend/src/mutations/ProposalMutations.ts
```diff
@@ -380,13 +380,21 @@ export default class ProposalMutations {
   @Authorized([Roles.USER_OFFICER])
   async notify(
     user: UserWithRole | null,
-    { proposalPk }: { proposalPk: number }
+    {
+      proposalPk,
+      ignoreNotifiedFlag,
+    }: { proposalPk: number; ignoreNotifiedFlag: boolean }
```
jekabs-karklins, line 386: suggestion: maybe you can use the datatype we have already defined in `apps/backend/src/resolvers/mutations/NotifyProposalMutation.ts` 
`NotifyProposalArgs` , so we don't have to define it twice

jekabs-karklins (commented):
Looks good to me—just one small comment. Feel free to re-request review.

jekabs-karklins (approved):

PR #1194, opened 2025-09-23
fix: add count_self_hosted_runners job to optimize e2e testing based …

jekabs-karklins:
## Description
This PR optimizes the e2e testing by introducing a job to count self-hosted runners and distribute the e2e tests based on the available runners.

## Motivation and Context
This change is required to improve efficiency and speed of the e2e testing process by better utilizing available resources and avoiding bottlenecks.

## Changes
1. Introduced a new job `count_self_hosted_runners` to count the number of idle self-hosted runners. This helps in the distribution of e2e tests.
2. The `e2e` job is modified to depend on `count_self_hosted_runners` and the distribution of tests is based on the count of idle runners.
3. Removed redundant repository download step in e2e tests.

## Fixes Jira Issue


yoganandaness (approved):
Seems good, although looking forward to see in action.

TCMeldrum (approved):

TCMeldrum:
If we have updated the job names  don't we have to update the jobs in the repos settings as well?

PR #1202, opened 2025-09-26
fix: removed proposal sample review submitted event for shipment update

yoganandaness:
## Description

Currently, when a shipment is updated, the event PROPOSAL_SAMPLE_REVIEW_SUBMITTED is triggered, which in turn triggers the Proposal workflow and Experiment workflow, both of which are unneeded and leading to errors. The Sample review process is completely isolated from shipment process. 

## Motivation and Context

To ensure that the Sample review process and Shipment process are completely independant

## How Has This Been Tested

Manually

## Changes

Removed the decorator for Update Shipment mutation, which will therefore avoid triggering of the workflow


jekabs-karklins (approved):
👍🏻

TCMeldrum (approved):

PR #1204, opened 2025-09-29
fix: instrument availability time can be null

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1172

## Description
This PR allows instrument availability time to be null in the AssignedInstrumentsTable component.

## Motivation and Context
This change is required to handle cases where the instrument availability time is unknown or not provided as the call can be se up ahead of time.

## Changes
- Added a condition in the validation logic of the AssignedInstrumentsTable component to allow `null` for the `availabilityTime` property.

## Additional Context
The motivation behind this change comes from the need to handle real-life scenarios where sometimes the availability time of an instrument might not be known, hence allowing `null` values helps to maintain the integrity of the application.

## Fixes Jira Issue

### DIFF
--- a/apps/frontend/src/components/call/AssignedInstrumentsTable.tsx
+++ b/apps/frontend/src/components/call/AssignedInstrumentsTable.tsx
@@ -148,6 +148,13 @@ const AssignedInstrumentsTable = ({
           return { isValid: true };
         }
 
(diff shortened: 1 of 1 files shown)


jekabs-karklins (approved):
Great, this aligns with table definition in the database, where availability_time is already nullable 👍🏻

PR #1206, opened 2025-09-29
feat: allow closed calls fap reviews to appear in fap page

TCMeldrum:


apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -305,11 +305,14 @@ export default class PostgresFapDataSource implements FapDataSource {
...
-    callId?: number | null;
-    instrumentId?: number | null;
-  }): Promise<FapProposal[]> {
+  async getFapProposals(
+    filter: {
+      fapId: number;
+      callId?: number | null;
+      instrumentId?: number | null;
+    },
+    currentUserNumber?: number
```
yoganandaness, line 314: Is there any other better place to bring in the business logic as the datasource can continue to act as a stateless source of fetching the data and that the Business layer(Query and Mutation) could decide over the logic. 

I can understand that this is the query to fetch proposals of the FAP_REVIEWER, where he/she is himself/herself not the Principal Investigator or Co-Proposers. Since this additional parameter currentUserNumber is not self documentation and adds in more complexity in understanding, i would probably move this logic to the business layer, where i will filter the returned list by removing unneeded ones.

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -322,15 +325,33 @@ export default class PostgresFapDataSource implements FapDataSource {
...
+          query.andWhere(function () {
+            this.where('s.short_code', 'ilike', 'FAP_%');
+          });
         }
+
         if (filter.instrumentId) {
           query.andWhere('fp.instrument_id', filter.instrumentId);
         }
+
+        if (currentUserNumber) {
```
yoganandaness, line 344: Is there any specific reason to ignore the self proposals?

apps/backend/src/datasources/postgres/FapDataSource.ts
```diff
@@ -322,15 +325,33 @@ export default class PostgresFapDataSource implements FapDataSource {
...
+            this.where('p.submitted', true);
           });
 
-        if (filter.callId) {
+        if (filter.callId && filter.callId > 0) {
           query.andWhere('fp.call_id', filter.callId);
+        } else if (filter.callId === -1) {
+          // -1 is the value we use for "current"
+          query.andWhere(function () {
+            this.where('s.short_code', 'ilike', 'FAP_%');
```
yoganandaness, line 336: Could you write more about this?

It looks to me that, when the callId is -1, we return Proposals with statuses(FAP_) or else, return everything. 

Also, is there any other best way to avoid -1 and implement more explicit filter?

apps/backend/src/datasources/FapDataSource.ts
```diff
@@ -71,11 +71,14 @@ export interface FapDataSource {
...
-    callId?: number | null;
-    instrumentId?: number | null;
-  }): Promise<FapProposal[]>;
+  getFapProposals(
+    filter: {
+      fapId: number;
+      callId?: number | null;
+      instrumentId?: number | null;
+    },
+    currentUerId?: number
```
yoganandaness, line 80:
```suggestion
    currentUserId?: number
```

Spell check. But I am still trying to figure out the best way to rename for making it more self documented

apps/backend/src/queries/FapQueries.ts
```diff
@@ -76,7 +76,12 @@ export default class FapQueries {
       this.userAuth.isUserOfficer(agent) ||
       (await this.userAuth.isMemberOfFap(agent, fapId))
     ) {
-      return this.dataSource.getFapProposals({ fapId, callId, instrumentId });
+      return this.dataSource.getFapProposals(
```
yoganandaness, line 79: I understand that this is to avoid FAP_REVIEWERS from self reviewing their own Proposals. If i am right, here is my suggestion. Let me know in case of any concerns. 

Instead of querying at the data source level, i would instead filter the response(As the result won't typically be a lot). This way, the datasource continue to remain stateless and generic.

yoganandaness: Or i would not mind writing a separate query for the FAP Reviewers.

yoganandaness (commented):
I feel that some of the file changes and the PR description is slightly not aligned. I could be wrong here. 

I would also like to add in a point that, if we want to show the Proposals belonging to Closed call, it can be new component(Like archived) all together for better usability, this way the Call filter component continues to remain intact and generic.

yoganandaness:
@TCMeldrum Although i understand a bit about the requirement, i am definitely a step behind what you have. I have reviewed from the technical stability point of view. In case, my comments are completely irrelevant, I am sorry for the annoying comments and it is best if we could catch up and clarify things. 

Looking forward.

PR #1228, opened 2025-10-16
fix: user office experiment safety document disables

yoganandaness:
## Description

Fixed refreshing of Exp safety download button visibility by doing request polling of 3 times with delay of 2 seconds each, thereby avoid hard page refresh. Attached Video and Screenshot. 

## Motivation and Context

When an experiment safety review is approved, the download pdf button should be enabled. Since the status is computed through workflow, it takes some time. Currently, the browser needs to be refreshed to get the visibility of the button. However since this is an inconvenience, request polling has been implemented that fetches the new status and based on the new status, the button visibility is decided

## How Has This Been Tested

Manually and attached e2e

## Fixes

Avoiding Browser reload

## Changes

In the Experiment Safety review summary page, the download button is placed. Instead of window reload, request polling using timeout has been implemented

In addition to the above, the menu has been slightly adjusted, where the experiment safety workflow has been moved next to the Proposal workflow


yoganandaness:
Screenshot 2025-10-16 at 17 48 44

simonfernandes (approved):
It looks good. I assumed this is branched from https://github.com/UserOfficeProject/user-office-core/pull/1209 and so I skipped over reviewing those profile button changes

PR #1259, opened 2025-11-13
chore: upgrade eslint backend to 9.0

### LINKED ISSUE #1488: Update ESlint version
We are currently using a deprecated version of ESLint (eslint@8.57.1), which is no longer receiving support, bug fixes, or security patches from the ESLint team. We've started seeing the following warning: eslint@8.57.1: This version is no longer supported. Please see https://eslint.org/version-support for other options. Continuing to use an unsupported version introduces a risk of encountering unaddressed bugs and missing out on new linting capabilities.

TCMeldrum:
Part of https://github.com/UserOfficeProject/issue-tracker/issues/1488

## Description

Our eslint version is end of life and out of support: https://eslint.org/version-support/ so I have updated the backend to 9.0.0 and will open new prs for the frontend and tests to avoid the prs form getting too big. 9.0.0 comes with a new config format so I have migreated to that however this has come with some rule updated and changes. I have tryed to keep the rules as consistant as before but in some cases it doesnt seem possible right now e.g. https://github.com/import-js/eslint-plugin-import/pull/2721. I have also updated the code base to follow some new (or previously missing? idk) rules that seem sensible to follow.

**I suggest reviewing this PR commit by commit to see the rule updates applyed one at a time and not all at once.**

## Motivation and Context

Eslint v8 is end of life and out of support

## How Has This Been Tested

Created new eslint.config.js for v9, updated to code base to follow updated rules. 


Scott-James-Hurley (approved):

jekabs-karklins:
question/suggestion: I see this cleans up nicely many places when it comes to formatting, at the same time I wonder if we could be less oppinionated on the import order. A lot of times it causes unneeded conflicts and I rarely do I find it helpful if imports are alphbetical with linespaces withing the group.  Do you feel the same way and if so is there a way to downprioritize the import rules

TCMeldrum:
I haven't felt that it is too opressive my code editor is very good at correcting it with one click, I also quite like the seperation of the external npm modules and relivite imports. I think it also nice to be able to see which module is being imported when we have some classes with the same name e.g. [Fap](https://github.com/UserOfficeProject/user-office-core/blob/4e6d7b63668a8e1b660866ecb223584ddd53effb/apps/backend/src/resolvers/types/Fap.ts#L27) and [Fap](https://github.com/UserOfficeProject/user-office-core/blob/4e6d7b63668a8e1b660866ecb223584ddd53effb/apps/backend/src/models/Fap.ts#L8). 

That being said I am not wedded to it, happy to be over ruled.

jekabs-karklins (approved):

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -88,42 +83,7 @@ import {
 } from '../../resolvers/types/ProposalView';
 import { FeedbackStatus } from './../../models/Feedback';
 
-// Adds types to datasources: https://knexjs.org/guide/#typescript
-declare module 'knex/types/tables' {
```
TCMeldrum, line 92: I have just removed this, it didnt seem to be typechecking and seems to work without it

PR #1278, opened 2025-11-27
fix: proposaltablereviewer call and instrument filters

### LINKED ISSUE #1498: ProposalTableReviewer Call and Instrument filters show too many options
The call and instrument filters dropdown show all calls and instruments when they should only show the relevent ones.

TCMeldrum:
Closes UserOfficeProject/issue-tracker#1498

## Description
This PR enhances the functionality of proposal review by ensuring that reviewers only see calls and instruments that are relevant to their assigned reviews.

I have added new endpoint for these as I could not find a easy way of intergrating it into the current methods.   

## Motivation and Context
This change is required to streamline the review process and make it more efficient. It solves the problem of reviewers having to sift through irrelevant calls and instruments not assigned to their FAP.

## Changes
- Added methods to retrieve calls and instruments based on FAP IDs in the CallDataSource and InstrumentDataSource interfaces.
- Implemented these methods in the corresponding data sources.
- Updated queries in CallQueries and InstrumentQueries to utilize these new methods.
- Updated resolvers to use the updated queries.
- Updated frontend to use the new queries and display relevant data to reviewers. 

Please note that these changes have been tested and verified to work as expected.

## Fixes Jira Issue


Scott-James-Hurley (approved):
Looks great. Worth noting that a similar thing with happens with the call filter but in practice I don't see how it would affect users. A reviewer wouldn't be assigned proposals from a closed call to review:

Image

apps/backend/src/queries/CallQueries.ts
```diff
@@ -57,4 +60,21 @@ export default class CallQueries {
   async getCallOfAnswersProposal(user: UserWithRole | null, answerId: number) {
     return this.dataSource.getCallByAnswerIdProposal(answerId);
   }
+
+  async getCallsOfReviewer(agent: UserWithRole | null) {
```
yoganandaness, line 75: Recommending to add @Authorized for additional layer of security

ex., We dont want to make an sql query, if an instrument scientist is trying to call this query.

apps/backend/src/queries/InstrumentQueries.ts
```diff
@@ -98,6 +101,24 @@ export default class InstrumentQueries {
     }
   }
 
+  @Authorized()
+  async getFapReviewerInstruments(agent: UserWithRole | null) {
+    if (!agent || !agent.id) {
+      return [];
+    }
+
+    if (!agent.currentRole) {
```
yoganandaness: Can we club this and above?

yoganandaness (changes requested):
@TCMeldrum I have given my comments. 

But, this looks a bit related to multi facility feature. I could be wrong. 

@jekabs-karklins If it is, do you think your PR 1282 could solve this and make this one redundant?

TCMeldrum:
> @TCMeldrum I have given my comments.
> 
> But, this looks a bit related to multi facility feature. I could be wrong.
> 
> @jekabs-karklins If it is, do you think your PR 1282 could solve this and make this one redundant?

This is just updating the fap reviews table to only show relivent calls as it shows all calls right now. It would be relevent as it could be different for 2 users in the same facility.

apps/backend/src/queries/CallQueries.ts
```diff
@@ -57,4 +60,18 @@ export default class CallQueries {
   async getCallOfAnswersProposal(user: UserWithRole | null, answerId: number) {
     return this.dataSource.getCallByAnswerIdProposal(answerId);
   }
+
+  @Authorized([Roles.FAP_REVIEWER, Roles.FAP_CHAIR, Roles.FAP_SECRETARY])
```
yoganandaness, line 74: Appreciated. 

And since we don't want the old query to be exposed to the FAPs, could we remove the Authorization access to those queries including Calls and Instruments?


An e2e will be amazing.

apps/backend/src/queries/InstrumentQueries.ts
```diff
@@ -98,6 +101,20 @@ export default class InstrumentQueries {
     }
   }
 
+  @Authorized([Roles.FAP_REVIEWER, Roles.FAP_CHAIR, Roles.FAP_SECRETARY])
```
yoganandaness, line 98: Appreciated. 

And since we don't want the old query to be exposed to the FAPs, could we remove the Authorization access to those queries including Calls and Instruments?


An e2e will be amazing.

yoganandaness (changes requested):

TCMeldrum:
I have removed the autherzied persmissions on the instrument query, but there wern't any on the calls query. Thanks for flagging that @yoganandaness

yoganandaness (approved):

PR #1282, opened 2025-12-03
feat: introducing root_role and derived_role (based on tags)

bolmsten:
Screenshot 2025-12-10 at 11 46 35

## Description

My proposal is to continue using pre-defined roles (now called root roles) and dashboards, but allow for the creation of dervied roles that has two attributes: “data access” and “permissions/config”.  The derived roles link to the root roles and inherits the dashboard and the permissions of the root role. This means that much of the code can stay the same. 

For example: I added a new root role with associated dashboard and created two derived roles ISIS_READ and CLF_READ—both using  from the root role proposal_reader role. The only difference between the dervied roles is the data access tags. That would allow users to login and see a proposal table with only the proposals that has the instrument associated with their data access tag. 

We can do minor configurations for each dashboard using a permission/config defined on the derived role and we can set the data access either by using tags. 

It would also allow for different facilities to rename roles like FAP Reviewer to PEP Reviewer.

A derived role has the same functionality like the root role until tagged and then the data access is determined by the tag. 

image

image


apps/backend/db_patches/db_seeds/0002_InstrumentScientiests.sql
jekabs-karklins: suggestion: Looks like an empty file

apps/backend/db_patches/0208_AddTagsToRoles.sql
```diff
@@ -0,0 +1,25 @@
+DO
+$$
+BEGIN
+  IF register_patch('Add tags to role and set variable for default roles', 'Fredrik Bolmsten', 'tag_to_role', '2025-12-09') THEN
```
mutambaraf, line 4: We may need to update this date its now to far in the past.

jekabs-karklins:
Hello @mutambaraf,

After a couple of false positives, I got the PR to pass the end‑to‑end tests.

This PR introduces the proposal_reader role. It lets a tagged user view all proposals for the tagged instruments or calls—same access as a User Officer, but without permission to modify anything

mutambaraf (commented):
Just this one comment still trying to get it to work on my dev.

apps/backend/src/buildContext.ts
```diff
@@ -151,3 +155,5 @@ const context: BasicResolverContext = {
 };
 
 export default context;
+
```
mutambaraf: Do we also need this comment

apps/e2e/cypress/e2e/tag.cy.ts
```diff
@@ -19,7 +19,10 @@ const instrument1 = {
 context('Tag tests', () => {
   beforeEach(function () {
     cy.getAndStoreFeaturesEnabled().then(() => {
-      if (!featureFlags.getEnabledFeatures().get(FeatureId.TAGS)) {
+      if (
+        !featureFlags.getEnabledFeatures().get(FeatureId.TAGS) ||
+        !featureFlags.getEnabledFeatures().get(FeatureId.OAUTH) // TODO implement method setUserRoles in StfcUserDataSource and remove this condition
```
mutambaraf, line 24: I have opened this [issue](https://github.com/UserOfficeProject/issue-tracker/issues/1534) to remove this comment.

apps/frontend/src/components/admin/RoleModal.tsx
```diff
@@ -0,0 +1,217 @@
...
+        })
+      : api({
+          toastSuccessMessage: 'Role created successfully',
+        }).createRole({
+          args: {
+            shortCode,
+            title,
+            description,
+            permissions,
+          },
```
mutambaraf, line 124: We have not implemented this but l am getting a success message when l try to create a role we may need to find a way to wait because the backend looks okay.

apps/backend/src/resolvers/mutations/CreateRoleMutation.ts
```diff
@@ -0,0 +1,54 @@
...
+  @Mutation(() => CreateRoleResponse)
+  async createRole(
+    @Arg('args') args: CreateRoleArgs,
+    @Ctx() context: ResolverContext
+  ): Promise<CreateRoleResponse> {
+    const role = (await context.mutations.user.createRole(
+      context.user,
+      args
+    )) as Role | Error;
+    if (role instanceof Error) {
```
mutambaraf: It may be nice to also log the error.

mutambaraf (commented):
Looks good l have added some few comments

apps/backend/src/buildContext.ts
jekabs-karklins: Thanks, removed 👍

apps/backend/src/resolvers/mutations/CreateRoleMutation.ts
jekabs-karklins: 👍

apps/e2e/cypress/e2e/tag.cy.ts
jekabs-karklins: Thanks 👍

apps/frontend/src/components/admin/RoleModal.tsx
jekabs-karklins: createRole is creating a role in UserOffice so it should work for everyone, I don't think you need to implement this.

mutambaraf: There might be a problem then , because I tried several times to create a custom role and nothing was being created.

jekabs-karklins: Hi, @mutambaraf ,

Thanks for checking this out. I just reviewed and noticed two things were wrong.
1. The error was not properly handled - fixed
2. The STFC data source was not implementing the role update method - fixed

Thank you for spotting that!

apps/backend/db_patches/0208_AddTagsToRoles.sql
jekabs-karklins: Done

apps/frontend/src/components/admin/RoleModal.tsx
mutambaraf: Now having validation error because of ```@ValidateArgs(upsertUserByOidcSubValidationSchema)```  decorator  ``` "error":"{\"userMessage\":\"Input validation errors\",\"error\":{\"response\":{\"errors\":[{\"message\":\"Input validation errors\",\"locations\":[{\"line\":2,\"column\":3}],\"path\":[\"createRole\"],\"extensions\":{\"code\":\"INTERNAL_SERVER_ERROR\",\"context\":{\"errors\":[\"firstName is a required field\",\"lastName is a required field\",\"oidcSub is a required field\",\"Exactly one of rorId or institutionData must be provided\",\"email is a required field\"],\"inputArgs\":{\"shortCode\":\"proposal_reader\",\"title\":\"Test\",\"description\":\"Test\",\"permissions\":[]}},\"stacktrace\":[\"GraphQLError: Input validation errors\",\"    at new Rejection (/app/backend/src/models/Rejection.ts:33:5)\",\"    at rejection (/app/backend/src/models/Rejection.ts:65:10)\",\"    at UserMutations.<anonymous> (/app/backend/src/decorators/ValidateArgs.ts:54:25) ```

jekabs-karklins: Hi, @mutambaraf ,

Can you try running 
`npm i` from the root of the project?

I think this error comes from another PR that was merged in yesterday

mutambaraf: We do not have some of the values being validated like `oidcSub` and   also we are just creating a role but looks like we are checking for user details.

apps/backend/src/mutations/UserMutations.ts
```diff
@@ -365,6 +367,48 @@ export default class UserMutations {
   }
 
   @ValidateArgs(upsertUserByOidcSubValidationSchema)
```
mutambaraf, line 367: I think this validation might be at wrong end point.

jekabs-karklins: Hi @mutambaraf , thank you for bearing with me.

I see where the problem ocurred. I tested yestterdat that role creation worked and pushed the fix. 
Then pulled from develop, which did automatic merge and   @ValidateArgs(upsertUserByOidcSubValidationSchema) ended up above wrong mutation 🤦🏻 . Since there were no tests the build was green.

I have addressed the issue and added some unit tests. 

Thank you for the patience.

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
mutambaraf, line 1: We are not implementing `setUserRoles` in this file , since this is not behind a feature flag can this be move to the default implementation and maybe override it if necessary.

mutambaraf: Or do as we have done for ```  async deleteRole(id: number): Promise<Role> {
    return postgresUserDataSource.deleteRole(id);
  }```

jekabs-karklins: Hi @mutambaraf ,
I have addressed the issue by moving the role realted methods `createRole`, `updateRole` and `deleteRole` to the RoleDataSource, where I think they fit in better and this way we also don't need to change the `apps/backend/src/datasources/stfc/StfcUserDataSource.ts` at all.

jekabs-karklins:
@mutambaraf , Do you have any more comments on this PR, or would you like to discuss this further?

This PR essentially enables us to have a much needed PROPOSAL_READER role.
Additionally we can extend this to tag USER role  which would allow different user segragate groups to allow them to access different instruments and calls.

mutambaraf:
> @mutambaraf , Do you have any more comments on this PR, or would you like to discuss this further?
> 
> This PR essentially enables us to have a much needed PROPOSAL_READER role. Additionally we can extend this to tag USER role which would allow different user segragate groups to allow them to access different instruments and calls.

I think we might need to do some more changes for it to work for STFC .Let me do a last check that we are not breaking anything else and we can get it approved.

mutambaraf (approved):

PR #1294, opened 2025-12-16
fix: call order is not persistent for user officers

ellen-wright:
## Description
refs: https://github.com/UserOfficeProject/issue-tracker/issues/1473
Fixes the bug where the order selected by User Officers for the calls is not persistent when they close the order editor and reopen. 

When the calls are required to be ordered the ordering is done in the backend. 
A ordered flag is added to calls filter. 

## Motivation and Context
This was a bug 


apps/backend/src/resolvers/queries/CallsQuery.ts
```diff
@@ -58,6 +58,9 @@ export class CallsFilter {
 
   @Field(() => Boolean, { nullable: true })
   public isCallUpcoming?: boolean;
+
```
jekabs-karklins, line 61: Question: I wonder if we ever have scenarion where we would *not* want calls to be sorted? would not that be simpler if we always sort by the sort order?

ellen-wright: I guess my thinking is in the calls table the initial order would be interrupted

apps/frontend/src/components/proposal/ProposalChooseCall.tsx
```diff
@@ -30,27 +30,26 @@ const ProposalChooseCall = () => {
           isActive: true,
           isEnded: false,
           isActiveInternal: true,
+          isOrdered: true,
         }
       : {
           isActive: true,
           isEnded: false,
+          isOrdered: true,
         };
```
jekabs-karklins:
```suggestion
return {
  isActive: true,
  isEnded: false,
  isActiveInternal: isInternalUser ? true : undefined,
            isOrdered: true,
  };
```

Since we are adding more params I think it makes sense to consolidate this bit.

jekabs-karklins (commented):
Just one more suggestion I just noticed WDYT. Otherwise looks good!

jekabs-karklins (approved):

Scott-James-Hurley (approved):
Maybe a couple of e2e tests should be added to check if the order is persisted on the call ordering and when creating a new proposals?

PR #1297, opened 2026-01-06
feat: workflow overhaul

jekabs-karklins:
## Description
This PR 
- introduces new database structures to improve the workflow management system
- extends functionality by making workflows more capable
- improves UI to enable new capabilties and align with material design guidelines

## Motivation and Context
The current workflow management system has limitations and inefficiencies that are addressed by the introduction of new database structures. This change is required to enhance the system's performance and user experience.

### State machine
This workflow overhaul aims to align the state changes of proposal to **classical state machine behaviour.** The implementation of abstract state  machine can be found in `stateMachnine.ts`. It matches the popular `xState` library interface just is much more simple. 
image
There are 5 main concepts in the state machine.
- Status (State): A specific, stable stage in a proposal's lifecycle (e.g., DRAFT). The proposal stays in this state until a transition is made.
- Event: A signal or trigger that initiates a potential transition (e.g., PROPOSAL_SUBMITTED). It informs the state machine engine that an action has occurred and a state change may be required (guard might abort the transition).
- Guard: A validation or condition (e.g., isProposalInstrumentSelected) that must return true for the transition to proceed. It acts as a gatekeeper to ensure business rules are met.
- Action (Side-effect): An automated task executed during the transition itself (e.g., notifying RabbitMQ). These are the reactive behaviors that happen as a result of the state change.
- Transition: The process of moving from one status (source) to another (target). A transition is triggered by an Event, must clear any Guards, and may execute one or more Actions as it resolves.

### ER diagram of the new Workflow Data Structure
Untitled Diagram-1754399401220-Page-1 drawio

> Note: in order to keep the PR smaller I have not moved some tables into the DB but kept them in the code (color gray), but it makes sence to depict them here in the ER diagram.

Entity Tables:
- workflows: contains the main record for workflows containing name, description and workflow_type
- statuses: contains all statuses that the entity (proposal, exp.safety) can be in
- proposals: the propoals table
- workflow_status_connections: defines the transitions (see `Transition` under "State machine" section)
- workflow_status_changing_events: defines the events (see `Event` under "State machine" section)
- workflow_status_actions: defines the actions/side-effects (see `Action` under "State machine" section)
- workflow_status_changing_guards: defines the guards effects (see `Guard` under "State machine" section)

## Changes
This PR removes a lot of unneeded code in a way simplifying the implementation

### Here are the main changes to pay attention when reviewing
1. Cleanup of the namings DefaultStatus and InitialStatus. 
`Initial status` is the status of the entity when it enters the workflow. `DRAFT` for proposal and `AWAITING_ESF` for safetyReview. 
`Defaultstatus` is status which system is aware of and is not created by user with through "Create status" button. These two namings were mixed and used interchangebly, but now they are separated.

2. The table `proposal_events` and `experiment_safety_events` has been dropped and their alternative implementation has been implemented in `guards` which now checks if the condition is true during the transition. (see apps/backend/src/workflowEngine/guards/*.ts)

2.1 There might be an important change for STFC. Since I have removed the `proposal_events` table and along with it  this `proposalEvents.proposal_all_fap_reviews_submitted`. Since this is not available any more I have changed to check if all fap reviews are submited by inspecting the fap review table columns https://github.com/UserOfficeProject/user-office-core/pull/1297/changes#diff-f3b982a00651344db8c6516a09c950e0e30df0ac84e1cd543826a0d45875588eR209

3. `Status` primary key now uses the semantic string statusId (formerly short_code), dropping the surrogate int. This removes number => string mapping in the code, and keeps IDs readable in logs/DB dumps and tables that reference status. This prevents mismatch bugs where int keys and short codes drifted. Migration 0205_RenameShortCodeToStatusId.sql rewires FK references across proposals/experiment_safety/workflow tables and reconstitutes dependent views. 
Operationally this makes debugging and API payloads self-describing and aligns status identity with the business identifier.

4. Status actions are now attached on transition. Before they were attached the status. This means, that if you change the status manually, the status action will not be run.

## How Has This Been Tested?

Added and updated e2e tests

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-4949](https://jira.esss.lu.se/browse/SWAP-4949)


ACLay:
What is this PR actually trying to do? The body mentions "limitations and inefficiencies" but it seems so vague I have no idea, and I don't have access to the linked ESS jira item. If this is for the "allowing a status to be on a workflow multiple times" thing mentioned at sprint planning, I thought that was going to go to a discussion thread so we could figure out what the approach to take should be?

And from a review standpoint, I don't see how I'm practically going to be able to review 10k lines of code changes.

jekabs-karklins:
> What is this PR actually trying to do? The body mentions "limitations and inefficiencies" but it seems so vague I have no idea, and I don't have access to the linked ESS jira item. If this is for the "allowing a status to be on a workflow multiple times" thing mentioned at sprint planning, I thought that was going to go to a discussion thread so we could figure out what the approach to take should be?
> 
> And from a review standpoint, I don't see how I'm practically going to be able to review 10k lines of code changes.

Hi @ACLay , 

All valid points, and I have therefore not marked this as review:please just yet. 

Side note: When started on this overhaul I did not realize that workflow touches so many things in the application, and this turned out the be the biggest PR I have made. 
This PR is a bit of a special case and I was hoping we could set up the meeting as there are number of important things to talk about what this PR affects, as well get feedback from you guys on the topic.

Meanwhile I will try to improve the description of the PR to give high level overview.

Thanks for the proactive comment

yoganandaness:
@TCMeldrum Kindly review and ensure, it does not break your merged PR - https://github.com/UserOfficeProject/user-office-core/pull/1355

jekabs-karklins:
@ACLay I have added the ability to specify status actions when User Officer sets the status manually in this PR
https://github.com/UserOfficeProject/user-office-core/pull/1441

We can merge the #1441after we get this one in. Please have a look at your earlies convenience. 

Thanks.

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
```diff
@@ -0,0 +1,192 @@
...
+              ON DELETE CASCADE
+          );
+
+          -- (2) Prevent duplicate edges within a workflow.
+          ALTER TABLE workflow_status_connections
+            ADD CONSTRAINT uq_wsc_edge UNIQUE (workflow_id, prev_workflow_status_id, next_workflow_status_id);
+
+
+
+          -- ============================================
```
ACLay: It seems weird to add an empty block referencing something that doesn't seem to exist anywhere in the code

apps/backend/db_patches/db_seeds/0001_ProposalForScheduling.sql
```diff
@@ -7,14 +7,30 @@ DECLARE
...
+    (1, 10, 390, 'NOT_ALLOCATED'),
+    (1, 10, 450, 'SCHEDULING'),
+    (1, 10, 520, 'EXPIRED'),
+    (1, 10, 580, 'EDITABLE_SUBMITTED_INTERNAL'),
+    (1, 10, 640, 'EDITABLE_SUBMITTED');
+
   INSERT INTO instruments (instrument_id, name, short_code, description, manager_user_id) VALUES (1, 'Instrument 1', 'INSTR1', 'Test instrument 1', 0);
   INSERT INTO instruments (instrument_id, name, short_code, description, manager_user_id) VALUES (2, 'Instrument 2', 'INSTR2', 'Test instrument 2', 0);
   INSERT INTO instruments (instrument_id, name, short_code, description, manager_user_id) VALUES (3, 'Instrument 3', 'INSTR3', 'Test instrument 3', 0);
+  PERFORM setval('public.instruments_instrument_id_seq', 4, false); -- Reset sequence to avoid conflicts
```
ACLay: What are these sequence manipulation calls for? It seems odd that we'd need to go back and alter the seed data for some table's sequences when:

- the PR doesn't seem to alter anything about these tables
- this sort of sequence manipulation isn't applied in any other db seed files

If we do need to change things here, would it be better to not explicitly setting the ID and let the sequence generate the id normally? That seems to be how it's done elsewhere in these seed files. These PERFORM setval calls seem inconsitent.

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
```diff
@@ -0,0 +1,192 @@
...
+          -- REMOVED: workflow_status_changing_events table is removed.
+          -- Events are now stored as strings in the code (apps/backend/src/events/event.enum.ts).
+
+
+
+
+          -- =====================================================================
+          -- 4) workflow_status_connection_has_workflow_status_changing_events (edge→events)
+          -- =====================================================================
+          CREATE TABLE workflow_status_connection_has_workflow_status_changing_events (
```
ACLay: This table name is a bit of a mouthful. Would it still work if it simplified to something shorter and closer to the acronymn used in the primary/foreign keys like `workflow_status_connection_has_events` ?

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
```diff
@@ -0,0 +1,192 @@
...
+
+          -- Rename existing status_actions table to workflow_status_actions
+          ALTER TABLE status_actions RENAME TO workflow_status_actions;
+          ALTER table workflow_status_actions
+            RENAME COLUMN status_action_id TO workflow_status_action_id;
+
+          -- ==========================================================
+          -- 5) workflow_status_connection_has_workflow_status_actions (edge→actions)
+          -- ==========================================================
+          CREATE TABLE workflow_status_connection_has_workflow_status_actions (
```
ACLay: Similarly, could this name simplify to `workflow_status_connection_has_actions` ?

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
```diff
@@ -0,0 +1,192 @@
...
+
+          ALTER TABLE experiment_safety
+            ADD CONSTRAINT fk_experiment_safety_workflow_status
+            FOREIGN KEY (workflow_status_id)
+            REFERENCES workflow_has_statuses (workflow_status_id);
+
+          -- ==================================================================
+          -- 8) Sync status_id from workflow_status_id
+          -- ==================================================================
+          CREATE OR REPLACE FUNCTION set_status_from_workflow_status()
```
ACLay: This looks like we're introducing a reference loop where a row in proposals or experiment_safety refers to both a status and a workflow_status, and we want to make sure that we're pointing at the same status that the workflow_status points at.
Maybe I'm missing something, but I think it would be a simpler, more normalised database setup to get rid of the status reference from proposals & experiment_safety, and when we need the status go through workflow_status.

apps/backend/db_patches/0208_MigrateOldWorkflowsToNewDataStructures.sql
```diff
@@ -0,0 +1,314 @@
...
+        IF FOUND THEN
+          RAISE EXCEPTION 'workflow_has_statuses already contains rows. Migration must run on an empty target table.';
+        END IF;
+
+        PERFORM 1 FROM workflow_status_connections LIMIT 1;
+        IF FOUND THEN
+          RAISE EXCEPTION 'workflow_status_connections already contains rows. Migration must run on an empty target table.';
+        END IF;
+
+        PERFORM 1
```
ACLay: Why are we including guards in this script to check that the tables and columns exist? I don't think this is something we've ever done before, and it feels unnecessary, as wouldn't the entire system have fallen apart in errors if these didn't exist?

apps/backend/db_patches/0208_MigrateOldWorkflowsToNewDataStructures.sql
```diff
@@ -0,0 +1,314 @@
...
+          ON conn.workflow_connection_id = action.connection_id;
+
+        SELECT COUNT(*) INTO v_unmapped_actions
+        FROM workflow_connection_has_actions action
+        LEFT JOIN tmp_workflow_edge_map edge_map
+          ON edge_map.workflow_connection_id = action.connection_id
+        WHERE edge_map.workflow_connection_id IS NULL;
+
+        IF v_unmapped_actions > 0 THEN
+          RAISE WARNING USING MESSAGE = format('%s workflow actions could not be migrated because their connection does not reference a previous node.', v_unmapped_actions);
```
ACLay, line 175: I'm not sure what would be appropriate way to handle this. An exception mid-migration might be problematic for resuming, but doing it at warning feels like if a facilities workflow gets broken mid-migration it might not get noticed until much later mid-call when proposals start getting into incorrect states or failing to move to the one we'd expect them to.

apps/backend/db_patches/0208_MigrateOldWorkflowsToNewDataStructures.sql
```diff
@@ -0,0 +1,314 @@
...
+        IF call_has_workflow_id THEN
+          INSERT INTO tmp_call_workflow_map (call_id, workflow_id)
+          SELECT call_id, workflow_id
+          FROM call
+          WHERE workflow_id IS NOT NULL
+          ON CONFLICT (call_id) DO UPDATE
+            SET workflow_id = COALESCE(EXCLUDED.workflow_id, tmp_call_workflow_map.workflow_id);
+        END IF;
+
+        CREATE TEMP TABLE tmp_workflow_status_lookup (
```
ACLay, line 243: This file seems to be making use of a lot of temporary tables in loops. It doesn't seem typical for this repo (I can't see any other temp table usages), and I was under the impression that looping in SQL is pretty inefficient compared to single statements where you insert the entire output of a select.

ACLay (commented):
I've looked over some of the SQL so far, and it feels like there's a lot of deviation from how we've done things before in the repo. That combined with it being an extremely long PR, I think this could take quite a while to review.

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
jekabs-karklins: Removed!

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
jekabs-karklins: good idea. I will still keep workflow_.. in front of it to group the related tables together

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
jekabs-karklins: done

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
jekabs-karklins: I created the trigger that updates `status` from `workflow_status` whenever `workflow_status` is updated. This information, as you properly say, is duplicated. I kept `status` field to reduce the changes in this PR, as quiet a few places were using it.

I just checked that it is actually only about 10 places so it's doable within this PR. Do you think if I derive the status in `proposal_view` and use `proposal_view.status` where it currently uses `proposal.status` is a good idea?

apps/backend/db_patches/0208_MigrateOldWorkflowsToNewDataStructures.sql
jekabs-karklins: This is a sanity check, and in fact should never happen, if you would like we can raise an `EXCEPTION` instead of `WARNING`.

What is it checking here is if the edge has a parent. Since in the old workflow we were attaching `actions` to the Node, not the edge, in theory we could have some status action defined on initial state like `DRAFT`. But the UI should have prevented it. I think this is fine to keep as it is more of a sanity check then something that we expect to happen.

apps/backend/db_patches/0208_MigrateOldWorkflowsToNewDataStructures.sql
jekabs-karklins: I agree that this should not happen, but it is also a part of defensive programming. 

But, should we clean this up anyway, i have no strong oppinion here?

apps/backend/db_patches/db_seeds/0001_ProposalForScheduling.sql
jekabs-karklins: I had to change them because they were not well designed and there is an issue with sequence synchronization after manual ID inserts.

when we insert instruments with explicit IDs 1, 2, 3. with IDs are provided manually, PostgreSQL sequence state does not reliably advance to match those values.

```sql
duo=$ select * from instruments;
 instrument_id | name | short_code | description | manager_user_id | selectable | multiple_tech_reviews_enabled 
---------------+------+------------+-------------+-----------------+------------+-------------------------------
(0 rows)

duo=$   INSERT INTO instruments (instrument_id, name, short_code, description, manager_user_id) VALUES (1, 'Instrument 1', 'INSTR1', 'Test instrument 1', 0);
  INSERT INTO instruments (instrument_id, name, short_code, description, manager_user_id) VALUES (2, 'Instrument 2', 'INSTR2', 'Test instrument 2', 0);
  INSERT INTO instruments (instrument_id, name, short_code, description, manager_user_id) VALUES (3, 'Instrument 3', 'INSTR3', 'Test instrument 3', 0);
INSERT 0 1
INSERT 0 1
INSERT 0 1

--- now insert using the auto sequence
duo=$   INSERT INTO instruments (name, short_code, description, manager_user_id) VALUES ('without id', 'WOID', 'Without id', 0);
ERROR:  duplicate key value violates unique constraint "instruments_pkey"
DETAIL:  Key (instrument_id)=(1) already exists.

```


Alternatively we could delete the `id` from the insert statements.

apps/backend/db_patches/0208_MigrateOldWorkflowsToNewDataStructures.sql
jekabs-karklins: You’re right that it’s inefficient, but for a one-time migration, I don't think it's a major concern. We'll only run it once, and the temporary tables will be cleaned up upon the transaction COMMIT

jekabs-karklins:
@TCMeldrum I have added the pre-flight check that runs before migration starts to make sure the old data is compatible with the new data stucture and prints out what exactly is wrong if the requirement is not met.
Check the changes here
https://github.com/UserOfficeProject/user-office-core/pull/1297/changes/ad9709dd2feab234bc227954b1b9a30f155cf43b

TCMeldrum:
> @TCMeldrum I have added the pre-flight check that runs before migration starts to make sure the old data is compatible with the new data stucture and prints out what exactly is wrong if the requirement is not met. Check the changes here [ad9709d](https://github.com/UserOfficeProject/user-office-core/commit/ad9709dd2feab234bc227954b1b9a30f155cf43b)

I have had a look and that work to tag all our workflows that need updating thanks for adding that. I checked with emma and being only able to set proposals to statuses that are on the workflow should be fine

@TCMeldrum Kindly review and ensure, it does not break your merged PR - https://github.com/UserOfficeProject/user-office-core/pull/1355

Had a look at permissions for the fap review visiblity and it looked fine thanks! 

In terms of reviewing this PR I am a bit lost, trying to understand 10000 lines of changes in one go seems a bit overwhelming. Is it worth having a meeting were we can go over all the changes with more people as this seems to a fundermental change of how the system works.

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
ACLay: I definitely think it'd be better to avoid having data duplicated across columns where it could potentially fall out of sync.

apps/backend/db_patches/0209_RenameShortCodeToStatusId.sql
```diff
@@ -0,0 +1,343 @@
+DO
+$$
+DECLARE
+    -- Variable to hold count of affected rows for logging/verification
+    v_count bigint;
+BEGIN
+    IF register_patch(
+       '0205_RenameShortCodeToStatusId',
```
ACLay: This doesn't line up with the number in the filename

apps/backend/src/datasources/mockups/ProposalDataSource.ts
```diff
@@ -258,7 +259,7 @@ export class ProposalDataSourceMock implements ProposalDataSource {
     return proposal;
   }
 
-  async updateProposalStatus(
+  async updateProposalWfStatus(
```
ACLay: updateProposalWfStatus seems named very similar, and look functionally similar to changeProposalsWorkflowStatus. Do we need both? If so, can the name changes be made consistent? My preference would be the full "Workflow" be in the name instead of the abbreviated "Wf"

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -627,23 +628,44 @@ export interface WorkflowRecord {
...
+  readonly prev_workflow_status_id: number;
+  readonly next_workflow_status_id: number;
+  readonly source_handle: string;
+  readonly target_handle: string;
+}
+
+// NOTE: This is a data structure returned by getWorkflowStructure function in WorkflowDataSource,
+// and contains full information about workflow statuses and connections,
+// so we do not need to make multiple calls to database to get workflow structure
+export interface WorkflowStructure {
```
ACLay, line 606: All the other interfaces here seem to mark their fields as readonly. Why doesn't this one?

apps/backend/src/datasources/postgres/WorkflowDataSource.ts
```diff
@@ -162,6 +148,16 @@ export default class PostgresWorkflowDataSource implements WorkflowDataSource {
         return this.createWorkflowObject(records[0]);
       });
   }
+
+  async updateWorkflowTimestamp(workflowId: number): Promise<void> {
```
ACLay, line 153: I don't know if there is an ideal solution, as triggers could fall out of date or be missed if addiational tables are added, but this also feels prone to being forgotten if we add other methods that update a workflow.

ACLay (commented):
Just to echo my [reply from yesterday](https://github.com/UserOfficeProject/user-office-core/pull/1297#discussion_r3059393468), I do think we should avoid duplicating the status field so that we aren't storing the value in multiple places that could fall out of sync

jekabs-karklins:
> > @TCMeldrum I have added the pre-flight check that runs before migration starts to make sure the old data is compatible with the new data stucture and prints out what exactly is wrong if the requirement is not met. Check the changes here [ad9709d](https://github.com/UserOfficeProject/user-office-core/commit/ad9709dd2feab234bc227954b1b9a30f155cf43b)
> 
> I have had a look and that work to tag all our workflows that need updating thanks for adding that. I checked with emma and being only able to set proposals to statuses that are on the workflow should be fine
> 
> @TCMeldrum Kindly review and ensure, it does not break your merged PR - #1355
> 
> Had a look at permissions for the fap review visiblity and it looked fine thanks!
> 
> In terms of reviewing this PR I am a bit lost, trying to understand 10000 lines of changes in one go seems a bit overwhelming. Is it worth having a meeting were we can go over all the changes with more people as this seems to a fundermental change of how the system works.

We have had that meeting on 24.march. If I remember correctly you might not have been in the office, I still have slides and we can go over them again, if you preffer, 20, or 21 Apr when I'm back. Let me know and I'll book a meeting 👍🏻

apps/backend/db_patches/db_seeds/0001_ProposalForScheduling.sql
jekabs-karklins: I have deleted the ids as per your request

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
jekabs-karklins: Just to confim on this. If the main concern is fallint out of sync then as the check has been added to the db trigger level this is unlikely to go out of sync ans whennever wf connection id is updated so is status_id.

If the main concern is inconsistency, then I can update the 10 places within this PR. Let me know what you think.

TCMeldrum:
> We have had that meeting on 24.march. If I remember correctly you might not have been in the office, I still have slides and we can go over them again, if you preffer, 20, or 21 Apr when I'm back. Let me know and I'll book a meeting 👍🏻

@jekabs-karklins Cool either way  works for me shoot me a message on teams either date works for me

apps/backend/src/datasources/WorkflowDataSource.ts
```diff
@@ -1,53 +1,56 @@
...
-      WorkflowConnection,
-      'id' | 'entityType' | 'prevConnectionId'
-    >
-  ): Promise<WorkflowConnectionWithStatus>;
+  ): Promise<WorkflowConnection[]>;
+  deleteWorkflowConnection(
+    connectionId: number
+  ): Promise<WorkflowConnection | null>;
+
+  addStatusToWorkflow(newWorkflowStatusInput: {
```
ACLay: Can this use the `AddStatusToWorkflowInput` type like instead of redefining it?

apps/backend/src/eventHandlers/proposalWorkflow.ts
```diff
@@ -45,21 +42,21 @@ const publishProposalStatusChange = async (
         isRejection: false,
         key: 'proposal',
         loggedInUserId: null,
-        description: `From "${previousProposalStatus?.name}" to "${proposalStatus?.name}"`,
```
ACLay, line 48: I believe the aim was for statusId to contain the old name values, but I wonder if it'd be better, or at least more readable, if we kept the old `name` name, and used it as an id field/column without having it be called `statusId`?

ACLay (changes requested):

apps/backend/db_patches/0209_RenameShortCodeToStatusId.sql
jekabs-karklins: Thanks!

apps/backend/src/datasources/WorkflowDataSource.ts
jekabs-karklins: Of course 💯

apps/backend/src/datasources/mockups/ProposalDataSource.ts
jekabs-karklins: Excellent catch. This will also make sure proposal marked as 'not submitted'.

apps/backend/src/eventHandlers/proposalWorkflow.ts
jekabs-karklins: Sorry, I don't follow, which `statusId` and `name` are you reffering to. Do you mean the `updatedProposal.statusId` in the eventBus message? Could you add also a little more context on the problem you are reffering to. Thank you.

apps/backend/db_patches/0208_MigrateOldWorkflowsToNewDataStructures.sql
jekabs-karklins: I have removed the checks that check if the database is in the expected state (i.e. in the state the previous db patch left it at). We should safely assume that it is.

apps/backend/src/datasources/postgres/WorkflowDataSource.ts
jekabs-karklins: I have solely kept update of timestamp in the boundaries of this file. Unless we add a DB trigger, which is also not optimal as it works like "magic", I think this is the best we can do here. Benefit of the cache is quiet significant here.

apps/backend/src/datasources/postgres/records.ts
jekabs-karklins: Done Thanks for spotting.

apps/backend/src/queries/FapQueries.ts
```diff
@@ -252,4 +255,46 @@ export default class FapQueries {
...
+    fapId: number,
+    proposalPk: number,
+    isApiToken: boolean,
+    isUserOfficer: boolean,
+    isChairOrSecretary: boolean
+  ) {
+    const canSeeAllAssignments =
+      isApiToken || isUserOfficer || isChairOrSecretary;
+
+    if (canSeeAllAssignments) {
```
ACLay: The method and this check don't seem to line up to me. The latter code and name seem to be asking if a proposal has been reviewed, but this seems to be treating it as completely reviewed based on a users role.

apps/backend/src/queries/WorkflowQueries.ts
```diff
@@ -40,10 +40,25 @@ export default class WorkflowQueries {
...
+
+  @Authorized()
+  async getWorkflowStatus(
+    agent: UserWithRole | null,
+    workflowStatusId: number
+  ) {
+    const status = await this.dataSource.getWorkflowStatus(workflowStatusId);
+
+    return status;
+  }
```
ACLay: If we're not doing anything with the response, can we simplify it and make it more consistent to the preceeding methods?
```suggestion
    return this.dataSource.getWorkflowStatus(workflowStatusId);
  }
```

jekabs-karklins: Sure

apps/backend/src/queries/FapQueries.ts
jekabs-karklins: hi, yes I can see this has been one of the latter adjustments. Although it worked, as you say it didn't quiet ligned up. I have moved the role checks outside the method to before it is called

jekabs-karklins:
> > We have had that meeting on 24.march. If I remember correctly you might not have been in the office, I still have slides and we can go over them again, if you preffer, 20, or 21 Apr when I'm back. Let me know and I'll book a meeting 👍🏻
> 
> @jekabs-karklins Cool either way works for me shoot me a message on teams either date works for me

Hi, @TCMeldrum 
Can I propose next Tuesday, 28. Apr? Let me know which time works for you best. I'm busy until 11.00.

jekabs-karklins:
> Just to echo my [reply from yesterday](https://github.com/UserOfficeProject/user-office-core/pull/1297#discussion_r3059393468), I do think we should avoid duplicating the status field so that we aren't storing the value in multiple places that could fall out of sync

Hi @ACLay . I have now normalized te database and removed the column `status_id` and solely relying in workflow_status_id, this way there is no duplicationl

apps/backend/src/queries/FapQueries.ts
```diff
@@ -178,42 +166,58 @@ export default class FapQueries {
...
+        const isAllFapReviewsComplete = await this.allFapReviewsComplete(
+          fapId,
+          proposalPk
+        );
+        const canSeeAllAssignments =
+          isApiToken ||
+          isUserOfficer ||
+          isChairOrSecretary ||
+          isAllFapReviewsComplete;
+
```
ACLay: The allFapReviewsComplete method has to call out to an additional datasource, but we only use that value if none of the other known values are true. We should be able to speed up the function if we only perform that check when needed
```suggestion
        const canSeeAllAssignments =
          isApiToken ||
          isUserOfficer ||
          isChairOrSecretary ||
          (await this.allFapReviewsComplete(fapId, proposalPk));

```

apps/backend/src/workflowEngine/guards/isProposalAllFapMeetingInstrumentSubmittedGuard.ts
```diff
@@ -0,0 +1,19 @@
+import { container } from 'tsyringe';
+
+import { Tokens } from '../../config/Tokens';
+import { FapDataSource } from '../../datasources/FapDataSource';
+import { Entity, GuardFn } from '../simpleStateMachine/stateMachnine';
+
+export const isProposalAllFapMeetingInstrumentSubmittedGuard: GuardFn = async (
```
ACLay: Some of these guard names, especially the isProposalAll... ones, seem rather tricky to parse and understand. I think it'd be more readable if we renamed them. For instance, based on the code I think isProposalAllFapMeetingInstrumentSubmittedGuard would be more better as isEveryInstrumentSubmittedForProposalAtFapGuard

apps/backend/src/workflowEngine/simpleStateMachine/createWorkflowMachine.ts
```diff
@@ -0,0 +1,87 @@
+import { container } from 'tsyringe';
+
+import { createMachine, GuardFn, StateConfig } from './stateMachnine';
+import { Tokens } from '../../config/Tokens';
+import { WorkflowDataSource } from '../../datasources/WorkflowDataSource';
+import { Event, EventMetadataByEvent } from '../../events/event.enum';
+
+const createWfStatusName = (statusId: string, workflowStatusId: number) =>
```
ACLay: In general, could we avoid using the Wf abbreviation? It feels inconsistent to see names with Wf next to ones with the full Workflow, and it feels less readable.

ACLay (changes requested):
Given that we only seem to have the one state machine, and it seems designed to cover all our types of workflow, the "simple" in it's directory name seems unnecessary. Can we rename "simpleStateMachine" to just "stateMachine"?

apps/backend/src/queries/FapQueries.ts
jekabs-karklins: Thanks, that makes sense

apps/backend/src/workflowEngine/experiment.ts
```diff
@@ -1,318 +1,179 @@
...
+      )
     );
-    const newStatusChangingEvents =
-      await workflowDataSource.getStatusChangingEventsByConnectionIds(
-        nextNextWorkflowConnections.map((connection) => connection.id)
-      );
 
-    if (!experimentSafetyWithEvents.experimentSafetyEvents) {
+    const validExperiments = experimentsWithChangedStatuses.filter(
+      (p): p is WorkflowEngineExperimentType => !!p
```
ACLay: Guessing it's a holdover from starting as a copy of proposal.ts, but I don't think `p` is the right name for this.

apps/e2e/cypress/e2e/instruments.cy.ts
```diff
@@ -201,6 +214,18 @@ context('Instrument tests', () => {
...
+        id: 'FEASIBILITY',
+        name: 'Feasibility',
+        description: 'Feasibility status',
+        entityType: WorkflowType.PROPOSAL,
+      });
+
+      cy.addStatusToWorkflow({
+        workflowId: initialDBData.workflows.defaultWorkflow.id,
+        statusId: 'FEASIBILITY',
+      });
```
ACLay: We're adding this FEASIBILITY setup to beforeEach in every one of the three `describe` blocks. Is it something we can do in the top level beforeEach instead to reduce duplication?

apps/e2e/cypress/e2e/settings.cy.ts
```diff
@@ -1008,7 +989,7 @@ context('Settings tests', () => {
       cy.contains('FAP_REVIEW');
     });
 
-    it('Proposal status should update multiple times if conditions are met', () => {
+    it.skip('Proposal status should update multiple times if conditions are met', () => {
```
ACLay: Do we want to be leaving this `.skip()` in?

ACLay (commented):
Read through to the end of /e2e

apps/backend/src/workflowEngine/guards/isProposalAllFapMeetingInstrumentSubmittedGuard.ts
jekabs-karklins: You are right. Updated! Let me know if any more are tricky to parse.

apps/e2e/cypress/e2e/settings.cy.ts
jekabs-karklins: This test should be deleted, because we do *not* want status to change multiple times if condition is met. This was one of the core problems with old workflows. In old workflows we could not have loops because of this behavior. If proposal would move to next status if conditions were met then the infinite loop could happen.

apps/backend/src/workflowEngine/experiment.ts
jekabs-karklins: Yes, thanks!

apps/backend/src/workflowEngine/simpleStateMachine/createWorkflowMachine.ts
jekabs-karklins: I have replaced the ocurrances I found, but let me know if you stumble upon any more

apps/e2e/cypress/e2e/settings.cy.ts
ACLay: @jekabs-karklins Just to check, was the "enable test" commit meant to delete the test?

apps/frontend/src/components/proposal/ChangeProposalStatus.tsx
```diff
@@ -4,60 +4,124 @@ import Container from '@mui/material/Container';
...
+      {} as Record<string, string[]>
+    );
+
+    return Object.entries(counts).map(([statusId, entities]) => ({
+      statusId,
+      entities,
+    }));
+  }, [selectedProposals, proposalStatuses]);
+
+  if (!allProposalsHaveSameWorkflow) {
```
ACLay, line 85: I think we can optimise this and move this check-and-return up to be done immediately after calculating `allProposalsHaveSameWorkflow`

apps/frontend/src/components/settings/workflow/WorkflowEditor.tsx
```diff
@@ -307,33 +183,24 @@ const WorkflowEditor = ({ entityType }: { entityType: WorkflowType }) => {
...
-
-      // Update target node (B) - set its prevStatusId to source (A) and prevConnectionId to source connection ID
-      dispatch({
-        type: EventType.WORKFLOW_STATUS_UPDATE_REQUESTED,
+        type: EventType.ADD_WORKFLOW_CONNECTION_REQUESTED,
         payload: {
-          connectionId: targetConnection.id, // Use connection ID for persistence
-          prevStatusId: sourceConnection.statusId,
-          prevConnectionId: sourceConnection.id, // Store the previous connection ID
+          sourceWorkflowStatusId: sourceWfStatus.workflowStatusId, // Use connection ID for persistence
```
ACLay: This comment doesn't quite match the line now

apps/frontend/src/components/settings/workflow/workflowUtils.ts
```diff
@@ -0,0 +1,117 @@
...
+  highlightedNodes?: { statusId: string; entities: string[] }[],
+  isReadOnly?: boolean
+): { nodes: Node[]; edges: Edge<EdgeData>[] } => {
+  const newNodes: Node[] = [];
+  const newEdges: Edge<EdgeData>[] = [];
+
+  state.statuses.forEach((workflowStatus) => {
+    const statusId = workflowStatus.status.id.toString();
+    const nodeId = workflowStatus.workflowStatusId.toString();
+    // Use database coordinates if available, otherwise fall back to grid layout
```
ACLay: What's the fallback that this mentions? It looks like it only uses the posX/posY values from the database.

apps/frontend/src/components/proposal/ChangeProposalStatus.tsx
jekabs-karklins: I could move the `selectedProposalsWorkflowStatus` check but as per rules of hooks the useMemo() hook needs to be on top.

apps/frontend/src/components/settings/workflow/WorkflowEditor.tsx
jekabs-karklins: Thank you, I cleaned it up now 💪 🧹

apps/frontend/src/components/settings/workflow/workflowUtils.ts
jekabs-karklins: This is a leftover comment from earlier version. In fact pos_x and pos_y has the default value of 0. so there is no fallback.
(https://github.com/UserOfficeProject/user-office-core/blob/cc12911f14f10aeeea6fcc4c64848b8dd394d32d/apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql)
Thanks.

apps/e2e/cypress/e2e/instruments.cy.ts
jekabs-karklins: Done, thanks!

apps/e2e/cypress/e2e/settings.cy.ts
jekabs-karklins: @aclay I see the commit message does not really reflect the changes. Yes, I should have said *delete* test not *enable* test.

apps/backend/db_patches/0207_AddNewWorkflowDataStructures.sql
jekabs-karklins: This has now been solved as per your recommendation

simonfernandes:
I haven't reviewed the code but this looks fantastic 👍

I was just getting this running locally and there are some workflow changes that STFC would need to make, where we have proposals in statuses that the workflows don't contain:

```sql
-- Add SUBMITTED_LOCKED status to 'Import workflow'
INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
VALUES(5, 15, 1, 1, 0);

-- Add SUBMITTED_LOCKED status to 'Import workflow for Xpress'
INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
VALUES(92, 15, 1, 1, 0);

-- Add UNSUCCESSFUL status to 'Import workflow for Xpress'
INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
VALUES(92, 20, 1, 1, 0);

-- Add FINISHED status to 'Import workflow for Xpress'
INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
VALUES(92, 21, 1, 1, 0);

-- Add EXPIRED status to 'Import workflow'
INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
VALUES(5, 9, 1, 1, 0);

-- Add EXPIRED status to 'ISIS Rapid workflow'
INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
VALUES(4, 9, 1, 1, 0);

-- Add EXPIRED status to 'ISIS Xpress workflow'
INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
VALUES(91, 9, 1, 1, 0);
```

These statuses are all added to the workflows as unconnected and I don't see a problem with doing this.

Automatic status changes (e.g. on proposal submission) are working for Xpress/Technique proposals and regular proposals, and status actions are being executed.

But manual status changes are only partly working for Xpress/Technique proposals and regular proposals. The status is changing successfully but I'm seeing the below error and status actions are not being executed:

>[2026-05-06T13:03:36.187+01:00] ERROR - Unhandled NODE exception 
 {"exception":{"name":"Error","message":"Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from \"workflow_status_connections\" where \"workflow_status_connection_id\" = ? limit ?","stack":"Error: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from \"workflow_status_connections\" where \"workflow_status_connection_id\" = ? limit ?\n    at QueryCompiler_PG.toSQL (/app/backend/node_modules/knex/lib/query/querycompiler.js:112:13)\n    at QueryBuilder_PostgreSQL.toSQL (/app/backend/node_modules/knex/lib/query/querybuilder.js:84:44)\n    at ensureConnectionCallback (/app/backend/node_modules/knex/lib/execution/internal/ensure-connection-callback.js:4:30)\n    at Runner.ensureConnection (/app/backend/node_modules/knex/lib/execution/runner.js:318:20)\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at Runner.run (/app/backend/node_modules/knex/lib/execution/runner.js:30:19)"},"level":"error","message":"Unhandled NODE exception","trace_id":"835d6d47d64dbb656316046cb9748618","span_id":"c8c52cfd98351876","trace_flags":"01","service_name":"proposal-backend","timestamp":"2026-05-06T13:03:36.187+01:00"}
[ERROR] 13:03:36 Error: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from "workflow_status_connections" where "workflow_status_connection_id" = ? limit ?

apps/frontend/src/components/settings/workflow/WorkflowView.tsx
```diff
@@ -0,0 +1,92 @@
...
+  entityType,
+  highlightedNodes,
+  onNodeClicked,
+  selectedStatusId,
+}) => {
+  const reactFlowWrapper = useRef<HTMLDivElement>(null);
+  const [nodes, setNodes, onNodesChange] = useNodesState([]);
+  const [edges, setEdges, onEdgesChange] = useEdgesState([]);
+
+  const reducerMiddleware = () => {
```
ACLay: If all this middleware does is tell the framework to proceed to the next one, is it needed?

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -725,7 +739,9 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
               referenceNumbers,
               dateFilter,
               excludeProposalStatusIds:
-                currentRole === UserRole.INSTRUMENT_SCIENTIST ? [9] : [], // Hide expired from scientists
+                currentRole === UserRole.INSTRUMENT_SCIENTIST
+                  ? ['EXPIRED']
```
ACLay: Can this use StatusCode.EXPIRED instead of redefining the string?

ACLay (commented):
Okay, initial readthrough complete.

apps/backend/src/datasources/postgres/FileDataSource.ts
```diff
@@ -178,28 +177,28 @@ export default class PostgresFileDataSource implements FileDataSource {
...
-      const [transactionError] = await to(connection.query('BEGIN'));
-      if (transactionError) {
-        database.client.releaseConnection(connection);
+    const [transactionError] = await to(connection.query('BEGIN'));
+    if (transactionError) {
+      database.client.releaseConnection(connection);
 
-        return reject(`Could not begin transaction: ${transactionError}`);
-      }
+      throw new Error(`Could not begin transaction: ${transactionError}`);
```
ACLay, line 184: The first couple of `reject` calls were replaced with `throw new GraphQLError(...)`, but from here it looks like they're replaced with `throw new Error(...)`. Is that difference significant?

ACLay (commented):
Starting to look over the changes made [since I did my initial comments](https://github.com/UserOfficeProject/user-office-core/pull/1297/changes/79a250ff8d3d146fb0d7664da9506233f4c07b47..c44ce5dc31549eab91cdd324de3be5f8753c676d)

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -25,7 +25,7 @@ type StfcRolesToEssRole = { [key: string]: Roles[] };
 /*
  * Must not contain user role, this is appended at the very last step.
  */
-const stfcRolesToEssRoleDefinitions: StfcRolesToEssRole = {
+const stfcRolesToSystemRoleDefinitions: StfcRolesToEssRole = {
```
ACLay: If we're renaming the variable to use a more facility independent term for the roles, shouldn't we also do that renaming in the variable's type?

apps/backend/src/utils/stringStripHtml.ts
```diff
@@ -0,0 +1,171 @@
+export function stripHtml(input: string): string {
```
ACLay, line 1: This feels like we're writing our own xml/html parser, why do we need this?

ACLay (commented):
Okay, finished reading through all the changes since I started reviewing.

jekabs-karklins:
> Given that we only seem to have the one state machine, and it seems designed to cover all our types of workflow, the "simple" in it's directory name seems unnecessary. Can we rename "simpleStateMachine" to just "stateMachine"?

Yes, I think it makes sense. Thanks

jekabs-karklins:
> I haven't reviewed the code but this looks fantastic 👍
> 
> I was just getting this running locally and there are some workflow changes that STFC would need to make, where we have proposals in statuses that the workflows don't contain:
> 
> ```sql
> -- Add SUBMITTED_LOCKED status to 'Import workflow'
> INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> VALUES(5, 15, 1, 1, 0);
> 
> -- Add SUBMITTED_LOCKED status to 'Import workflow for Xpress'
> INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> VALUES(92, 15, 1, 1, 0);
> 
> -- Add UNSUCCESSFUL status to 'Import workflow for Xpress'
> INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> VALUES(92, 20, 1, 1, 0);
> 
> -- Add FINISHED status to 'Import workflow for Xpress'
> INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> VALUES(92, 21, 1, 1, 0);
> 
> -- Add EXPIRED status to 'Import workflow'
> INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> VALUES(5, 9, 1, 1, 0);
> 
> -- Add EXPIRED status to 'ISIS Rapid workflow'
> INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> VALUES(4, 9, 1, 1, 0);
> 
> -- Add EXPIRED status to 'ISIS Xpress workflow'
> INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> VALUES(91, 9, 1, 1, 0);
> ```
> 
> These statuses are all added to the workflows as unconnected and I don't see a problem with doing this.
> 
> Automatic status changes (e.g. on proposal submission) are working for Xpress/Technique proposals and regular proposals, and status actions are being executed.
> 
> But manual status changes are only partly working for Xpress/Technique proposals and regular proposals. The status is changing successfully but I'm seeing the below error and status actions are not being executed:
> 
> > [2026-05-06T13:03:36.187+01:00] ERROR - Unhandled NODE exception
> > {"exception":{"name":"Error","message":"Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from "workflow_status_connections" where "workflow_status_connection_id" = ? limit ?","stack":"Error: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from "workflow_status_connections" where "workflow_status_connection_id" = ? limit ?\n    at QueryCompiler_PG.toSQL (/app/backend/node_modules/knex/lib/query/querycompiler.js:112:13)\n    at QueryBuilder_PostgreSQL.toSQL (/app/backend/node_modules/knex/lib/query/querybuilder.js:84:44)\n    at ensureConnectionCallback (/app/backend/node_modules/knex/lib/execution/internal/ensure-connection-callback.js:4:30)\n    at Runner.ensureConnection (/app/backend/node_modules/knex/lib/execution/runner.js:318:20)\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at Runner.run (/app/backend/node_modules/knex/lib/execution/runner.js:30:19)"},"level":"error","message":"Unhandled NODE exception","trace_id":"835d6d47d64dbb656316046cb9748618","span_id":"c8c52cfd98351876","trace_flags":"01","service_name":"proposal-backend","timestamp":"2026-05-06T13:03:36.187+01:00"}
> > [ERROR] 13:03:36 Error: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from "workflow_status_connections" where "workflow_status_connection_id" = ? limit ?

Hi, @TCMeldrum  Regarding the status actions on manual changes there is another PR addressing it
https://github.com/UserOfficeProject/user-office-core/pull/1441

apps/frontend/src/components/settings/workflow/WorkflowView.tsx
jekabs-karklins: Thanks for noticing, I also cleared up the WorkflowEditor. It had the same dead code

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
jekabs-karklins: Excellent, thanks!

apps/backend/src/utils/stringStripHtml.ts
jekabs-karklins: Hi, @ACLay, I'm sorry but this is not a change of this Pull Request. I am not sure why is it showing up here. Github bug?

About this changhe. Three weeks ago we were having some problems with this library due to depdendencies, and we decided to remove it completely as we were using only one function.
https://github.com/UserOfficeProject/user-office-core/pull/1476

apps/backend/src/datasources/postgres/FileDataSource.ts
jekabs-karklins: This change is also not part of this Pull Request

apps/backend/src/workflowEngine/stateMachine/stateMachine.ts
ACLay, line 1: Only realising on the third pass that this filename is misspelled, too many n's.

jekabs-karklins: Right. 👍🏻

ACLay (approved):
I'd still want confirmation that @TCMeldrum and @simonfernandes are okay with this since they've also made comments, but I think the code's looking alright now.

Hopefully there's no more giant refactorings like this for a while.

TCMeldrum (approved):
Testing it locally it seems good, the things i found initally have been adressed.I'm happy for it to go in

simonfernandes:
> > I haven't reviewed the code but this looks fantastic 👍
> > I was just getting this running locally and there are some workflow changes that STFC would need to make, where we have proposals in statuses that the workflows don't contain:
> > ```sql
> > -- Add SUBMITTED_LOCKED status to 'Import workflow'
> > INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> > VALUES(5, 15, 1, 1, 0);
> > 
> > -- Add SUBMITTED_LOCKED status to 'Import workflow for Xpress'
> > INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> > VALUES(92, 15, 1, 1, 0);
> > 
> > -- Add UNSUCCESSFUL status to 'Import workflow for Xpress'
> > INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> > VALUES(92, 20, 1, 1, 0);
> > 
> > -- Add FINISHED status to 'Import workflow for Xpress'
> > INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> > VALUES(92, 21, 1, 1, 0);
> > 
> > -- Add EXPIRED status to 'Import workflow'
> > INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> > VALUES(5, 9, 1, 1, 0);
> > 
> > -- Add EXPIRED status to 'ISIS Rapid workflow'
> > INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> > VALUES(4, 9, 1, 1, 0);
> > 
> > -- Add EXPIRED status to 'ISIS Xpress workflow'
> > INSERT INTO workflow_connections(workflow_id, status_id, pos_x, pos_y, sort_order)
> > VALUES(91, 9, 1, 1, 0);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > These statuses are all added to the workflows as unconnected and I don't see a problem with doing this.
> > Automatic status changes (e.g. on proposal submission) are working for Xpress/Technique proposals and regular proposals, and status actions are being executed.
> > But manual status changes are only partly working for Xpress/Technique proposals and regular proposals. The status is changing successfully but I'm seeing the below error and status actions are not being executed:
> > > [2026-05-06T13:03:36.187+01:00] ERROR - Unhandled NODE exception
> > > {"exception":{"name":"Error","message":"Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from "workflow_status_connections" where "workflow_status_connection_id" = ? limit ?","stack":"Error: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from "workflow_status_connections" where "workflow_status_connection_id" = ? limit ?\n    at QueryCompiler_PG.toSQL (/app/backend/node_modules/knex/lib/query/querycompiler.js:112:13)\n    at QueryBuilder_PostgreSQL.toSQL (/app/backend/node_modules/knex/lib/query/querybuilder.js:84:44)\n    at ensureConnectionCallback (/app/backend/node_modules/knex/lib/execution/internal/ensure-connection-callback.js:4:30)\n    at Runner.ensureConnection (/app/backend/node_modules/knex/lib/execution/runner.js:318:20)\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at Runner.run (/app/backend/node_modules/knex/lib/execution/runner.js:30:19)"},"level":"error","message":"Unhandled NODE exception","trace_id":"835d6d47d64dbb656316046cb9748618","span_id":"c8c52cfd98351876","trace_flags":"01","service_name":"proposal-backend","timestamp":"2026-05-06T13:03:36.187+01:00"}
> > > [ERROR] 13:03:36 Error: Undefined binding(s) detected when compiling FIRST. Undefined column(s): [workflow_status_connection_id] query: select * from "workflow_status_connections" where "workflow_status_connection_id" = ? limit ?
> 
> Hi, @TCMeldrum Regarding the status actions on manual changes there is another PR addressing it #1441

Ah I see, I hadn't realised that was being dealt with seperately. In that case everything else looked good locally so I'm also fine with this 👍

PR #1345, opened 2026-02-09
fix: bug for searching by proposal ID failing to find proposal due to whitespace

YufanKambang:
## Description
This pr is in response to the issue : https://github.com/UserOfficeProject/issue-tracker/issues/1211

## Motivation and Context

This is to improve user experience as well as to not cause confusion if they copied a proposal ID with a white space and see no proposals being shown

## How Has This Been Tested

I have tested it using the localhost proposal portal, I first checked to see if the given proposal ID had a proposal in the system,  when confirmed I tested it by adding whitespace to the start of the proposal ID, and the at the end of the ID.


YufanKambang:
Found the function `getProposalById(){}` in three different locations, it was confusing.
This is where the query is happening: [line 1036](https://github.com/UserOfficeProject/user-office-core/blob/9d2d2e616be7c7d8b60c972731e1ca458da91364/apps/backend/src/datasources/postgres/ProposalDataSource.ts#L1036) of /apps/backend/src/datasources/postgres/ProposalDataSource.ts

However the `getProposalById(){}` [here](https://github.com/UserOfficeProject/user-office-core/blob/9d2d2e616be7c7d8b60c972731e1ca458da91364/apps/backend/src/queries/ProposalQueries.ts#L158) if of a higher level, as it includes the authorisation, and I think we should add the check and fix for the bug here

YufanKambang:
Have made the fix, but yet to test it. Having issues with logging into office.localhost and user.localhost.

YufanKambang:
Fixing my local set up of the ICD system to be able to test the code changes, and assess that it actually fixes the whitespace bug

apps/backend/src/queries/ProposalQueries.ts
```diff
@@ -154,7 +156,7 @@ export default class ProposalQueries {
       searchText
     );
   }
-
+  // here is where we use the query function by proposal ID
```
mutambaraf: We can remove this comment

apps/backend/src/queries/ProposalQueries.ts
```diff
@@ -109,7 +111,7 @@ export default class ProposalQueries {
         offset,
         sortField,
         sortDirection,
-        searchText
+        noWhitespaceSearchText
```
mutambaraf: We can  just remove the whitespace here like ```searchText?.trim()```  and not declare another variable.

apps/backend/src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -1033,6 +1033,7 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
     return relatedUsers;
   }
 
+  //something here about `proposalId` is being given without validating before being given to graphql
```
mutambaraf: We may not need this comment.

mutambaraf (commented):
Just a some few comments and can we also have tests e2e to this bug or backend unit tests.

apps/backend/src/queries/ProposalQueries.ts
ellen-wright, line 1: Would be good if this change was done for the instrument scientist table too

YufanKambang: I do not have the instrument scientist role. Is it the same as the experiment scientists?

apps/backend/src/queries/ProposalQueries.ts
YufanKambang: Removed the comment

apps/backend/src/queries/ProposalQueries.ts
YufanKambang: Made this change, the variable is not used again and so no need to declare

apps/backend/src/datasources/postgres/ProposalDataSource.ts
YufanKambang: You are right, this comment was to help me with investigation

apps/backend/src/queries/ProposalQueries.ts
YufanKambang: I was looking into this and the way the proposal id get passed to be turned into a query is a bit different for instrument-scientist compared to user-officers. I think it would help to put this into a different issue, that I can pick up

YufanKambang:
> Fixing my local set up of the ICD system to be able to test the code changes, and assess that it actually fixes the whitespace bug

I have added a unit test for the role user-officer, for the whitespace bug.

YufanKambang:
The unit test produced actually tests the wrong function that is being called, as we are not trying to get a proposal from the database, but to get filter the view from the user-officer table using the `proposalId`.

YufanKambang:
I have tried to test this locally but I have failed facing an import error with `muhammara`.
I am pushing the test so that the github workflow can run it

YufanKambang:
I also noticed no instructions on how to run unit and e2e tests locally on your work machine in `user-office-core` readme.md, or in the [user office dev documentation](https://userofficeproject.github.io/user-office-core/develop/developer-guide/CONTRIBUTING/). Updating this so that new developers to the repository will know how to run the tests.

apps/backend/src/queries/ProposalQueries.ts
YufanKambang: I have made the update so that the instrument/experimental scientist table also does not get effected by whitespace. tested it on localhost using the role experimental scientist

YufanKambang: For the roles user-officer, the searchText that they put into the search bar int he proposal tables is directly used for the view query.
However, for the role instrument-scientist the searchText is put into a filter class called `ProposalsFilter`, and this `ProposalsFilter` is passed into the method `getInstrumentScientistProposals` and so we have to update the `ProposalsFilter.text` field with the trimmed searchtext that the intrument/experimental scientist will input into the intrument/experimental scientist proposal tables.

apps/backend/src/queries/ProposalQueries.ts
```diff
@@ -109,7 +109,7 @@ export default class ProposalQueries {
         offset,
         sortField,
         sortDirection,
-        searchText
+        searchText?.trim()
```
yoganandaness, line 112: I would recommend trim to happen in datasource layer during sql query rather than the business logic layer here. 

Let me know if you have any concerns

apps/backend/src/queries/ProposalQueries.ts
```diff
@@ -126,9 +126,18 @@ export default class ProposalQueries {
     first?: number,
     offset?: number
   ) {
+    const cleanText = filter?.text?.trim();
```
yoganandaness: I would recommend trim to happen in datasource layer during sql query rather than the business logic layer here.

Let me know if you have any concerns

yoganandaness (changes requested):

apps/backend/src/queries/ProposalQueries.ts
YufanKambang: This was moved nicely to the datasource layer

YufanKambang: image
And it still works just as intended

apps/backend/src/queries/ProposalQueries.ts
YufanKambang: This one is I have not been able to successfully move it to the datasource layer

image

I placed console.logs but have not been able to see them in any container logs.

How should I be handling this object filter for the `@Authorized([Roles.INSTRUMENT_SCIENTIST, Roles.INTERNAL_REVIEWER])` proposal view as they have a whole object that gets passed compared to the string for `@Authorized([Roles.USER_OFFICER])`?

YufanKambang:
I have been unable to push these changes due to the pre-commit linter. That is the reason why I have snipped my code changes

apps/backend/src/queries/ProposalQueries.ts
YufanKambang: This has now been fixed, the blocker was that the datasource layer that the fix was being applied to was the wrong one, it should have been `StfcProposalDataSource.ts` as that is the actual datasource layer being used for this proposal view.

yoganandaness (approved):

ellen-wright (approved):
Looks good

PR #1346, opened 2026-02-10
fix: fap reviewer filter sql bug

TCMeldrum:
## Description

The fap reviewer filter was broken due to a bad sql this fixes it


William-Edwards-STFC (approved):

ACLay (commented):
How does this work without the proposals join? It looks like if just `active` was specified we'd be trying to generate sql like "select from fap_reviews join call on call.call_id = proposals.call_id" and proposals would have no definition.

ACLay (approved):

PR #1347, opened 2026-02-11
refactor: migrate validation library into monorepo

jekabs-karklins:
## Description
This PR moves duo-validation package into user-office-core monorepo

> This PR moves the validation package into a local workspace (validation/) consumed by backend/frontend via file deps, **requiring a one-time deletion of their node_modules after pulling** and thereafter edits in validation/src propagate after running npm install at repo root. [Read the announcement](https://github.com/UserOfficeProject/user-office-core/discussions/1361)

## Motivation and Context
The validation library was introduced before user office was a monorepo to share request validation rules with Yup library. This has always been slighly cumbersom to work with the validation library because one would be required to merge PR into validation library prior PR in user-offce could be raised due to the fact that new package was published only after changes were merged unto main branch of the duo-validation library.

The main issue was that such flow caused blockades for other PRs which wanted to modify the validation library.
Besides that the work with the validation was tedious as everything involved two repositories.

Wiith this change there is no need to publish validation library, just change the Yup validation object, then run 
npm run build in the validation folder and changes will be immediatelly reflected to both, backend and frontend.

Changes to backend/frontend and validation library will now part of the same Pull Request and can be merged together.

## Changes
1. Library now lives as "local workspace library", so all files are moved from the validation package here
2. The build process has been updated to include the validation library.
3. The package.json files of both the frontend and backend applications have been updated to reference the local copy of the validation library.


janosbabik:
Hi @jekabs-karklins, this library is used in the scheduler too: https://github.com/UserOfficeProject/user-office-scheduler/blob/b83a4f9856676c244eafce5aa84ad6546ad6a524/apps/backend/package.json#L31

jekabs-karklins:
> Hi @jekabs-karklins, this library is used in the scheduler too: https://github.com/UserOfficeProject/user-office-scheduler/blob/b83a4f9856676c244eafce5aa84ad6546ad6a524/apps/backend/package.json#L31

Hi, @janosbabik 
The types in sheduler are 
-  createScheduledEventValidationSchema,
-  updateScheduledEventValidationSchema,
-  activateBookingValidationSchema,
-  finalizeBookingValidationSchema,

Which are all speciffic to scheduler. So I would suggest to do the same for the scheduler. Keep scheduler speciffic request types in the scheduler.

gnyiri (approved):

yoganandaness (approved):

PR #1354, opened 2026-02-16
refactor: remove verbose logging from user details batch loading

janosbabik:
It removes verbose logging from the user details batch loading to reduce log noises.


jekabs-karklins (commented):
Just a though, we could also lover this to `debug` level if STFC need these logs and set our log level >= `INFO`

TCMeldrum (approved):

PR #1355, opened 2026-02-16
feat: add review visibility feature to faps

### LINKED ISSUE #1522: Faps should be able to configure when reviewers can see other reviewers reviews
Different facilitys have different requirments on when reviewers can see each other reviews, currently it is hard coded to `proposalEvents.proposal_all_fap_reviews_submitted` but thats not always the case. We should have a option in the fap config to allow different facilitys/fap panels decide when reviewers can see other peoples reviews.

TCMeldrum:
Closes  https://github.com/UserOfficeProject/issue-tracker/issues/1522

## Description
This PR introduces the review visibility feature to faps.

## Motivation and Context
This change is required to add more flexibility and control over the visibility of the reviews. It allows setting different stages when the reviews can be visible which enhances the review process. 

## Changes
1. A new column 'review_visibility' is added to the 'faps' table in the database. This column will hold the review visibility status.
2. The 'FapDataSource.ts' and 'FapDataSourceMock.ts' are updated to include 'review_visibility' in the 'Fap' model.
3. The 'ProposalDataSource.ts' has been updated to include a new boolean column 'call_fap_review_ended' to track the end status of the 'fap' review for a proposal. 

Finally, these changes have been reflected across the backend where the 'Fap' model is used.

## How Has This Been Tested?

Introduces backend tests to cover the new review visiiblity setting

## Fixes Jira Issue


apps/backend/db_patches/0207_FapReviewsVisibility.sql
```diff
@@ -0,0 +1,30 @@
...
+BEGIN
+    IF register_patch(
+        '0207_FapReviewsVisibility.sql',
+        'TCMeldrum',
+        'Add review visibility column to faps',
+        '2026-02-10'
+    ) THEN
+
+        ALTER TABLE faps
+            ADD COLUMN IF NOT exists review_visibility text;
```
ACLay: If there's only a few different values this should have, I wonder if a separate table which `faps` has a foreign key reference to might be more suitable?

apps/backend/src/queries/FapQueries.ts
```diff
@@ -182,12 +183,29 @@ export default class FapQueries {
...
+        break;
+      case FapReviewVisibility.REVIEWS_VISIBLE_FAP_ENDED:
+        reviewsVisibleOnFap = proposalEvents?.call_fap_review_ended || false;
+        break;
+      case FapReviewVisibility.REVIEWS_VISIBLE:
+        reviewsVisibleOnFap = true;
+        break;
+    }
+
     // NOTE: If not officer, Fap Chair or Fap Secretary should return all proposal assignments only if everything is submitted. Otherwise for Fap Reviewer return only it's own proposal reviews.
```
ACLay: Does this comment need updating to reflect the new logic?

apps/frontend/src/components/fap/General/FapGeneralInfo.tsx
```diff
@@ -48,6 +49,21 @@ const FapGeneralInfo = ({ data, onFapUpdate }: FapPageProps) => {
     return <UOLoader style={{ marginLeft: '50%', marginTop: '100px' }} />;
   }
 
+  const reviewVisibilityOptions = [
```
ACLay: These options look identical to the ones in AddFap.tsx. Would this make sense moving to a separate file to avoid duplication and keep the texts from falling out of sync?

apps/backend/db_patches/0207_FapReviewsVisibility.sql
TCMeldrum: I have refactored it to do this. I think in terms of data it is better but in terms of reability of code it is as nice, I prefered seeing `Fap` objects with `reviewVisibility: FapReviewVisibility.PROPOSAL_REVIEWS_COMPLETE` rather than `reviewVisibility: 1`. I couldnt see a nice way around it tho

ACLay (approved):
Agree it's a shame we couldn't keep the API property as the enum, but I think this looks good.

apps/frontend/src/components/fap/General/FapGeneralInfo.tsx
```diff
@@ -15,6 +16,21 @@ import { useCheckAccess } from 'hooks/common/useCheckAccess';
 import { StyledButtonContainer } from 'styles/StyledComponents';
 import useDataApiWithFeedback from 'utils/useDataApiWithFeedback';
 
+export const reviewVisibilityOptions = [
```
jekabs-karklins: suggestion: could the description be added to the DB table column `description`? It would be easier to comprehend and maintain.

apps/frontend/src/components/fap/General/FapGeneralInfo.tsx
```diff
@@ -158,6 +176,20 @@ const FapGeneralInfo = ({ data, onFapUpdate }: FapPageProps) => {
                 }
                 disabled={!hasAccessRights || isExecutingCall}
               />
+              <Grid item sm={6} xs={12}>
```
jekabs-karklins: suggestion: Could we improve the layout a little, and make the components aligned. I believe that instead of using two `Grid container`s, if we just use one it will square the form nicely, like here https://mui.com/material-ui/react-grid/#row-and-column-spacing
WDYT?

Image

apps/frontend/src/components/fap/General/FapGeneralInfo.tsx
```diff
@@ -158,6 +176,20 @@ const FapGeneralInfo = ({ data, onFapUpdate }: FapPageProps) => {
...
+                <Field
+                  id="reviewVisibility"
+                  name="reviewVisibility"
+                  label="Review visibility"
+                  onChange={handleChange}
+                  component={Select}
+                  onClose={() => {}} // Override FormikUISelect.tsx custom on close as it is not needed and is chang the int to string
+                  data-cy="fap-review-visibility-filter"
+                  options={reviewVisibilityOptions}
+                  fullWidth
```
jekabs-karklins, line 183: suggestion: IIt would be nice if the dropdown would be the size of the column width, I see it is added here but appearantly it cuts of the text

Image

apps/backend/src/queries/FapQueries.ts
```diff
@@ -182,12 +183,29 @@ export default class FapQueries {
     const proposalEvents =
       await this.proposalDataSource.getProposalEvents(proposalPk);
 
-    // NOTE: If not officer, Fap Chair or Fap Secretary should return all proposal assignments only if everything is submitted. Otherwise for Fap Reviewer return only it's own proposal reviews.
+    const visibility = await this.dataSource.getFapReviewVisibility(fapId);
+
+    let reviewsVisibleOnFap = false;
+
+    switch (visibility) {
```
jekabs-karklins, line 190: comment: This is concerning part for me, because with the upcoming workflow changes the table proposalEvents is deleted and replaced with Guards, which dynamically will check if the condition is still true.
So after this is merged I will modify in my pending PR this part

jekabs-karklins (commented):
Hi, I think this is a great feature and I reached out to our User Officer to inquiry if this is something that could be used by us. I am not too familiar of how FAP works but if I understand corectly could we improve the wording "Reviews visible after the FAP has ended" to more like "After the FAP review period has ended"?

apps/backend/src/queries/FapQueries.ts
TCMeldrum: Oh sorry I haden't clocked that table was being remove sorry. Thanks for incorperations this change!

TCMeldrum:
I have updated the layout hopefully this is what you meant:
image

apps/frontend/src/components/fap/General/FapGeneralInfo.tsx
```diff
@@ -38,6 +43,8 @@ const FapGeneralInfo = ({ data, onFapUpdate }: FapPageProps) => {
   };
 
   const sendFapUpdate = async (values: Fap): Promise<void> => {
+    console.log(values);
```
yoganandaness: Can this be removed. ?

yoganandaness (changes requested):
Recommending e2e, so that when @jekabs-karklins merges the new workflow system, this would potentially break and be a reminder for fixing it up.

jekabs-karklins:
@TCMeldrum  thanks for addressing the comments, it looks good now. I see @yoganandaness has added some additional comments. Looks good from my side otherwise 👍🏻

yoganandaness (changes requested):
It will also be great to update the documentation.

TCMeldrum:
@yoganandaness I have added a test but there were some issues, please let me know what you think.

From my comment in the test:


      // Right now there is not really a easy way to test reviews_visible_fap_ended
      // The only way to set the proposals to have the call_fap_review_ended on the proposals events table
      // Is to have the checkCallsFAPReviewEndedJob async job to to run ait only runs once a day at 00:07

yoganandaness (approved):

PR #1357, opened 2026-02-16
fix: have technique management page default to most recent active call

ellen-wright:
## Description
refs: https://github.com/UserOfficeProject/issue-tracker/issues/1521

Sets the default call filter in technique management page to be the newest active call. 
## Motivation and Context
We have a bug where the call filter is still defaulting to the previous technique call instead of moving over to the newly created one for the new year. 

This PR sets the URL parameter as the active call with the highest call_id which will be the newest technique call and therefore as there is only one technique call open all year the relevant one will be the default. 


TCMeldrum (approved):

apps/frontend/src/components/menu/MenuItems.tsx
```diff
@@ -74,8 +74,8 @@ const MenuItems = ({ currentRole }: MenuItemsProps) => {
     },
     CallsDataQuantity.MINIMAL
   ).calls;
-
-  const openCall = calls?.find((call) => call.isActive);
+  const sortedcalls = calls.sort((a, b) => (a.id > b.id ? -1 : 1));
```
yoganandaness: Should we take it to the backend instead of frontend?

yoganandaness (changes requested):

apps/frontend/src/components/menu/MenuItems.tsx
yoganandaness: Also recommending unit tests and e2e, as necessary

apps/backend/src/resolvers/queries/CallsQuery.ts
```diff
@@ -59,8 +59,11 @@ export class CallsFilter {
   @Field(() => Boolean, { nullable: true })
   public isCallUpcoming?: boolean;
 
-  @Field(() => Boolean, { nullable: true })
-  public isOrdered?: boolean;
+  @Field(() => String, { nullable: true })
```
yoganandaness: This is great, still slightly deviating from other places. 

Recommending to follow [this](https://github.com/UserOfficeProject/user-office-core/blob/a5434c0d7920c1ebf0f3739d29f34fe1d643c173/apps/backend/src/resolvers/queries/ExperimentsQuery.ts#L103) approach as we have the same in most other places 

1. Letting the sortField and sortDirection open as string could result in sql injection error. This is the reason, we want to use [PaginationSortDirection](https://github.com/UserOfficeProject/user-office-core/blob/a5434c0d7920c1ebf0f3739d29f34fe1d643c173/apps/backend/src/resolvers/queries/ExperimentsQuery.ts#L65)

Note: The term pagination is misleading, as this can be used for non pagination entities like here in calls
2. I would recommend sort outside the filter object, since filter is more subject oriented to call and sort is quite generic. It also aligns with the other implementations as well(Proposals, Experiments, Users, Questions etc., )

apps/frontend/src/components/call/CallsTable.tsx
```diff
@@ -367,7 +367,8 @@ const CallsTable = ({ confirm, isArchivedTab }: CallTableProps) => {
   const getCallOrder = (): void => {
     setCallsFilter(() => ({
       ...getFilterStatus(callStatus as CallStatusFilters, isArchivedTab),
-      isOrdered: true,
+      sortField: 'sort_order',
+      sortDirection: 'asc',
```
yoganandaness:
```suggestion
      sortDirection: PaginationSortDirection.asc,
```

apps/frontend/src/components/menu/MenuItems.tsx
```diff
@@ -71,6 +71,8 @@ const MenuItems = ({ currentRole }: MenuItemsProps) => {
   const calls = useCallsData(
     {
       proposalStatusShortCode: 'QUICK_REVIEW',
+      sortField: 'call_id',
+      sortDirection: 'desc',
```
yoganandaness:
```suggestion
      sortDirection: PaginationSortDirection.desc,
```

apps/frontend/src/components/proposal/ProposalChooseCall.tsx
```diff
@@ -29,7 +29,8 @@ const ProposalChooseCall = () => {
       isActive: true,
       isEnded: false,
       isActiveInternal: isInternalUser,
-      isOrdered: true,
+      sortField: 'sort_order',
+      sortDirection: 'asc',
```
yoganandaness:
```suggestion
      sortDirection: PaginationSortDirection.asc,
```

yoganandaness (changes requested):

yoganandaness (approved):

TCMeldrum (approved):

PR #1362, opened 2026-02-19
fix: add error notification for sample creation when no sample declaration was found in proposal template

jekabs-karklins:
## Description
This PR introduces an error notification for users when attempting to create a sample if no sample declaration was found in the proposal template.

## Motivation and Context
This is not a main use case and should ideally never occur. However, if a proposal template is missing a sample declaration, we should handle it gracefully: instead of application throwing a generic error, we display a clear notification explaining that a sample can’t be created and advising the user to contact the User Office. This provides better feedback and reduces confusion in this edge case.

## Changes
- Implemented the use of `useSnackbar` hook to display an error notification when no sample declaration is found.
- Added clientLog to report the issue

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/SWAP-1983](https://jira.esss.lu.se/browse/SWAP-1983)


jekabs-karklins:
@deepaksftc I have marked this as `review:please`. Check it at your earlies convenience :)

yoganandaness (approved):

PR #1366, opened 2026-02-20
feat: automatically open for editing the cloned sample

jekabs-karklins:
## Description
This PR introduces a new feature that automatically opens for editing the cloned sample.
output_highres

## Motivation and Context
This change was necessary to improve the user experience by reducing the number of steps a user has to take to edit a cloned sample. It also feels like natural thing to open the cloned sample because user would want to do some modifications.

## Changes
1. In `QuestionaryComponentSampleDeclaration.tsx`, after cloning a sample, the cloned sample is automatically set as the selected sample for editing.
2. (bugfix) A new condition is added in `SampleAuthorization.ts` that checks if the user is an officer and returns true if they are.


TCMeldrum:
Should we add the same feature to gerneric templates as they work basically in the same way?

jekabs-karklins:
> Should we add the same feature to gerneric templates as they work basically in the same way?

Let me add it

yoganandaness (approved):

jekabs-karklins:
@TCMeldrum I have added the same functionality to the generic templates

Scott-James-Hurley:
Just a thought: could a user mistake this for editing the existing sample the first time they do this? If that could be the case, perhaps the pop up could mention "copy" somewhere.

jekabs-karklins:
> Just a thought: could a user mistake this for editing the existing sample the first time they do this? If that could be the case, perhaps the pop up could mention "copy" somewhere.

When you clone the sample it has the words "Copy of..." in the title. Is that what you mean?

Scott-James-Hurley (approved):

PR #1369, opened 2026-02-22
fix: optimise users query frontend and backend

yoganandaness:
## Description

This PR introduces a refactor and consolidation of user and participant selection components across the frontend, along with backend enhancements to user search and access control.

## Motivation and Context

The participant selection logic had diverged over time, with multiple overlapping components and proposal-specific variations. This increased cognitive load, duplicated logic, and made future changes error-prone.

This refactor consolidates participant selection into clearer, reusable components, improving maintainability and consistency across the application. Removing legacy components and unused props reduces technical debt and simplifies the codebase.

On the backend, expanding search fields improves administrative and user workflows when locating users. Tightening access control and removing reliance on externally provided `userId` parameters strengthens security by ensuring queries are scoped to the authenticated agent context.

## How Has This Been Tested

Manually

## Fixes

N/A

## Changes

- Refactor: Consolidated and renamed participant selector components across the frontend.
- Cleanup: Removed `ProposalParticipantLegacy.tsx` and eliminated unused props and proposal-specific modal logic.
- Enhancement: Expanded backend user search to include `email` and `oidc_sub`.
- Security: Strengthened role-based access control for user queries.
- Security/Correctness: Updated `getPreviousCollaborators` to use `agent.id` and guard against missing agent.

## Depends on

None.


apps/backend/src/queries/UserQueries.ts
```diff
@@ -114,7 +114,15 @@ export default class UserQueries {
     return this.dataSource.checkEmailExist(email);
   }
 
-  @Authorized()
+  @Authorized([
```
jekabs-karklins: question: I was wonder what is the reason of adding this? I think the least permissive one is the USER. we also add now PROPOSAL_READER, so we would need to remember adding it here as well as for all the future roles. Do you want to exclude a role here?

apps/backend/src/queries/UserQueries.ts
```diff
@@ -133,10 +141,9 @@ export default class UserQueries {
     };
   }
 
-  @Authorized()
+  @Authorized([Roles.USER_OFFICER])
   async getPreviousCollaborators(
```
jekabs-karklins, line 137: breaking: I think this is used in participant selector and needed for the USER role as well

apps/backend/src/queries/UserQueries.ts
```diff
@@ -145,8 +152,15 @@ export default class UserQueries {
     userRole?: UserRole,
     subtractUsers?: [number]
   ) {
+    if (!agent) {
```
jekabs-karklins, line 147: suggestion: I tink we normally mark agent with "!" if there is @Authorized abobe the function, as for some reason TS can not infer that the agent must be defined. Or i wonder can you just change 146 and remove the `null` ? Not a strong oppinion here just observation.

apps/frontend/src/components/technique/TechniqueTable.tsx
```diff
@@ -253,21 +253,23 @@ const TechniqueTable = () => {
...
-        addParticipants={assignScientistsToTechnique}
-        selectedUsers={selectedTechnique?.scientists.map(
-          (scientist) => scientist.id
-        )}
-        selection={true}
-        userRole={UserRole.INSTRUMENT_SCIENTIST}
-        title={t('instrumentSci')}
-        invitationUserRole={UserRole.INSTRUMENT_SCIENTIST}
-      />
+      {isUserOfficer && (
```
jekabs-karklins, line 256: question: What happens if this condition is not in place?

apps/backend/src/queries/UserQueries.ts
yoganandaness: My Mistake. Thanks a lot.

apps/backend/src/queries/UserQueries.ts
yoganandaness: Copy that.

apps/backend/src/queries/UserQueries.ts
yoganandaness: Shall we take this to a separate improvement? I see some potential fix to this overall.  @jekabs-karklins

apps/frontend/src/components/technique/TechniqueTable.tsx
yoganandaness: The Technique page is currently accessible only to the User Officer. Check [here](https://github.com/UserOfficeProject/user-office-core/blob/85255eb1a7f0b61ae111a2d6fc833de107d671d3/apps/frontend/src/components/menu/MenuItems.tsx#L184). 

The PeopleSelectorModal is the generic modal, that is used to assign people to different entities including Technique. Since the Technique - Scientist assignment is a User Officer's only role, I thought to add this condition, so that other Role can't access this, if this component is used in the future

This is just a frontend condition and doesn't add a lot of value, other than self-documentation. 

Let me know if you would like to remove this.

jekabs-karklins: No need to remove

jekabs-karklins (approved):
LGTM

TCMeldrum (approved):
Looks good to me testing in stfc mode seems to be fine

PR #1370, opened 2026-02-23
feat: only show calls that are related to techniques assigned to an instrument scientist

### LINKED ISSUE #1518: Add call filtering to (Xpress) management page
We need to add filtering of calls by a user assigned role tags on the management page only.

mutambaraf:
## Description
This PR improves the filtering of calls related to techniques assigned to an instrument scientist.
Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1518

## Motivation and Context

Currently, the Xpress Management page displays all calls with a QUICK_REVIEW status. This PR introduces changes that allow filtering calls based on assigned techniques. Additionally, the "All" filter has been updated to return only proposals with calls relevant to those filters. This ensures instrument scientists only see proposals related to their specific facility and assigned techniques.

## Changes
- The filter now checks for call IDs in the `callIds` array in addition to the `callId` field.
- This change is implemented in the `PostgresProposalDataSource` and `StfcProposalDataSource` classes in the backend, and reflected in the frontend in the `ProposalsQuery` class and the `MenuItems` and `ProposalMenuListItem` components.
- The `callIds` field is added to the `ProposalsFilter` class.
- The `techniqueProposalUrl` is now set to `/TechniqueProposals` instead of being dependent on the `openCall.id`.
  
These changes ensure that only the calls related to techniques assigned to the instrument scientist are displayed, improving the relevance of the displayed data for the user.

## Fixes Jira Issue


apps/frontend/src/components/menu/MenuItems.tsx
ellen-wright, line 1: I think we want to keep the call.id so that it defaults on the open call of that year

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -179,6 +197,31 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
     excludeProposalStatusIds: excludedStatusIds,
   });
 
+  const lastProcessedCallId = useRef<number | null>(null);
+  useEffect(() => {
+    if (calls && calls.length > 0 && !searchParams.get('call')) {
+      const activeCall = calls.find((call) => call.isActive);
```
mutambaraf: @ellen-wright  .Yes we are still setting it, but I am setting it here inside the table because if I set it higher up, the set ID might technically not be part of the call the user has access to.

ellen-wright: I am currently dealing with this issue in this PR https://github.com/UserOfficeProject/user-office-core/pull/1357 so might be worth us trying to coordinate this

mutambaraf: Will wait for your PR as discussed.

apps/backend/src/resolvers/queries/ProposalsQuery.ts
```diff
@@ -78,6 +78,9 @@ export class ProposalsFilter {
   @Field(() => Int, { nullable: true })
   public callId?: number;
 
+  @Field(() => [Int], { nullable: true })
```
yoganandaness, line 81: If we are going to have `callIds`, will it make sense to remove the `callId`?

apps/backend/src/datasources/postgres/CallDataSource.ts
```diff
@@ -29,7 +29,7 @@ import {
 
 const fieldMap: { [key: string]: string } = {
   sort_order: 'sort_order',
-  call_id: 'call_id',
+  call_id: 'call.call_id',
```
yoganandaness, line 32: Any specific reason to this? If important, should the `sort_order` also be `call.sort_order`?

apps/backend/src/resolvers/queries/ProposalsQuery.ts
```diff
@@ -78,6 +78,9 @@ export class ProposalsFilter {
...
+  public callIds?: number[];
```
yoganandaness, line 82: I feel, there should be an Authorizer in the backend that validates the callIds and decide if they are associated with the Instrument Scientists through the Technique. Else, it will be easy to manipulate the graphql query to pass the callIds, that are not associated and get unauthorized access. 


Also would be worth to check if the [PR](https://github.com/UserOfficeProject/user-office-core/pull/1282) already solved this use case

yoganandaness (changes requested):

apps/backend/src/datasources/postgres/CallDataSource.ts
mutambaraf: It will not be necessary  because , only added that because `call_has_instruments` has also a call_id and the database was complaining because of this being ambiguous .It was broken but looks like we never sent instruments as part of filters.

apps/backend/src/resolvers/queries/ProposalsQuery.ts
mutambaraf: I can do it but just wanted to maintain backward compatibility because l saw so many places it is being used.

apps/backend/src/resolvers/queries/ProposalsQuery.ts
mutambaraf: We have not started implementing the PR for it to work with our current environment and we have a very short time to release for LSF Xpress. For `Authorizer`, we already have have one here `@Authorized([Roles.INSTRUMENT_SCIENTIST, Roles.USER_OFFICER])` on ```getTechniqueScientistProposals``` . These are only filter values on what the user already have access to we are just being more explicitly that we only need proposals associated with these call ids instead of all of them.

ellen-wright (approved):

yoganandaness (approved):

PR #1371, opened 2026-02-23
feat: allow proposal template name to be used in email templates

ellen-wright:
## Description

For a new technique proposal access route we have emails that need to include the generic access route name, which is in the proposal template name - as apposed to using the call short code that will include the year - the template name will not change for all the calls for this access route. 

It also allows us to keep one email template for multiple technique access routes. 

This allows us to keep the duplication in email templates to a minimum and include the access route name in the email.


apps/backend/src/statusActionEngine/statusActionUtils.ts
```diff
@@ -170,6 +179,15 @@ export const getEmailReadyArrayOfUsersAndProposals = async (
         let techniques: Technique[] = [];
         let hazardAnswers: Answer[] = [];
         let sampleAnswers: Answer[] = [];
+        let proposalTemplateName: string | undefined = '';
+
+        const questionary = await questionaryDataSource.getQuestionary(
+          proposal.questionaryId
+        );
+        const temaplteId = questionary ? questionary?.templateId : 0;
```
Scott-James-Hurley:
```suggestion
        const templateId= questionary ? questionary?.templateId : 0;
```

Scott-James-Hurley: Suggestion: log that we couldn't fetch the template id

Scott-James-Hurley (changes requested):

jekabs-karklins (approved):
LGTM

apps/backend/src/statusActionEngine/statusActionUtils.ts
```diff
@@ -184,9 +185,15 @@ export const getEmailReadyArrayOfUsersAndProposals = async (
         const questionary = await questionaryDataSource.getQuestionary(
           proposal.questionaryId
         );
-        const temaplteId = questionary ? questionary?.templateId : 0;
-        const template = await templateDataSource.getTemplate(temaplteId);
-        proposalTemplateName = temaplteId ? template?.name : '';
+        const templateId = questionary ? questionary?.templateId : 0;
+        if (templateId == 0) {
```
Scott-James-Hurley: Is it possible that the template ID could actually be zero?

ellen-wright: I guess it could be possible, I'll change it to -1

Scott-James-Hurley (approved):

PR #1380, opened 2026-02-25
feat: update PDF templates to use relative paths for resources

janosbabik:
With this change we will have compatible default PDF templates for the factory with internal browser and external browser too.


yoganandaness (approved):

TCMeldrum (approved):

mutambaraf:
@janosbabik Will this solve the font issue we discussed.

mutambaraf (approved):

mutambaraf:
> @janosbabik Will this solve the font issue we discussed.

Have found where the issue is coming from.

PR #1397, opened 2026-03-04
fix: on xpress management page only load calls filter when there are instrument(s) in the technique(s)

### LINKED ISSUE #1536: fix xpress page call filter shows all call in technique has no instrument assigned
When logged in as an Experimental Scientist, all calls will be loaded on the call filter if the user is assigned a technique with no associated instruments.

mutambaraf:
## Description
This PR fixes an issue where call filters were being loaded on the Xpress management page even when no instruments were associated with the techniques.

## Motivation and Context
We only want to show calls when the is an associated instruments

## Changes
- The condition for loading calls has been updated to check for the existence of instruments associated with techniques. Now, calls will only load when there is at least one instrument in the technique(s). This change is reflected in the file `apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx`.

## Fixes Jira Issue

Closes: https://github.com/UserOfficeProject/issue-tracker/issues/1536

### DIFF
--- a/apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
+++ b/apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
@@ -100,7 +100,7 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
     },
     CallsDataQuantity.MINIMAL,
     currentRole !== UserRole.USER_OFFICER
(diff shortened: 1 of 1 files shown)


ellen-wright (approved):

yoganandaness (approved):
I am approving this. 

But it think, it is best to add the condition on the backend rather than frontend. Frontend can be easily hijacked.

PR #1407, opened 2026-03-10
casl Proof of Concept

Scott-James-Hurley:
## Description

This is a proof of concept for implementing [casl](https://casl.js.org/v6/en) in UOP.

- Checking permissions for viewing a proposal
- Checking if a user can edit a FAP
- Checking permissions for fetching FAP proposal reviews
- Showing appropriate menu items for user
- Viewing, editing, deleting permissions in the frontend

## Motivation and Context

STFC is investigating how to offer more granular and customisable permissions for facilities. casl is an [Attribute Based Access Control](https://en.wikipedia.org/wiki/Attribute-based_access_control) (ABAC) system. It tests whether the user is allowed to do (action) what they're asking to do to a resource (subject), depending on a specified set of criteria (conditions). These tuples of action, subject and condition are to be stored in the database to allow facilities to define them themselves, instead of relying on how their behaviour defined defined in the code.

### Database

Two tables have been added: `permissions` and `role_has_permission`. `permissions` defines what action a role can do on a subject and under what conditions. For example, a User Officer can view any proposal, a FAP chair can update a FAP if they chair that particular FAP. `role_has_permission` ties these permissions to user roles.

### Backend

- A datasource for fetching and manipulating permissions from the DB and turning them into [casl abilities](https://casl.js.org/v6/en/guide/intro#basics) has been added. 
- An authorisation class has been added for checking if a user can edit a FAP
- A new function has been added to knex to allow filtering db queries using casl abilities (fetching only the entries the user is allowed to see)
- the proposal authorisation class uses casl to see if a user can view a proposal

My initial thought was that an authorisation class is added for every entity where we perform authorisation checks whose role it is to construct the context object, and test the user's permissions (fetched from the permissions datasource) against said context object, returning a `true` or `false` if the user is allowed to do what they're asking.

### Frontend

A user interface has been added for interacting with permissions. This is very basic and not very user friendly. Ideally we'd abstract away the syntax for conditions and make it as hard as possible for users to define incorrect permissions. Please note that the table has some visual bugs.

https://github.com/user-attachments/assets/853a2c40-8635-4bca-a5d5-11956305d8e8

The menu items are shown based on the permissions the user currently has. This use case could simplify the code (currently a react component is defined for each user role) and would be a good candidate for permissions caching as currently a user's permissions are fetched on every page load.

### Pros

- Frontend integration is very easy with the [react package](https://casl.js.org/v6/en/package/casl-react)
- Database design for permissions is straightforward
- Expressive conditions, [casl uses a subset of the MongoDB query language](https://casl.js.org/v6/en/guide/conditions-in-depth)
- Creating SQL queries using casl abilities is also quite straightforward

### Cons

- casl only allows you to pass one context object for use in testing conditions, meaning you have to do some manual work to compare two variables (e.g. checking that the `proposer_id` on a proposal is the same as the user's ID, see)
- Gathering all the data you need for a context object can be difficult without either explicit references to user roles in the code or collecting _everything_ you might need into one large context object, potentially incurring a performance hit
- The above may be an inherent flaw of ABAC that requires us to redesign how we structure data in the backend

Please let me know your thoughts and feedback. If there's a use case you'd like to be explored, please post them here.


apps/backend/src/datasources/postgres/PermissionDataSource.ts
```diff
@@ -0,0 +1,285 @@
...
+import database from './database';
+import {
+  PermissionRecord,
+  createPermissionRuleObject,
+} from './records';
+import { CreatePermissionRuleArgs } from '../../resolvers/mutations/CreatePermissionRuleMutation';
+import { UpdatePermissionRuleArgs } from '../../resolvers/mutations/UpdatePermissionRuleMutation';
+import { PermissionRulesArgs } from '../../resolvers/queries/PermissionsQuery';
+
+export const actions = ['update', 'read', 'delete'] as const;
```
Scott-James-Hurley, line 13: Instead of just defining actions and subjects as string, I've provided a list of possible actions and subject to try to help with [type detection](https://casl.js.org/v6/en/guide/subject-type-detection). This is intended as a small, illustrative example

apps/backend/src/datasources/postgres/PermissionDataSource.ts
```diff
@@ -0,0 +1,285 @@
...
+          subject: permissionRecord.subject,
+          conditions: permissionRecord.conditions == null ? null : this.substituteConditionsValues(JSON.parse(permissionRecord.conditions), this.flatten(object))
+        }
+      )
+    });
+
+    return rules;
+  }
+
+  substituteConditionsValues(conditions: any, object?: any | undefined) {
```
Scott-James-Hurley, line 59: A condition may look like the following in the database:
`{ "user.userId": { "$in": "fap.fapChairUserIds" } }`

When we convert this into an object the object values will still be string literals instead of the values they are supposed to be (e.g. "fap.fapChairUserIds" instead of [12345]) so we have to do the substitution ourselves, looking for context object key references in the conditions object and replacing them with the actual values. In the provided example, the resulting conditions object would be:
`{ 12345: { "$in": [12345] } }`

apps/backend/src/datasources/postgres/knexTest.ts
```diff
@@ -0,0 +1,75 @@
...
+    ...$and,
+    new CompoundCondition('or', $or),
+  ]);
+  } else if (!!$or) {
+    condition = new CompoundCondition('or', $or);
+  } else {
+    return ['()', []];
+  }
+
+  const [sql, replacements] = interpret(condition, {
```
Scott-James-Hurley, line 69: Converts from ast to SQL.

apps/backend/db_patches/9998_CaslPOC.sql
```diff
@@ -0,0 +1,117 @@
...
+        ('read', 'status_action_logs', NULL, false),
+        ('read', 'template', NULL, false),
+        ('read', 'proposal', '{"isVisitorOfProposal": true}', false),
+        ('read', 'proposal', '{"isDataAccessUserOfProposal": true}', false),
+        ('read', 'proposal', '{"isMemberOfProposal": true}', false),
+        ('read', 'proposal', '{"isInstrumentManagerToProposal": true}', false),
+        ('read', 'proposal', '{"isScientistToProposalTechnique": true}', false),
+        ('read', 'proposal', '{"isScientistToProposal": true}', false),
+        ('read', 'fap_proposal_assignment', NULL, false),
+        ('read', 'fap_proposal_assignment', '{"user_id": "userId"}', true),
```
simonfernandes, line 55: This seems nice if I've interpreted it correctly: all of the permission filtering is handled at db level, so that the main check can be written without a condition for simplicity and performance.

In Casbin, if the filtering takes place and provides the exact outcome, it's still necessary to collect the context and evaluate each proposal again.

apps/backend/db_patches/9998_CaslPOC.sql
```diff
@@ -0,0 +1,117 @@
...
+        ('read', 'proposal_workflow', NULL, false),
+        ('read', 'institution', NULL, false),
+        ('read', 'people', NULL, false),
+        ('read', 'question', NULL, false),
+        ('read', 'setting', NULL, false),
+        ('read', 'status_action_logs', NULL, false),
+        ('read', 'template', NULL, false),
+        ('read', 'proposal', '{"isVisitorOfProposal": true}', false),
+        ('read', 'proposal', '{"isDataAccessUserOfProposal": true}', false),
+        ('read', 'proposal', '{"isMemberOfProposal": true}', false),
```
simonfernandes, line 50: By having conditions for a subject split across separate rows, is it somehow possible to know which particular condition has returned false for clearer error messaging?

apps/frontend/src/components/PageLayout.tsx
```diff
@@ -49,6 +50,7 @@ const PageLayout = ({
   header: string;
   children: React.ReactNode;
 }) => {
+  //caching would be very useful here
```
simonfernandes, line 53: Is this referring to caching the list of dashboard read permissions for each role so that there's no delay in loading the menu items on switching roles?

apps/backend/src/auth/ProposalAuthorization.ts
```diff
@@ -258,50 +261,45 @@ export class ProposalAuthorization {
...
         break;
       case Roles.FAP_REVIEWER:
       case Roles.FAP_SECRETARY:
       case Roles.FAP_CHAIR:
-        hasAccess = await this.isMemberOfFapProposal(
-          agent,
-          proposal.primaryKey
-        );
+        permissionContext = {
+          isMemberOfFapProposal: this.isMemberOfFapProposal(agent, proposal.primaryKey)
```
simonfernandes, line 293: In case there's any weirdness with this permission - looks like there's an `await` forgotten. Happens to me often and usually has me scratching my head for far too long

apps/backend/db_patches/9998_CaslPOC.sql
simonfernandes: Actually I'm not sure if my thinking was right. I can see the fap_reviewer role has these two permissions:

Image

But it looks like only the db permission one (34) is being used?

simonfernandes (commented):
Looks great👍

apps/backend/db_patches/9998_CaslPOC.sql
Scott-James-Hurley: separate rules collected into a single ability object are treated with the `OR` operator. So here the logic would be: "is the user a visitor of the proposal `OR` a data access user of the proposal `OR` a member of the proposal"

We could probably use something like [this](https://github.com/stalniy/casl/issues/264#issuecomment-630734303) to see which conditions aren't met.

yoganandaness:
@Scott-James-Hurley @simonfernandes @deepaksftc Thanks for the insightful work. As of now, the stakeholders are reviewing at the current implementation of [Dynamic Roles using tags](https://github.com/UserOfficeProject/user-office-core/pull/1282). Until we have a decision from them, the POC's of Casbin and Casl can stay here as is. 

Thank you again.

PR #1410, opened 2026-03-12
Casbin POC

simonfernandes:
## Description

This is a proof of concept for implementing [Casbin](https://casbin.org/) in UOP.

- Proposal **read** permission for user role
	- Which proposals are returned in the user's list of proposals
	- Checking the permission when the user clicks the view button on an individual proposal
- Proposal **update** permission for user role
	- Whether to show the view or edit button on each proposal
	- Checking the permission when the proposal is updated in some way
- Proposal **delete** permission for user role
	- Whether to enable/disable the delete button on each proposal
	- Checking the permission when the delete button is clicked

### Backend
1. Hardcoded permissions checks are replaced with a call to Casbin. The call to Casbin passes two objects representing the subject (user) and resource (e.g. proposal), and a string for the action (e.g. 'read')
2. Casbin checks for a policy (i.e. checks `casbin_rule` table for a row that matches)
	- If a policy doesn't exist - return false decision
	- If a policy exists without a condition, return true/false decision based on effect column ('allow' or 'deny')
	- If a policy condition exists, pass all context to a custom JSON evaluation function to compare field attributes and run functions, and return a decision based on that
3. The result can be used to return a resource/rejection/boolean (etc.), or for filtering a list of resources

### Frontend

1. Each resource has a UI permissions field on its GraphQL type where necessary. This is a subset of permissions for a resource that are needed in the frontend to show/hide UI information.
2. A page will request UI permissions in its GraphQL query when it needs to show/hide content based on permissions.
3. Casbin evaluates the UI permissions.
4. Hardcoded frontend permissions checks can be replaced with boolean checks on the UI permissions field.

## Motivation and Context

STFC needs more granular and customisable permissions for facilities. Casbin is used to provide a [Attribute Based Access Control](https://en.wikipedia.org/wiki/Attribute-based_access_control) (ABAC) system. It tests whether the user (subject) is allowed to do something (action) on a given resource (resource/object), depending on a specified set of criteria (condition). These tuples of action, subject and condition are stored in the database to allow facilities to define them themselves, instead of defining their behaviour in code.

### Database

Two tables have been added:

1. `casbin_rule` - core Casbin table that contains a single row for each action (e.g. `read`) on a resource (e.g. `proposal`) for a role (e.g. `user`), along with an optional id reference to a condition.
2.  `casbin_conditions` -  defines the condition in JSON format.

### Backend

This is roughly divided into Casbin setup, resource configuration, and condition handling.

#### Resource configuration

Each resource (proposal, call, etc.) has the following:
 - **Context**
	 - This is a complete object that is passed to Casbin. It has a `type` string attribute e.g. 'proposal' that Casbin uses for policy matching. Everything needed for evaluating attributes is in this object and for performance reasons it shouldn't contain more than what's required. It may also include attributes that wouldn't necessarily be associated with a resource, but are needed for evaluating, e.g. `statusShortCode` on a `Proposal`.
	 - Context is fetched using a resource's context fetcher class. It should ideally avoid looping through resources to fetch data (e.g. `getProposals(proposalpk)`), and instead should make larger database calls (`getProposals(proposalPks)`) which are mapped back
 - **UI attributes**
	 - Which context attributes should be shown in the UI. Filters out context attributes we don't care about, e.g. `type`
 - **Functions**
	 - This is for logic that can't easily be represented/evaluated in ABAC. It may require logic or database calls that makes it unfeasible to fetch in context for multiple resources efficiently.
	 - Including it as a context attribute is much preferred where possible.
 - **Filters**
	 - These are database filters derived from a policy condition, built by looking at a policy's JSON condition. They should only be needed for getting multiple allowed resources (i.e. 'read' actions).
	 - They're an important performance optimisation to reduce the number of resources that Casbin needs to check where possible.

#### Casbin setup

This is mostly around configuring Casbin with the Postgres database adapter, exposing Casbin endpoints for adding/removing/checking policies it has stored, and providing an enforcer for use in checking permissions.

#### Condition handling

Casbin usually uses a string to store expressions using its own expression language, which I used in an initial design. This design instead stores conditions in JSON, in a separate table not managed by Casbin. The reason for this is that Casbin condition expression strings cannot be parsed into structured data, which makes it very difficult to a) set a condition in the UI that is user friendly and restricted to what's technically possible and b) read the condition for the purpose of filtering at database level.

In the frontend I have used [React Query Builder](https://react-querybuilder.js.org) to easily generate conditional JSON. 

In the backend, when a condition ID is present in the policy, Casbin calls a custom function which fetches the JSON condition from the database, parses it, and evaluates it against the provided context. It supports synchronous checks (attribute comparisons) and asynchronous checks (function calls like `isCallEnded()`).

### Frontend

- Each resolver type has a field resolver for UI permissions where necessary, e.g. `uiPermissions` on  `Proposal`.
- It will call a GraphQL data loader, which allows the context to be fetched for all resource items at once to help performance.
- Within the dataloader, the context and actions for all UI permissions defined are passed to Casbin's batch function and the results are mapped back.

I haven't considered menu permissions in this POC, which would likely need to be handled a different way.

### Permissions management page

The management page loads the condition based on the combination of subject, resource and action chosen. The dropdown is restricted to attributes and functions for the current resource and subject, and the operators allowed are simple equality checks. The right hand side is just a text field, even if defining a boolean or referencing another attribute.

### Performance

Based on limited playing around, performance seems reasonable. It will be heavily dependent on how individual permissions are defined, as function calls will have most impact, and UI permissions will increase the amount of decisions made on a page load. Things like pagination, creating/utilising getAll() database calls, caching, and potentially batching Casbin calls, would be worth consideration.

### Pros

- Declarative permissions that are flexible and generally make permission checks easier to write and maintain.
- Frontend management UI is mostly user friendly and restricts the user to only setting permissions that are technically possible.
- More expensive function calls are only executed when defined in the permission.
- Mature ecosystem and proven scalability.

### Cons

- Complexity. (The resource registry and Typescript typing try to mitigate this).
- Custom JSON condition handling and condition management bypasses some of Casbin's built-in optimisations.
- [Database adapters](https://casbin.org/docs/adapters) are community maintained and have varying levels of support and features.
- Permissions can often be implemented in multiple ways and require thoughts on best practice.

https://github.com/user-attachments/assets/22accaa4-79ea-4b06-996a-768cb5f54b48


apps/backend/db_patches/9998_CasbinPOC.sql
```diff
@@ -0,0 +1,179 @@
...
+  read_con_id_var BIGINT;
+  update_con_id_var BIGINT;
+  del_con_id_var BIGINT;
+BEGIN
+    IF register_patch('9998_CasbinPOC.sql', 'simonfernandes', 'Casbin', '2026-01-16') THEN
+      BEGIN
+        CREATE TABLE casbin_rule (
+          id BIGSERIAL PRIMARY KEY,
+          ptype VARCHAR(128) NOT NULL,
+          v0 VARCHAR(128) NOT NULL DEFAULT '',
```
Scott-James-Hurley, line 13: Are these column names needed for Casbin or would they be able to have more descriptive names?

apps/backend/db_patches/9998_CasbinPOC.sql
```diff
@@ -0,0 +1,179 @@
...
+
+        /* Create default conditions so that everything works initially until overwritten in UI */
+        INSERT INTO casbin_condition (condition)
+        VALUES ($json$
+        {
+          "id": "7246f7a3-ab6b-41cb-a24e-44cbb976be98",
+          "rules": [
+            {
+              "id": "afd24676-2125-4964-bf86-86ac220b2714",
+              "field": "proposal.statusShortCode",
```
Scott-James-Hurley, line 38: I like that these conditions are readable without knowing anything about Casbin.

Scott-James-Hurley: I wonder if it be worth adding a name column so it's easier to understand what the rules are used for?

apps/backend/src/auth/ProposalAuthorization.ts
```diff
@@ -262,10 +270,21 @@ export class ProposalAuthorization {
...
-          (await this.isMemberOfProposal(agent, proposal)) ||
-          (await this.isVisitorOfProposal(agent, proposal.primaryKey)) ||
-          (await this.isDataAccessUserOfProposal(agent, proposal.primaryKey));
+        const userContext = await this.userContext.toContextData(agent);
+
+        if (!userContext) {
+          return false;
+        }
+
+        const proposalContext = (
```
Scott-James-Hurley, line 279: How easy would it be to extend this to cover the other role authorisation checks here? I struggled to think of a good way to do this in my POC.

apps/backend/src/queries/PermissionQueries.ts
```diff
@@ -0,0 +1,59 @@
...
+    @inject(Tokens.AuthRegistry)
+    public authRegistry: AuthRegistry
+  ) {}
+
+  @Authorized([Roles.USER_OFFICER])
+  async getAuthResourceMetadata(
+    agent: UserWithRole | null,
+    resourceType: ResourceType
+  ): Promise<AuthResourceMetadata> {
+    const userAttributes =
```
Scott-James-Hurley, line 25: I really like the idea of defining the shape of the context object needed for conditions checks, as well as the functions needing for gathering the data needed. I think something similar could be done with casl with some legwork.

Scott-James-Hurley (commented):
This looks terrific, well done!

apps/backend/db_patches/9998_CasbinPOC.sql
simonfernandes: It would depend on the database adapter but I think they'd probably have to stay this way a the v0–v5 naming seems to be a standard. The adapter used in this POC allows renaming the table name (`casbin_rule`) but not the columns, which is what I've seen from other adapters. https://casbin.org/docs/policy-storage#database-storage-format

apps/backend/db_patches/9998_CasbinPOC.sql
simonfernandes: Just to check, do you mean a description per condition section (the 3 rule groups in the screenshot)?

image

It looks like that would be doable but possibly a bit involved. A simpler alternative would be a single description field for the whole condition, which could possibly serve the same purpose if written clearly enough.

apps/backend/src/auth/ProposalAuthorization.ts
simonfernandes: I'm thinking it might be more of the same at that point. The proposal context (and probably the user context) would need to grow, and the list of functions would increase for anything that can't be included.

One possibility at the expense of more complexity may be that user context and resource context is fetched based on the role. E.g. when the role is FAP_CHAIR, the user context contains the list of faps the chair is member of, and the proposal context contains the faps the proposal is in. But this could get very complex and I think ideally it would be shared.

If all roles are covered in this `hasReadRights()` function then the switch could be completely removed and replaced with a call to casbin after context fetching.

Is there a particular part you think might be more tricky? I'd be happy to have more of a think or give something a go.

apps/backend/src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -752,56 +753,62 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
...
-            .where('visits_has_users.user_id', id);
-        })
-        .modify((qb) => {
-          if (filter?.instrumentId) {
-            qb.innerJoin('instrument_has_proposals as ihp', {
-              'p.proposal_pk': 'ihp.proposal_pk',
-            });
-            qb.where('ihp.instrument_id', filter.instrumentId);
-          }
+    return database
```
simonfernandes, line 756: There's only a small change required in this function:

```typescript
if (filter?.proposalStatusId) {
  qb.where('p.status_id', filter?.proposalStatusId);
}
```

The large diff is because the query was broken with filters due to the way the where clause was structured.

apps/backend/src/mutations/ProposalMutations.ts
```diff
@@ -149,12 +157,39 @@ export default class ProposalMutations {
...
-    ) {
-      return rejection('You do not have rights to update this proposal', {
-        args,
+    const userContext = await this.userContext.toContextData(agent);
+
+    if (!userContext) {
+      return rejection('Failed to update proposal', { args });
+    }
+
+    const proposalContext = (
```
simonfernandes, line 166: Context for a resource only needs to be fetched when a permission is conditional - it would be worth including this check beforehand.

apps/backend/db_patches/9998_CasbinPOC.sql
Scott-James-Hurley: Aye, the latter. Just trying to think of ways to make this stuff more accessible. I can imagine a scenario where someone unfamiliar with Casbin has to debug a faulty permission,

apps/backend/src/auth/ProposalAuthorization.ts
Scott-James-Hurley: Was wondering how you would remove the explicit references to roles in the code, thanks for answering!

yoganandaness:
@simonfernandes Thanks for the insightful work. As of now, the stakeholders are reviewing at the current implementation of [Dynamic Roles using tags](https://github.com/UserOfficeProject/user-office-core/pull/1282). Until we have a decision from them, the POC's of Casbin and Casl can stay here as is. 

Thank you again.

PR #1416, opened 2026-03-16
Casbin poc 2 -Alternate UI and db schema option

deepaksftc:
## Description
https://github.com/UserOfficeProject/user-office-core/pull/1410 is the main proof of concept for implementing [Casbin](https://casbin.org/) in UOP. This POC covers an alternate approach for permissions management page UI and db schema from that of one covered in [Casbin POC](https://github.com/UserOfficeProject/user-office-core/pull/1410).

## Motivation
Casbin would require a 'casbin_rule' table/view to do its internal logics, in which we have only limited columns with not-so user friendly naming. In the main [Casbin POC](https://github.com/UserOfficeProject/user-office-core/pull/1410) PR, we have explored the option of storing the policies in JSON format within another table 'casbin_conditions' and then reference it in 'casbin_rule' default table. In this PR, we have tried to explore the possibility of having 'casbin_rule' as a view and have a more user friendly table 'policies' which can be mapped one to one from the fields in permissions management page table. The script for creating the 'casbin_rule' view will handle the conversion logic to transform the columns in 'policies' table to casbin required format which is needed for 'casbin_rule'. This could help us to avoid the conversion part of transforming to/from the JSON format while we need to get/save the individual attributes for a particular policy and could also increase the performance for each backend calls.

1. Alternate UI design for permissions management page
2. Alternate db schema 

### Database

1. Added '[policies](https://github.com/UserOfficeProject/user-office-core/pull/1416/changes#diff-d55a00cfbf4c13852e5733819e819f2f38b01861d1f46a90c0fce607f69eca34R8)' table where, 
     - we can have separate column for each policy attributes and its operators 
     - we can have N number of policy attributes 
     - the fields are mapped directly from the permissions management page table
     - the fields can be fetched directly from the DB for a particular role-resource-action combination, which can be used for  populating filters before actual db call.

2. Created a [view](https://github.com/UserOfficeProject/user-office-core/pull/1416/changes#diff-d55a00cfbf4c13852e5733819e819f2f38b01861d1f46a90c0fce607f69eca34R45) which will transform the attributes and operators in 'policies' table to casbin required model. This view will just be used by casbin for its internal access checks.

    
### Permissions management Page

1. Added a permissions table to set permissions at attribute level
image

2. While creating/updating a permission, we get an option to choose the attribute values from pre loaded drop down list. (Only fetch permissions in implemented in this POC. Create and update not implemented yet)
image

3. We can have multiple values for any attribute, which could be accompanied by an operator value to decide if it's a AND /OR.
image

4. Conditions that doesn't fit in to any of the attribute columns can be added as a condition in condition filter
image


Scott-James-Hurley (commented):
cool stuff

yoganandaness:
@Scott-James-Hurley @simonfernandes @deepaksftc Thanks for the insightful work. As of now, the stakeholders are reviewing at the current implementation of [Dynamic Roles using tags](https://github.com/UserOfficeProject/user-office-core/pull/1282). Until we have a decision from them, the POC's of Casbin and Casl can stay here as is. 

Thank you again.

PR #1425, opened 2026-03-19
fix: handle error if the auth returns with key error

yoganandaness:
## Description

This pull request makes a small update to the `ExternalAuth` component to improve error handling for authentication. Now, the component checks for an `error` parameter in addition to `error_description` in the URL query parameters, ensuring that all possible OAuth errors are properly handled.

## Motivation and Context

To Fix infinite loop of the app caused by the new error parameter `error`

## Changes

* Enhanced error handling by checking for an `error` parameter in the query string and passing it to the `handleError` function if present. [[1]](diffhunk://#diff-6f39957ce26077cdbaf231dca87d67b685dd590febffd2c3e9b02eeb3074d3c9R38) [[2]](diffhunk://#diff-6f39957ce26077cdbaf231dca87d67b685dd590febffd2c3e9b02eeb3074d3c9R167-R176)
* 

### DIFF
--- a/apps/frontend/src/components/user/ExternalAuth.tsx
+++ b/apps/frontend/src/components/user/ExternalAuth.tsx
@@ -35,6 +35,7 @@ function ExternalAuth() {
   const code = searchParams.get('code');
   const token = searchParams.get('token');
   const errorDescription = searchParams.get('error_description');
+  const error = searchParams.get('error');
 
   const unauthorizedApi = useUnauthorizedApi();
   const navigate = useNavigate();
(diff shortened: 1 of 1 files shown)


simonfernandes (approved):

jekabs-karklins (approved):
Perfect I see this is a good fix, because according to the spec the `error` is actually the mandatory field, while `error_description` is optional.

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

PR #1434, opened 2026-03-25
perf: add delay to search for people table

ellen-wright:
## Description
refs: https://github.com/UserOfficeProject/issue-tracker/issues/1546

The people table has some performance issues, we have a lot of users and on every character typed the search is executed. 

This PR adds a delay into the search bar so the user has time to type the full name they are searching for, reducing the number of calls to the database. 

## How Has This Been Tested
Tested manually 

## Changes
Delay added to People table search 

### DIFF
--- a/apps/frontend/src/components/user/PeopleTable.tsx
+++ b/apps/frontend/src/components/user/PeopleTable.tsx
@@ -455,6 +455,7 @@ const PeopleTable = ({
           }}
           options={{
             search: search,
+            searchDebounceDelay: 1500,
             searchText: persistUrlQueryParams
               ? searchParams.get('search') || undefined
               : undefined,
--- a/apps/frontend/src/components/user/PeopleTable.tsx
+++ b/apps/frontend/src/components/user/PeopleTable.tsx
@@ -455,7 +455,7 @@ const PeopleTable = ({
           }}
           options={{
             search: search,
-            searchDebounceDelay: 1500,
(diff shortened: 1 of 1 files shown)


Scott-James-Hurley (approved):
looks good, maybe worth adding a comment that it's a 1500ms delay?

apps/frontend/src/components/user/PeopleTable.tsx
```diff
@@ -455,6 +455,7 @@ const PeopleTable = ({
           }}
           options={{
             search: search,
+            searchDebounceDelay: 1500, //1500ms delay on search
```
yoganandaness: Does it overwrite debounceInterval? 

Also, 1500ms is on the higher side. Best to stay below 500?

yoganandaness (changes requested):

apps/frontend/src/components/user/PeopleTable.tsx
ellen-wright: The interval at the moment is 400ms and it is not giving enough time, I don't think 500 is enough either, I agree 1500 is too high but we need a good amount of time to let someone type enough of a name in before the search happens, I think maybe around 700. I'd rather have a pause then have the table freeze loading which is what is does currently if you only type 1-2 letters

yoganandaness: As we agreed, i would recommend having a Search button, that will solve the complexity of making unnecessary api call, while typing.

PR #1441, opened 2026-03-27
feat: add the ability to execute status actions when the proposal status is updated manually

jekabs-karklins:
## Description
THIS PR IS BUILT ON TOP OF  #[SWAP-4949-workflow-overhaul](https://github.com/UserOfficeProject/user-office-core/tree/SWAP-4949-workflow-overhaul)

This PR adds the ability for the User Officer to execute status actions when he/she sets the proposal status manually.

## Motivation and Context
After the workflow overhaul the status actions are attached to the transition, and not the status. In same cases when setting the proposal status the user officer would want to run the status actions accompanied with the transition leading to that status.

This PR handles three scenarios:

- If there are no status actions before the chosen state, user can not select anything
- If there is one incoming transition with status action user can decide to run it.
- If there are multiple, user can decide if he/she wants to run it and if yes from which transition (see screenshots)

image

image

## Changes
- Added `workflowConnectionId` parameter to the `changeProposalsStatus` mutation in `ProposalMutations.ts`, which is used to fetch the proposal workflow and run the status engine.
- Updated `changeProposalsStatus` mutation to only run status actions if `workflowConnectionId` is provided, improving the control over status change side-effects.
- Updated the related test case in `proposals.cy.ts` to cover the new functionality.

## How Has This Been Tested?

Added e2e

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5434](https://jira.ess.eu//browse/SWAP-5434)


apps/backend/src/mutations/ProposalMutations.ts
```diff
@@ -546,51 +546,60 @@ export default class ProposalMutations {
     agent: UserWithRole | null,
     args: ChangeProposalsStatusInput
   ): Promise<Proposals | Rejection> {
-    const { workflowStatusId: statusId, proposalPks } = args;
+    const {
+      workflowStatusId: statusId,
+      proposalPks,
+      workflowConnectionId,
```
yoganandaness: Is it redundant to have both? workflowConnectionId and workflowStatusId?

yoganandaness (changes requested):

apps/backend/src/mutations/ProposalMutations.ts
jekabs-karklins: Actually, no we need both, the one is referring to the node, but the other to the edge

jekabs-karklins: If the workflowConnectionId is speciffied, then it will execute the status actions, if not then no status actions will be run . I realised this was a little confusing so I changed the variable name and added a little comment with this commit
https://github.com/UserOfficeProject/user-office-core/pull/1441/changes/bb03de39a7d8fa6e52515697ca7f9fa0522f7ed1

jekabs-karklins:
Hi, @mutambaraf when you have a bit of time could you check this PR. it is not urgent, just has been in the queue for a while :). Thanks

jekabs-karklins:
@gnyiri I ran into the issue you mentioned [here](https://teams.microsoft.com/l/message/19:rbRjSbIAOCCDh_gBJLMud7DxrEAnqAQOyjZFEBSm5rU1@thread.tacv2/1780910467880?tenantId=3f66361c-a87e-4158-8f61-99e82db3cac8&groupId=f29c28ff-1d98-4762-bf7f-3c7cbd55ba37&parentMessageId=1780910467880&teamName=User%20Office%20Collaboration&channelName=General&createdTime=1780910467880).

The problem: when a connection has status actions configured

image

and the proposal's status is changed manually, an exception is thrown, so the status actions never run. (Automatic status updates work fine.)

The cause: the code on develop assumes a workflow connection id is always set. That's true for automatic updates, but not for manual status changes — there's no single connection to point to. In fact, two different connections can lead to the same status:

image

so the engine has no way of knowing which connection's actions to run.

Good news: this PR already fixes it. [This line](https://github.com/UserOfficeProject/user-office-core/pull/1441/changes#diff-d2e55d75eeebca944dca6d1be21ae2b1fd2049ba4435e764d691c1c8a310d7d9R557) skips the status actions when no connection is specified, and the PR also lets the user choose which connection to use for the manual case. Would be great to get this merged.

gnyiri:
Two comments:
- is it possible to also display the status actions along with the states (just like for the proposal workflow panel)?
- as far as I understand, the actual status does not restrict the list of selectable transitions, that is, all transitions can be selected that are pointing to the new status independently from the actual status. Is this the desired behaviour?

jekabs-karklins:
> Two comments:
> 
> * is it possible to also display the status actions along with the states (just like for the proposal workflow panel)?
> * as far as I understand, the actual status does not restrict the list of selectable transitions, that is, all transitions can be selected that are pointing to the new status independently from the actual status. Is this the desired behaviour?

I have added the events and actions to be displayed when changing the proposal status manually..
image

You can only select status actions that lead to the status that is selected. This is to enable the behavior how itworkflows were functioning in the past.

yoganandaness (approved):
lgtm. 

@jekabs-karklins I hope I should be doing the same for Experiment safety status change as well.

mutambaraf (commented):
Looks like we have dropped a test file `experimentSafetyWorkflow.spec.ts` was it by mistake or we will add them back

jekabs-karklins:
> Looks like we have dropped a test file `experimentSafetyWorkflow.spec.ts` was it by mistake or we will add them back

Certainly a mistake. Thanks I'll put it back.

apps/frontend/src/components/proposal/ChangeProposalStatus.tsx
```diff
@@ -126,116 +200,233 @@ const ChangeProposalStatus = ({
...
-                      loading={loadingProposalStatuses}
-                      items={proposalStatuses.map((status) => ({
-                        value: status.workflowStatusId,
-                        text: status.status.name,
-                      }))}
-                      required
-                      disabled={isSubmitting}
-                      data-cy="status-selection"
+                <Grid item xs={12} md={8}>
+                  <div style={{ height: '500px', border: '1px solid #ddd' }}>
```
mutambaraf: I do not know is it a good idea to hard code some of these colours.

apps/frontend/src/components/proposal/ChangeProposalStatus.tsx
```diff
@@ -57,6 +80,38 @@ const ChangeProposalStatus = ({
...
+        workflowId: selectedProposalsWorkflowIds[0],
+        entityType: WorkflowType.PROPOSAL,
+      })
+      .then((data) => {
+        if (data.workflow?.connections) {
+          setConnections(data.workflow.connections);
+        }
+      });
+  }, [api, selectedProposalsWorkflowIds[0]]);
+
```
mutambaraf: This looks like a duplication we may need to remove one of the useEffect hook.

mutambaraf (commented):
Looks good just a few comments

apps/frontend/src/components/proposal/ChangeProposalStatus.tsx
jekabs-karklins: I removed the border, I think it looks better without actually

apps/frontend/src/components/proposal/ChangeProposalStatus.tsx
jekabs-karklins: Thanks

mutambaraf (approved):

PR #1465, opened 2026-04-08
feat: only show instruments and calls on tags for proposal reader role

ellen-wright:
## Description
refs: https://github.com/UserOfficeProject/issue-tracker/issues/1562

This PR makes it so that only calls and instruments on tags attached to the specific proposal reader role are shown in the proposal filters. 

## Motivation and Context
Currently all calls and instruments are shown even if a user does not have access to them through the tags attached tot eh proposal reader role. 

## Changes
Adds two new graphQL queries 
-   getInstrumentsByRole
-   getCallsByRole

These are called from the instrument and call hooks when logged in on the proposal reader role and returns only the calls and instruments that are assigned to any tag attached to the role. 
If there are no tags on a role or you are using the root proposal reader role then all calls and instruments are returned, similarly to user officer. 


apps/backend/src/resolvers/queries/CallsQuery.ts
```diff
@@ -110,4 +110,11 @@ export class CallsQuery {
   callsOfReviewer(@Ctx() context: ResolverContext) {
     return context.queries.call.getCallsOfReviewer(context.user);
   }
+
+  @Query(() => [Call])
+  callsByRole(@Ctx() context: ResolverContext) {
+    const t = context.queries.call.getTagCallsByRoleId(context.user);
+
+    return t;
```
mutambaraf: We can remove the variable declaration here.

apps/backend/src/datasources/mockups/RoleDataSource.ts
```diff
@@ -51,6 +55,12 @@ export class RoleDataSourceMock implements RoleDataSource {
       isRootRole: false,
     } as Role;
   }
+  async getCallsbyRoleId(agent: number): Promise<Call[]> {
```
mutambaraf: We can use camelCase here : getCallsByRoleId

apps/backend/src/datasources/postgres/RoleDataSource.ts
```diff
@@ -143,4 +159,56 @@ export default class PostgresRoleDataSource implements RoleDataSource {
       throw error;
     }
   }
+
+  async getCallsbyRoleId(agent: number): Promise<Call[]> {
```
mutambaraf: We can improve naming here to getCallsByRoleId

apps/backend/src/datasources/postgres/RoleDataSource.ts
```diff
@@ -143,4 +159,56 @@ export default class PostgresRoleDataSource implements RoleDataSource {
...
+      }
+    } catch (error) {
+      logger.logError('Failed to get calls by role id', {
+        agent,
+      });
+      throw error;
+    }
+  }
+
+  async getInstrumentsbyRoleId(agent: number): Promise<Instrument[]> {
```
mutambaraf: We can improve naming to : getInstrumentsbyRoleId

apps/backend/src/datasources/postgres/RoleDataSource.ts
```diff
@@ -143,4 +159,56 @@ export default class PostgresRoleDataSource implements RoleDataSource {
...
+    try {
+      const tags = await this.getTagsByRoleId(agent);
+      const tagIds = tags.map((tag) => tag.id);
+      if (tagIds.length != 0) {
+        const instruments = await database<InstrumentRecord>(
+          'tag_instrument as fi'
```
mutambaraf: Maybe we can name it to `ti` here.

apps/frontend/src/hooks/call/useCallsData.ts
```diff
@@ -51,17 +59,28 @@ export function useCallsData(
...
-        setCalls(data.calls as Call[]);
-      }
-      setLoadingCalls(false);
-    });
+        if (data.calls) {
+          setCalls(data.calls as Call[]);
+        }
+        setLoadingCalls(false);
+      });
+    }
```
mutambaraf: I think in the backend we technically do not care if a user has  `UserRole.PROPOSAL_READER` so do not know if we need the if switch here. Lets just add the filters `callsFilter, ...callsQueryParams` to `getCallsByRole` endpoint and use it for all calls.

apps/frontend/src/hooks/call/useCallsData.ts
```diff
@@ -51,17 +59,28 @@ export function useCallsData(
...
+        if (unmounted) {
+          return;
+        }
+        if (data.callsByRole) {
+          setCalls(data.callsByRole as Call[]);
+          setLoadingCalls(false);
+        }
+      });
+    } else {
+      getCalls({ filter: callsFilter, ...callsQueryParams }).then((data) => {
```
mutambaraf: If we are getting the calls by roles do we still need the if statement here we may just need to pass in the filters.

apps/frontend/src/hooks/instrument/useInstrumentsMinimalData.ts
```diff
@@ -43,6 +43,19 @@ export function useInstrumentsMinimalData(callIds?: number[]): {
...
+          }
+
+          if (data.instrumentByRole) {
+            setInstruments(data.instrumentByRole);
+          }
+          setLoadingInstruments(false);
+        });
     } else {
       api()
         .getMyInstrumentsMinimal()
```
mutambaraf: I have added a comment above.

apps/backend/src/datasources/postgres/RoleDataSource.ts
```diff
@@ -143,4 +159,56 @@ export default class PostgresRoleDataSource implements RoleDataSource {
...
+        agent,
+      });
+      throw error;
+    }
+  }
+
+  async getInstrumentsbyRoleId(agent: number): Promise<Instrument[]> {
+    try {
+      const tags = await this.getTagsByRoleId(agent);
+      const tagIds = tags.map((tag) => tag.id);
```
mutambaraf: The tags variable can be handled this way, If getTagsByRoleId(agent) returns null or undefined, the subsequent .map() call will throw a runtime error. Consider defaulting to an empty array to ensure safety: const tags = (await this.getTagsByRoleId(agent)) ?? [];."

mutambaraf (commented):
Just some comments looks good.

apps/frontend/src/hooks/call/useCallsData.ts
ellen-wright: This separation is because currently the proposal reader is the role that changes depending on what tags are assigned

apps/frontend/src/hooks/call/useCallsData.ts
ellen-wright: This is because the get calls and instruments by tag isn't needed for User Officers and other roles so it's not necessary to alter how the filters are populated for them

mutambaraf (approved):
LGTM

jekabs-karklins (commented):
question: will the proposal reader role will be able to get all calls by calling the previous call getCalls, say buy crafting the GraphQL request? If yes, I'm therefore thinking if it isn't better to simply modify the existing getCalls and getCallsMinimal and filter out calls that proposal reader is not part of in the backend query. Or was it what @mutambaraf already asked?

ellen-wright:
> question: will the proposal reader role will be able to get all calls by calling the previous call getCalls, say buy crafting the GraphQL request? If yes, I'm therefore thinking if it isn't better to simply modify the existing getCalls and getCallsMinimal and filter out calls that proposal reader is not part of in the backend query. Or was it what @mutambaraf already asked?

Yeah I agree, I have removed the new queries and added them to the original call and instrument queries

jekabs-karklins (approved):

PR #1466, opened 2026-04-08
fix: frontend validation for title and abstract on partial save

simonfernandes:
## Description
This adds frontend validation for the title and abstract fields when using the "Save" button (partial save).

It also aligns the update and import schemas with the existing model and Yup validation schema, by changing the title and abstract to required fields. This change shouldn't break anything as they would previously have failed without these fields, but it's more explicitly defined now.

I couldn't find a nicer way to handle all of this, but I think it's reasonable to specifically validate the title and abstract on a partial save when they are considered to be core, required fields elsewhere.

## Motivation and Context
Previously, we didn't validate the title or abstract when doing a partial save. It would reach the backend mutation to update the proposal and immediately fail because the Yup validation schema specifies that the title and abstract are required. At STFC we would regularly see these backend errors and would like to eliminate them.

## Changes
1. Added mandatory validation for 'title' and 'abstract' fields in the backend.
2. Updated the frontend to handle these validation errors and provide feedback to the user.
3. Updated relevant tests to reflect these changes.

## How Has This Been Tested?

Manual testing and e2e

## Fixes Jira Issue

https://github.com/UserOfficeProject/issue-tracker/issues/1548


deepaksftc (approved):

jekabs-karklins:
question: is there a reason why you decided to make title and abstract mandatory fields even for the partial save, rather than relaxing Yup rules and allowing users to save draft proposals without both fields being mandatory?

jekabs-karklins (approved):
LGTM, but there is one failng e2e

simonfernandes:
Thanks for your review @jekabs-karklins 

>question: is there a reason why you decided to make title and abstract mandatory fields even for the partial save, rather than relaxing Yup rules and allowing users to save draft proposals without both fields being mandatory?

My thinking was that making it mandatory in the frontend was the more gentle solution. Since the backend already validated these fields, we have seemingly always been requiring these fields and never had a problem from developers or users in doing so. I thought that changing the schema and Proposal model would be more involved and possibly have more implications.

PR #1468, opened 2026-04-08
chore: upgrade yup

yoganandaness:
## Description

Upgrade YUP to latest 1.7.1

## Motivation and Context

Recent bump [PR](https://github.com/UserOfficeProject/user-office-core/pull/1452) underwent build errors and this PR is to manually upgrade with appropriate fixes 

## How Has This Been Tested

Ensured all the tests passed

## Changes

Upgraded YUP across frontend, backend and validation. 

## Tests included/Docs Updated?

- [ ] I have added tests to cover my changes.
- [ ] All relevant doc has been updated

PR #1480, opened 2026-04-15
feat: add max and min options to number input

### LINKED ISSUE #252: As a user I should be allowed to select the min/max number of weeks access dependant on my access route
We need to put a limit on what LSF and Artemis can request in weeks, so that they can only request 1-4 weeks and not over 4.

For HPL, we also need a limit as they should be able to request max 10 weeks.

TCMeldrum:
Closes UserOfficeProject/issue-tracker#252

## Description
This PR introduces min and max options to the number input fields in the application.

## Motivation and Context
The change was required to provide more flexibility and control over the input fields, enabling the application to handle a wider range of data validation scenarios and improve the user experience.

## Changes
1. Added new fields (`numberMin`, `numberMax`, `numberMinInclusive`, `numberMaxInclusive`) to the `NumberInputConfig` class, which control the minimum and maximum allowed values in number input fields.
2. Adjusted the `createNumberInputQuestion` function to handle these new options.
3. Updated the `QuestionNumberForm` component to include new form fields for the new options, complete with validation logic.
4. Updated test cases to reflect the new changes and ensure they are working as expected.

## Fixes Jira Issue


Scott-James-Hurley (approved):

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
```diff
@@ -66,10 +66,25 @@ export const QuestionNumberForm = (props: QuestionFormProps) => {
               unit: Yup.string(),
             })
           ),
+          numberMin: Yup.number()
+            .typeError('Minimum score must be a number')
```
jekabs-karklins: suggestion: I think this should be "Value must be a number"

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
```diff
@@ -66,10 +66,25 @@ export const QuestionNumberForm = (props: QuestionFormProps) => {
...
+            .nullable(),
+          numberMax: Yup.number()
+            .typeError('Maximum score must be a number')
```
jekabs-karklins: Ditto

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
```diff
@@ -66,10 +66,25 @@ export const QuestionNumberForm = (props: QuestionFormProps) => {
...
+            .nullable()
+            // We cannot do the same for numberMin because it would create a circular dependency
+            .when('numberMin', (numberMin, schema) => {
+              return numberMin !== undefined
+                ? schema.moreThan(
+                    numberMin,
+                    'Maximum must be strictly greater than Minimum'
```
jekabs-karklins: suggestion: I think technically if the values are inclusive then min should be allowed to be the same as max 

Image

jekabs-karklins (changes requested):

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
TCMeldrum: Thanks! Going to wait until #1468 is in to fix this as would like to use latest feature of yup

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionTemplateRelationNumberInputForm.tsx
```diff
@@ -38,6 +38,35 @@ export const QuestionTemplateRelationNumberForm = (
               unit: Yup.string(),
             })
           ),
+          numberMin: Yup.number()
+            .typeError('Minimum score must be a number')
```
jekabs-karklins: suggestion: Should we also call this "Value must be a number" the same as for numberMax test?

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
```diff
@@ -186,6 +215,60 @@ export const QuestionNumberForm = (props: QuestionFormProps) => {
...
             />
+            <Field
+              name="config.numberMin"
+              label="Minimum"
+              id="config.numberMin"
+              type="number"
+              component={TextField}
+              fullWidth
+              inputProps={{ 'data-cy': 'numberMin' }}
+              onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
```
jekabs-karklins, line 226: suggestion: I noticed that when the value for minimum/maximum is cleard it hides the checkbox, which keeps the state. It is just maybe little illogical to have min set as null and it being inclusive? Maybe we should clear the flag/checkbox (inclusive) in that case? See the recording.

https://github.com/user-attachments/assets/55b88a26-7b8e-4739-8895-f0b7221a3604

jekabs-karklins (approved):
I am approving as it looks good, but I added two small comments still please check them.

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionNumberInputForm.tsx
TCMeldrum: updated it to reset to false on the blur event when min is cleared

apps/frontend/src/components/questionary/questionaryComponents/NumberInput/QuestionTemplateRelationNumberInputForm.tsx
TCMeldrum: missed chnaging that one thanks!

PR #1491, opened 2026-04-21
feat: move question title into dialog title bar and remove "generic template"

GrantDLS:
## Description

This PR addresses some duplication in the dialog box that contains a generic template. It moves the "question title" and the subsequent answer label into the dialog title bar and removes it from the dialog body.

## Motivation and Context

As shown in the example below, "Generic Template" is shown in the title bar of the dialog, and the question name is shown in the body.

image<br />

Once the answer has been created, when editing it, the answer label is duplicated in the title bar and in the body:

image<br />

With this change, when creating a new answer or editing an existing one, "Generic Template" is no longer shown since this has no meaning to the end user.

image

Further to this, when reviewing an answer, the question is now in the dialog title bar, rather than the text "GenericTemplate details".

image

## How Has This Been Tested

Manual testing has been carried out.

## Changes

Only frontend changes are present in this PR - all within the components/questionary directory.

N/A


koanarec:
I do see the value of this PR when it comes to users seeing the generic template. But actual real users I assume wouldn't see the generic template, so would anyone in production actually use the showTitle=False flag?

What would "generic template" be substituted for in the real world?

Edit: the answer is that you users can see generic templates which is not ideal

koanarec:
If you were to instead optionally hide the title={selectedGenericTemplate?.title} field, then the format of this template would match the other ones and it would result in less code changes. And you'd achieve the same effect of showing the fields you want to show.

joeshannon:
There is also the "GenericTemplate details" which appears when opening an item from the review screen:
Screenshot From 2026-04-22 16-47-50


Perhaps should be changed as part of the PR too?

GrantDLS:
I have updated this PR, to implement Zach's suggestion of keeping the heading within the dialog body and just removing "Generic Template" from the dialog title bar. I agree this is more consistent with the rest of the system where there is a genuine need for both text in the title bar and within the body. It also results in fewer code changes.

I've also addressed Joe's point about "GenericTemplate details" showing when reviewing a question answer by adding the question text in the title bar instead.

I've updated the description and added a screenshot to show how this looks.

zacharyhankin (approved):
Looks good now the code changes are very minimal.

TCMeldrum (approved):

TCMeldrum:
Just in case you dont know when you creat a pr it title has to follow this regex: https://github.com/UserOfficeProject/user-office-core/blob/db8cbac6f1dafb14b8403366da2d3ae19b1888d7/.github/pr-title-checker-config.json#L7 I think while that job is failing is the capital 'M' lol. Also when you merge use swash and merge so develop and master has a cleaner commit history. 

Sorry if you knew this already!

GrantDLS:
Thank you @TCMeldrum! I couldn't figure out why it was failing on the PR title :-)

I'll make sure I use squash and merge.

Thanks again :+1:

bolmsten (approved):

PR #1493, opened 2026-04-22
feat: add external homepage button

### LINKED ISSUE #1566: Add button to external homepage
Buttont to go back to a external homepage

ChongJiaChua:
Closes UserOfficeProject/issue-tracker/issues/1566
## Description

Added a button within the account menu which allows users to navigate back to the external homepage. The destination URL can be managed via the EXTERNAL_AUTH_HOMEPAGE_URL key in App Settings.  

## Motivation and Context

Users struggled when attempting to navigate from the proposal system back to the external homepage.

## How Has This Been Tested

Added a Cypress e2e test to verify that the "Facilities Homepage" button successfully renders in the account menu and displays the expected text.


apps/backend/db_patches/0209_AddExternalHomepageSetting.sql
```diff
@@ -0,0 +1,13 @@
+DO
+$$
+BEGIN
+	IF register_patch('AddExternalHomepageSetting.sql', 'ChongJiaChua', 'Adding external facility homepage url to settings', '2026-04-16') THEN
+	BEGIN
+        INSERT INTO settings (settings_id, settings_value, description) 
+        VALUES ('EXTERNAL_AUTH_HOMEPAGE_URL', 'https://users.facilities.rl.ac.uk/', 'External Facility Homepage URL')
```
TCMeldrum: I think we should leave this blank in the db patch and set the link within: apps/backend/src/config/stfc/configureSTFCEnvironment.ts

TCMeldrum (commented):
Sorry forgot to tell you that you need to add the link in here as well https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/backend/docker-compose-stfc.e2e.yml this should fix the tests

edit linked the wrong place: https://github.com/UserOfficeProject/user-office-core/blob/57002cf494254c292baf23c4e980ce06e57f85a3/.github/workflows/test-build.yml#L448

TCMeldrum (approved):

jekabs-karklins (approved):
💯

jekabs-karklins:
This looks good! However, I think we should make this feature more generic. If we store arbitrary external links in a database table, we can allow user officers to configure them as needed.

image

apps/e2e/cypress/e2e/appSettings.cy.ts
```diff
@@ -81,6 +81,31 @@ context('App settings tests', () => {
...
+
+      cy.get('[placeholder="Search"]')
+        .clear()
+        .type('EXTERNAL_AUTH_HOMEPAGE_URL');
+
+      cy.contains('tr', 'EXTERNAL_AUTH_HOMEPAGE_URL')
+        .find('[aria-label="Edit"]')
+        .click();
+
+      cy.get('[actions=""] > :nth-child(4)')
```
jekabs-karklins: sugggestion: Can we not rely on the order but either make sure settings input fields has a `data-cy` selectors or at least use cy query selector that involves the EXTERNAL_AUTH_HOMEPAGE_URL

something like cy.contains('EXTERNAL_AUTH_HOMEPAGE_URL').closest('TR').get(<text_input_selector>)...

jekabs-karklins (changes requested):
Can we make the cy test little more resillient?

ChongJiaChua:
> This looks good! However, I think we should make this feature more generic. If we store arbitrary external links in a database table, we can allow user officers to configure them as needed.
> 
> image

Sorry I'm having trouble following.

The current implementation allows for each facility to define their own URL for the 'Facilities Homepage' button. The EXTERNAL_AUTH_HOMEPAGE_URL is stored in the settings table and is initially empty. User Officers from different facilities can then go to Configuration Settings -> App settings and update that URL specifically for their facility.

Is that what you were looking for?

jekabs-karklins (approved):

apps/e2e/cypress/e2e/appSettings.cy.ts
ChongJiaChua: Yeah I agree. I've updated the test to target the input field by locating the specific row containing EXTERNAL_AUTH_HOMEPAGE_URL and then finding the text input within that row

jekabs-karklins:
> > This looks good! However, I think we should make this feature more generic. If we store arbitrary external links in a database table, we can allow user officers to configure them as needed.
> > image
> 
> Sorry I'm having trouble following.
> 
> The current implementation allows for each facility to define their own URL for the 'Facilities Homepage' button. The EXTERNAL_AUTH_HOMEPAGE_URL is stored in the settings table and is initially empty. User Officers from different facilities can then go to Configuration Settings -> App settings and update that URL specifically for their facility.
> 
> Is that what you were looking for?

No - I think the feature is good. I was just sharing an observation: here at ESS, we’ll also need links to external resources. So instead of limiting this to a single EXTERNAL_AUTH_HOMEPAGE_URL, I was suggesting we could have a more generic section where we can configure multiple useful links (both internal and external).

But that can be a separate PR once there’s a clear need 👍🏻

ChongJiaChua:
> > > This looks good! However, I think we should make this feature more generic. If we store arbitrary external links in a database table, we can allow user officers to configure them as needed.
> > > image
> > 
> > 
> > Sorry I'm having trouble following.
> > The current implementation allows for each facility to define their own URL for the 'Facilities Homepage' button. The EXTERNAL_AUTH_HOMEPAGE_URL is stored in the settings table and is initially empty. User Officers from different facilities can then go to Configuration Settings -> App settings and update that URL specifically for their facility.
> > Is that what you were looking for?
> 
> No - I think the feature is good. I was just sharing an observation: here at ESS, we’ll also need links to external resources. So instead of limiting this to a single EXTERNAL_AUTH_HOMEPAGE_URL, I was suggesting we could have a more generic section where we can configure multiple useful links (both internal and external).
> 
> But that can be a separate PR once there’s a clear need 👍🏻

Ah I see. thanks for clarifying

PR #1497, opened 2026-04-29
feat: add configuration, custom Auth and event handlers for Diamond Light Source

GrantDLS:
## Description

This PR is to add the necessary configuration files, custom authorization class and email event handler for Diamond Light Source to be able to use User Office Project.

## Motivation and Context

- apps/backend/src/auth/DlsUserAuthorization.ts
  - Our own custom authorization class that works the same as the `OAuthAuthorization` class that it extends, except for it excludes some user attributes being pulled from the OAuth provider.
- apps/backend/src/config/dls/configureDLSEnvironment.ts
  - This previously contained a helper method `getUASInstance()` which read the value of a `.env` variable, but this change has been removed after feedback from the collaboration.
- apps/backend/src/config/dependencyConfigDLS.ts
  - This is standard and just sets the dependency mappings
- apps/backend/src/eventHandlers/email/DLS/DLSEmailHandler.ts
  - This is our email handler in which we're currently handling the following 2 events: `PROPOSAL_SUBMITTED` and `PROPOSAL_CO_PROPOSER_INVITES_UPDATED`. The logic for building and sending each email type is separated into different files for readability.
- apps/frontend/public/images/diamond-white.svg
  - This is our logo for use in the header.

## How Has This Been Tested

A test has been written for the email handler to ensure the correct email template is used for each event that we're handling.

## Changes

There are only 2 changes to the core files of the code base:

- apps/backend/src/auth/OAuthAuthorization.ts
  - This class mostly meets our needs, but a small change is needed in `externalTokenLogin()`. I've extended the class, but needed access to the private method `getUserRole()`, so I've made it's access modifier `protected` rather than `private`.
- apps/backend/src/config/index.ts
  - A case has been added to the switch statement to load the DLS config file when it is set in the `.env` file.

N/A


apps/backend/src/auth/DLSUserAuthorization.ts
```diff
@@ -0,0 +1,116 @@
...
+        oidcSub: userId,
+      });
+
+      return updatedUser;
+    } else {
+      const newUser = await this.userDataSource.create(
+        (userInfo.title as string) ?? 'unspecified',
+        userInfo.given_name,
+        userInfo.family_name,
+        userInfo.given_name ?? '',
```
zacharyjhankin: Perhaps a little comment here for why we have given_name repeated just for future readability.

I understand its because we don't have access  to preferredname so use given_name instead

apps/backend/src/auth/DLSUserAuthorization.ts
zacharyjhankin, line 1: Possibly call it DLSUserAuthorization? 

https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN 
DLS would be an acronym not an abbreviation

apps/backend/src/config/dls/configureDLSEnvironment.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  return instance;
+}
+
+async function setDLSColourTheme() {
+  const db = container.resolve<AdminDataSource>(Tokens.AdminDataSource);
+
+  await db.waitForDBUpgrade();
+
+  await db.updateSettings({
```
zacharyjhankin: We could use await `Promise.all([async_function1, async_function2])  `  syntax to get them running all at once. Not sure if it matters here though. The awaits would block each other?

apps/backend/src/config/dls/configureDLSEnvironment.ts
```diff
@@ -0,0 +1,156 @@
...
+    settingsValue: 'true',
+  });
+
+  await db.updateSettings({
+    settingsId: SettingsId.PROFILE_PAGE_LINK,
+    settingsValue: getUASInstance() + '/#PersonalDetailsPlace:',
+  });
+}
+
+async function configureDLSEnvironment() {
```
zacharyjhankin: async function configureDLSEnvironment() {
await Promise.all([setDLSColourTheme(), 
enableDefaultDLSFeatures(),setTimezone, setDateTimeFormats, updateOIDCSettings]);
}

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
zacharyjhankin, line 1: DLSEmailHandler?

apps/backend/src/config/dependencyConfigDLS.ts
```diff
@@ -0,0 +1,146 @@
...
+
+mapClass(Tokens.MailService, SMTPMailService);
+
+mapValue(Tokens.FapDataColumns, FapDataColumns);
+mapValue(Tokens.FapDataRow, getDataRow);
+mapValue(Tokens.PopulateRow, populateRow);
+mapValue(Tokens.PopulateCallRow, callFapPopulateRow);
+
+//mapValue(Tokens.EmailEventHandler, skipEmailHandler);
+mapValue(Tokens.EmailEventHandler, dlsEmailHandler);
```
zacharyjhankin: Instead comment something like:

//DLSEmailHandler is dimond specific implementation of skipEmailHandler

apps/backend/src/config/dependencyConfigDLS.ts
```diff
@@ -0,0 +1,146 @@
...
+mapClass(Tokens.EmailTemplateDataSource, PostgresEmailTemplateDataSource);
+
+mapValue(Tokens.PostToMessageQueue, createPostToRabbitMQHandler());
+mapValue(Tokens.LoggingHandler, createSkipLoggingHandler());
+mapValue(Tokens.EventBus, createApplicationEventBus());
+mapValue(Tokens.ListenToMessageQueue, createListenToRabbitMQHandler());
+
+mapValue(Tokens.ConfigureEnvironment, configureDLSEnvironment);
```
zacharyjhankin, line 142: If we are capitalizing it here I think we can do it everywhere to keep it consistent

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
```diff
@@ -0,0 +1,277 @@
...
+  const questionaryDataSource = container.resolve<QuestionaryDataSource>(
+    Tokens.QuestionaryDataSource
+  );
+
+  const emailTemplateDataSource = container.resolve<EmailTemplateDataSource>(
+    Tokens.EmailTemplateDataSource
+  );
+
+  switch (event.type) {
+    case Event.PROPOSAL_SUBMITTED: {
```
zacharyjhankin: I think these cases should be their own functions. It just keeps going and going and going

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
```diff
@@ -0,0 +1,277 @@
...
+        day: 'numeric',
+        month: 'short',
+        year: 'numeric',
+      });
+
+      const allocationPeriod = `${shortDateFormat.format(call.startCycle)} - ${shortDateFormat.format(call.endCycle)}`;
+
+      let baseURL = process.env.BASE_URL || '';
+      if (baseURL.endsWith('/')) {
+        baseURL = baseURL.slice(0, -1);
```
zacharyjhankin: I think that this code is too forgiving and if the baseURL is missing or malformed we want to throw an uncaught error straight away so its obvious. 

throw new Error('Environment variable BASE_URL is required');
throw new Error('Environment variable BASE_URL cannot have trailing /');

What we are risking allowing base_URL to be malformed is that the code deploys properly and then later on it will break if somewhere else in code they forget to remove the trailing / and they use the BASE_URL variable. 

It seems too important to allow for it to be empty?

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
```diff
@@ -0,0 +1,277 @@
...
+        logger.logError('Email template not found', {
+          template,
+        });
+
+        return;
+      }
+
+      const invites = event.array;
+
+      for (const invite of invites) {
```
zacharyjhankin: for (const invite of event.array) {

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        const options = {
+          content: {
+            template: emailTemplate.id.toString(),
+          },
+          substitution_data: {
+            sender: inviter.preferredname + ' ' + inviter.lastname,
+            redeem_code: invite.code,
+            uos_instance: process.env.BASE_URL,
```
zacharyjhankin: See here you're not stripping the trailing / like you are above so that could introduce a bug if later people expect it to have been well formed

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
```diff
@@ -0,0 +1,277 @@
...
+      }
+
+      return;
+    }
+    case Event.PROPOSAL_CO_PROPOSER_INVITES_UPDATED: {
+      const template = 'co-proposer-invite';
+      const emailTemplate =
+        await emailTemplateDataSource.getEmailTemplateByName(template);
+      if (!emailTemplate) {
+        logger.logError('Email template not found', {
```
zacharyjhankin: Could we not throw an error here?

apps/backend/src/config/dls/configureDLSEnvironment.ts
GrantDLS: `waitForDBUpgrade()` waits for any DB migrations to be applied. I think it's probably safer to wait for it to finish before applying our settings.

With regards to applying the settings - yes, you're probably right. I can't see any issue with running them asynchronously. It basically creates the setting record in the DB if it doesn't exist, and then updates it. I'll try it out and test it :+1:

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
zacharyjhankin, line 1: Maybe it would be a good idea to add one or two unit tests, just to set a good precedent. I do see that there already exists tests for other parts

apps/backend/src/config/dls/configureDLSEnvironment.ts
zacharyjhankin: Sorry my comment wasn't clear, I was talking about the

"await db.updateSettings" calls

but absolutely might not make any difference so we can probably leave as-is

GrantDLS: It's a fair comment, even if the time saving is only minimal and it only runs on application startup, there's no reason not to. I've made all the settings updates and the enabling/disabling of features all run synchronously and it works fine.

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
GrantDLS: You're right that it shouldn't be empty, and it's a bit messy having this logic in here so for now I've followed the same logic as our `getUASInstance()` function and created one in our config class for getting the base url. These ought to be in some global helper class which we've discussed but I think they're fine in here for now.

But as to whether we throw an error I'm not sure. There are multiple scenarios:

1. We throw an error, and the user doesn't get an email on proposal submission.
2. The URL is malformed, so the user gets an email with a link that won't work.
3. The BASE_URL isn't set, and so it defaults to a sensible default (like uos.diamond.ac.uk). This way, production UOS should always work.

I think now that there's a default and we're using the function, I don't think any error checking is needed.

apps/backend/src/eventHandlers/email/dlsEmailHandler.ts
GrantDLS: Yes, that seems like a sensible thing to do. The application logs uncaught errors so we can throw here rather than logging.

yoganandaness:
@GrantDLS Is this PR ready to be reviewed? If so, could you add the label "review:please"

GrantDLS:
> @GrantDLS Is this PR ready to be reviewed? If so, could you add the label "review:please"

I've done quite a bit of refactoring since I created the PR. Just adding a couple of tests for the email handler and I'll add the label. Thank you, @yoganandaness .

apps/backend/src/eventHandlers/email/DLS/DLSEmailHandler.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  let callDataSourceMock: CallDataSourceMock;
+  let questionaryDataSourceMock: QuestionaryDataSourceMock;
+  let inviteDataSourceMock: InviteDataSourceMock;
+
+  beforeAll(() => {
+    container.registerInstance(Tokens.MailService, mockMailService);
+  });
+
+  beforeEach(() => {
+    // Initialize mocks
```
zacharyjhankin: // Initialise mocks

GrantDLS: I've just removed the comment since it doesn't add anything :-)

apps/backend/src/eventHandlers/email/DLS/DLSEmailHandler.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        array: await inviteDataSourceMock.getInvites({} as GetInvitesFilter),
+        proposalPKey: 1,
+        key: 'test',
+        loggedInUserId: 1,
+      } as ApplicationEvent;
+
+      const expectedEmailTemplate =
+        await emailTemplateDataSourceMock.getEmailTemplateByName(templateId);
+
+      // Call the eliEmailHandler with the mock event
```
zacharyjhankin: I think this might have been a copy paste from eli? You mean dls

GrantDLS: Good spot. Removed.

apps/backend/src/eventHandlers/email/DLS/proposalCoProposerInvitesUpdatedHandler.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  const template = EmailTemplateId.CO_PROPOSER_INVITE;
+  const emailTemplate =
+    await emailTemplateDataSource.getEmailTemplateByName(template);
+  if (!emailTemplate) {
+    logger.logError('Email template not found', {
+      template,
+    });
+
+    return;
```
zacharyjhankin: This should throw an error, not just log it. Because we are already catching it in emailActionHandler, but we should always get an email template, if we don't its because someone has introduced a bug and we want to find that asap. Logging it and moving on is just adding potential for a bug to be introduced and never discovered.

apps/backend/src/eventHandlers/email/DLS/proposalSubmittedHandler.ts
```diff
@@ -0,0 +1,182 @@
...
+  participants.push(principalInvestigator); // Ensure PI also gets an email
+
+  for (const participant of participants) {
+    if (!participant.email) {
+      logger.logError(
+        'Could not send email on proposal submission: participant has no email',
+        { participant, event }
+      );
+
+      return;
```
zacharyjhankin: Shouldn't this be a continue so we send it to the next participant?

zacharyjhankin:
Can you add an image of the email PROPOSAL_CO_PROPOSER_INVITES_UPDATED?

apps/backend/src/auth/DLSUserAuthorization.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (userInfo.institution_ror_id) {
+      institutionInput = userInfo.institution_ror_id as string;
+    } else if (userInfo.institution_name && userInfo.institution_country) {
+      institutionInput = {
+        country: userInfo.institution_country as string,
+        name: userInfo.institution_name as string,
+      };
+    }
+
+    const institution = await this.getOrCreateUserInstitution(institutionInput);
```
zacharyjhankin: Move this to line 88 so its closer to where the value is consumed?

apps/backend/src/eventHandlers/email/DLS/proposalSubmittedHandler.ts
GrantDLS: Yes, well spotted! :+1:

apps/backend/src/auth/DLSUserAuthorization.ts
```diff
@@ -0,0 +1,116 @@
...
+    const userWithOAuthSubMatch =
+      await this.userDataSource.getByOIDCSub(userId);
+
+    const userWithEmailMatch = await this.userDataSource.getByEmail(
+      userInfo.email
+    );
+
+    const user = userWithOAuthSubMatch ?? userWithEmailMatch;
+
+    if (user) {
```
zacharyjhankin, line 42: I think we should just make a new createUser function in the class OAuthAuthorization and then use our own implementation for diamond so the code isn't repeated as much

zacharyjhankin: There is a lot more repeted code than we need

GrantDLS: I think it would be best to leave the OAuthAuthorization class alone for now. I get what you're saying and it would be a good change to make, especially for other institutions that want to extend the class but I think it would be best done as a separate PR.

apps/backend/src/auth/DLSUserAuthorization.ts
```diff
@@ -0,0 +1,117 @@
...
+
+import { OAuthAuthorization } from './OAuthAuthorization';
+import { User, UserRole } from '../models/User';
+import { GetOrCreateInstitutionInput } from '../resolvers/mutations/UpsertUserMutation';
+
+export class DLSUserAuthorization extends OAuthAuthorization {
+  constructor() {
+    super();
+
+    if (OpenIdClient.hasConfig()) {
```
zacharyjhankin: The super() method already runs this code, so you're running it twice?

apps/backend/src/auth/DLSUserAuthorization.ts
```diff
@@ -0,0 +1,117 @@
...
+    iss: string | null
+  ): Promise<User | null> {
+    try {
+      const { userProfile, tokenSet } = await OpenIdClient.login(
+        code,
+        redirectUri,
+        iss
+      );
+
+      const user = await this.upsertUserDLS(userProfile, tokenSet);
```
zacharyjhankin: I see you have re-written the externalTokenLogin function with a different call to upsertUserDLS, instead of  upsertUser. But you don't need to do this. Because if you write upsertUser defined in this child class then the superclass will first read the child classes implementation. Its way more code than we need.

GrantDLS: Good point. I've made `upsertUser` protected in the parent class so that we can override it.

apps/backend/src/auth/DLSUserAuthorization.ts
GrantDLS: Good catch. Fixed.

yoganandaness (changes requested):
Other the new env, all other looks good.

joeshannon:
> Other the new env, all other looks good.

Thanks for taking a look.

By new env, do you mean that the new DLS environment needs some additional config?

apps/backend/src/config/dls/configureDLSEnvironment.ts
```diff
@@ -0,0 +1,163 @@
...
+
+import { AdminDataSource } from '../../datasources/AdminDataSource';
+import { FeatureId } from '../../models/Feature';
+import { SettingsId } from '../../models/Settings';
+import { setTimezone, setDateTimeFormats } from '../setTimezoneAndFormat';
+import { Tokens } from '../Tokens';
+import { updateOIDCSettings } from '../updateOIDCSettings';
+
+function getUASInstance() {
+  let instance = process.env.UAS_INSTANCE || 'https://uas.diamond.ac.uk/uas';
```
yoganandaness: Is it possible to avoid this env and instead of that, set the settings values directly in the dashboard?

yoganandaness:
> > Other the new env, all other looks good.
> 
> Thanks for taking a look.
> 
> By new env, do you mean that the new DLS environment needs some additional config?


Is it possible to avoid this env(UAS_INSTANCE) and instead of that, set the settings values directly in the dashboard?

joeshannon:
> Is it possible to avoid this env(UAS_INSTANCE) and instead of that, set the settings values directly in the dashboard?

We could do, the idea with this was that this would point to different systems depending on the environment e.g. prod/staging but I suppose it could be set in the app.

Looking at it now we could probably adapt to use the existing env var that STFC have `process.env.PROFILE_PAGE_LINK` to derive it.

Do any facilities add custom settings to the `settings` table or should it be kept to those in the `SettingsId` enum?

yoganandaness:
> `process.env.PROFILE_PAGE_LINK`

@joeshannon 

Agree with you over the process.env.PROFILE_PAGE_LINK

Usually, SettingsId enum and settings table are in sync

yoganandaness:
@zacharyjhankin Just wanted to add in here that  since this has length commits, i would recommend to do a Squash and Merge, so the master and develop branch holds only one commit sha.

GrantDLS:
> > > Other the new env, all other looks good.
> > 
> > 
> > Thanks for taking a look.
> > By new env, do you mean that the new DLS environment needs some additional config?
> 
> Is it possible to avoid this env(UAS_INSTANCE) and instead of that, set the settings values directly in the dashboard?

Thanks for taking a look at this @yoganandaness . The reason we went with an environment variable over a settings value is because it's for something that's specific to Diamond. Having it as an environment variable means we can use it without polluting the `SettingsId` enum with things that are institution specific.

Further to this, it's not something we'd want a user to be able to change in the UI. We want it set at deployment only.

EDIT: I should also add, that the environment variable isn't required to be set by other institutions, since it's only used by the helper function defined in our configuration setup file.

yoganandaness:
> > > > Other the new env, all other looks good.
> > > 
> > > 
> > > Thanks for taking a look.
> > > By new env, do you mean that the new DLS environment needs some additional config?
> > 
> > 
> > Is it possible to avoid this env(UAS_INSTANCE) and instead of that, set the settings values directly in the dashboard?
> 
> Thanks for taking a look at this @yoganandaness . The reason we went with an environment variable over a settings value is because it's for something that's specific to Diamond. Having it as an environment variable means we can use it without polluting the `SettingsId` enum with things that are institution specific.
> 
> Further to this, it's not something we'd want a user to be able to change in the UI. We want it set at deployment only.
> 
> EDIT: I should also add, that the environment variable isn't required to be set by other institutions, since it's only used by the helper function defined in our configuration setup file.

If I understand your use case correctly, you don't want the user to set this value in the dashboard and want to get the value directly from env during the deployment. However, if you can see the settings page, PROFILE_PAGE_LINK is a part of the settings enum, it becomes automatically available to the User Officer to edit this value immaterial of if it comes from env or hardcoded. Correct me if I'm wrong. 

However, if you feel that you would like to set it during run time, we can make this env applicable to all the facilities as well. Could you rename the env (UAS_INSTANCE ) to PROFILE_PAGE_LINK, so that it aligned with the settings value.

GrantDLS:
Sorry, @yoganandaness I don't think I quite understand. I'll explain our use case some more.

The environment variable is only used in one place, for the following function inside `configureDLSEnvironment.ts`.  This function is currently just used once when sending out an invite to join a proposal (since we currently want users to sign up via UAS).

```
function getUASInstance() {
  let instance = process.env.UAS_INSTANCE || 'https://uas.diamond.ac.uk/uas';
  if (instance.endsWith('/')) {
    instance = instance.slice(0, -1);
  }

  return instance;
}
```

I can see a couple of issues with using the `PROFILE_PAGE_LINK` variable. Firstly, I'd have to use a regex to extract just the domain - not a huge issue but seems unnecessary. Secondly, there might be a rare case where we want the UAS_INSTANCE and the PROFILE_PAGE_LINK to point to different domains.

Unless I'm missing something with how the application interprets environment variables, the way we've done it currently seems like the most straightforward and unobtrusive way. It doesn't affect any of the other collaborators instances or require any changes.

zacharyjhankin:
From the diamond side right now, we do not want UAS_INSTANCE to be configurable at run time, so do not want to use PROFILE_PAGE_LINK. So we hope to add UAS_INSTANCE as a environment variable as the PR currently stands.

yoganandaness:
@zacharyjhankin As we discussed over the call, we recommend trying with the existing env variable PROFILE_PAGE_LINK. Let us know, if you have any issues. 

Thanks

GrantDLS:
Thank you for taking the time to have a call with us this morning @yoganandaness 

We've decided to just hard code our UAS domain for now. This removes the need for the environment variable and the helper.

yoganandaness (approved):

zacharyjhankin (approved):
lgtm

PR #1504, opened 2026-05-08
chore: update mock server to latest

William-Edwards-STFC:
## Description

Mock server certificates are failing and there is an update to fix it https://github.com/mock-server/mockserver-monorepo/pull/2094 which is out on this https://hub.docker.com/r/mockserver/mockserver/

## Motivation and Context

Currently the e2e tests are failing due to mock-uows not handling certificates correctly and this update will fix them

## How Has This Been Tested

This will be tested in the PR e2e tests


Scott-James-Hurley:
Looks like the e2e tests are still running into `EXTERNAL_TOKEN_INVALID`

mutambaraf (approved):

mutambaraf:
LGTM but for this is it worthy having our own image with locked in dependences because we do not make lots of changes to the mock server its self.

TCMeldrum (approved):

PR #1507, opened 2026-05-11
feat: ms graph api mail service

### LINKED ISSUE #1471: Microsoft will discontinue SMTP basic auth
For those who are using SMTP mail service with Microsoft Exchange. 

https://learn.microsoft.com/en-us/exchange/clients-and-mobile-in-exchange-online/deprecation-of-basic-authentication-exchange-online

This issue should cover the required implementation.

gnyiri:
## Description

Add new MS Graph API based mail service to replace SMTP mail service. 

## Motivation and Context

At ELI, the SMTP mail service will be deprecated soon. 

## How Has This Been Tested

Manual + backend tests. 

## Fixes

n/a

## Changes

Add MSGraphMailService class. 

## Depends on

n/a

- [y] I have added tests to cover my changes.
- [] All relevant doc has been updated


apps/backend/src/eventHandlers/MailService/MSGraph/MSGraphMailService.ts
```diff
@@ -0,0 +1,177 @@
...
+      auth: {
+        clientId: process.env.MS_GRAPH_API_CLIENT_ID!,
+        clientSecret: process.env.MS_GRAPH_API_CLIENT_SECRET!,
+        authority: `${process.env.MS_GRAPH_API_AUTHORITY}/${process.env.MS_GRAPH_API_TENANT_ID}`,
+      },
+    });
+  }
+
+  protected isTokenExpired(): boolean {
+    if (!this.authToken?.expiresOn) {
```
jekabs-karklins, line 36: bug (maybe): I think this should return 'true'? If the field is missing then intuitively the token should be considered as expired... or?

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
```diff
@@ -1,51 +1,26 @@
...
-      }
-    }
+    logger.logInfo('Initializing SMTPMailService', {
+      host: process.env.EMAIL_AUTH_HOST,
+      port: process.env.EMAIL_AUTH_PORT,
+      user: process.env.EMAIL_AUTH_USERNAME,
+    });
+  }
 
+  protected createTransport(): NodeMailerTransportOptions | any {
```
jekabs-karklins: suggestion: do we need to have 'any' here? I think we can just remove it.

apps/backend/src/eventHandlers/MailService/MSGraph/MSGraphMailService.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    return Date.now() >= this.authToken.expiresOn.getTime();
+  }
+
+  private async getAccessToken(): Promise<string> {
+    if (this.authToken && !this.isTokenExpired()) {
+      return this.authToken.accessToken;
+    }
+
+    this.authToken = await this.msalClient.acquireTokenByClientCredential({
```
jekabs-karklins: question: do you know what happens to the previous issued token, is that invalidated? The reason I'm asking is lets say there are two invocations of the function. both will await the credential, and once complete will update the class variable. If the second requiest invalidates the first requeests token then there could be a small race condition chance which will result usage of the token that is not valid.

jekabs-karklins (approved):

jekabs-karklins:
@gnyiri as far as I rememeber you were using SMTP mail service so we can close this PR? and reopen if ever MS Graph API becomes a needed again?

PR #1508, opened 2026-05-11
fix: move frontend client names from metric labels to logs

Bhaswati1148:
## Description
This PR aims to reduce metric cardinality by moving UOP frontend client names from Prometheus/Mimir metric labels to logs.

## Motivation and Context
Previously, all client names were added to the `client_name` metric label, which significantly increased time series cardinality due to the dynamic and high-volume nature of frontend clients. This contributed to slower Grafana dashboard loads and increased query execution time in Mimir.
Internal API clients are a fixed and controlled set of values, so retaining them in metric labels should not significantly increase cardinality.

### DIFF
--- a/apps/backend/src/middlewares/metrics/apolloServerMetricsPlugin.ts
+++ b/apps/backend/src/middlewares/metrics/apolloServerMetricsPlugin.ts
@@ -4,6 +4,7 @@ import {
   GraphQLRequestContext,
   GraphQLRequestContextWillSendResponse,
 } from '@apollo/server';
+import { logger } from '@user-office-software/duo-logger';
 import { GraphQLError, GraphQLFormattedError } from 'graphql';
 import { Counter, Histogram } from 'prom-client';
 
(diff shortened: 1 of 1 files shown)


Scott-James-Hurley (approved):

apps/backend/src/middlewares/metrics/apolloServerMetricsPlugin.ts
```diff
@@ -53,12 +54,21 @@ export const apolloServerMetricsPlugin = (): ApolloServerPlugin => ({
...
+        operation_Type: operationType,
+      });
+    } else apiClient = clientName;
 
     const labels: MetricLabels = {
       operation: operationName,
       operation_type: operationType,
       status: 'success',
-      client_name: clientName,
+      client_name: apiClient,
```
YufanKambang: My understanding is that when client is True with name starting `UOP frontend`, then the client name at the `MetricLabels` is unknown, causing Mimir metrics to be categorised in unknow .
Is this on intentional?

Bhaswati1148: Yes, that was intentional to reduce the high-cardinality impact caused by dynamic frontend client names in Mimir metrics.

I’ve updated the implementation so that frontend requests are now grouped under a stable metric label value (client_name="frontend") instead of unknown, while the exact frontend client name is preserved in Loki logs for debugging and analysis.

YufanKambang (approved):

PR #1523, opened 2026-05-19
fix: add role guard to user officer proposals table

### LINKED ISSUE #990: As a user Proposals route is loading and empty ProposalTableOfficer table.
While logged in as a user role the `/Proposals` is loading an empty `ProposalTableOfficer` component .There is need to reroute this route either to the correct user proposals table or back to the main dashboard.

### Steps to reproduce (if it's a bug).
While logged in with the user role go to  route `/Proposals`

ChongJiaChua:
Closes UserOfficeProject/issue-tracker/issues/990
## Description

Currently, unauthorized roles could access the `ProposalTableOfficer` component by navigating to the `/Proposals` route manually. This change makes it such only the `User Officer` and `Proposal Reader` roles are allowed to access it. Any unauthorized roles attempting to access the `/Proposals` route directly will be redirected back to the main dashboard (`/`)

## Changes
Frontend AppRoutes.tsx: 
- Added conditional guards based on user roles (`isUserOfficer` || `isProposalReader`)
- Added a fallback route for `/Proposals` that redirects unauthorized roles back to `/`

### DIFF
--- a/apps/frontend/src/components/AppRoutes.tsx
+++ b/apps/frontend/src/components/AppRoutes.tsx
@@ -238,10 +238,16 @@ const AppRoutes = () => {
           }
         />
 
-        <Route
-          path="/Proposals"
-          element={<TitledRoute title="Proposals" element={<ProposalPage />} />}
-        />
+        {isUserOfficer || isProposalReader ? (
+          <Route
+            path="/Proposals"
+            element={
+              <TitledRoute title="Proposals" element={<ProposalPage />} />
+            }
+          />
+        ) : (
+          <Route path="/Proposals" element={<Navigate to="/" replace />} />
+        )}
         {isTechniqueProposalsEnabled &&
           (isInstrumentScientist || isUserOfficer) && (
             <Route


YufanKambang (dismissed):

simonfernandes (approved):

YufanKambang:
The client side for the code can still be affected. Not sure if this is needed but making sure that the at the API layer that supplies data to `/Proposals` page, also check for the correct roles of the user.

This exists here: https://github.com/UserOfficeProject/user-office-core/blob/045dd10e4d73ac76c049b0e774d044e3c23c6674/apps/backend/src/queries/ProposalQueries.ts#L104
it is defended

YufanKambang (approved):

ChongJiaChua:
> The client side for the code can still be affected. Not sure if this is needed but making sure that the at the API layer that supplies data to `/Proposals` page, also check for the correct roles of the user.
> 
> This exists here:
> 
> https://github.com/UserOfficeProject/user-office-core/blob/045dd10e4d73ac76c049b0e774d044e3c23c6674/apps/backend/src/queries/ProposalQueries.ts#L104
> 
> 
> it is defended

Yeap, the backend already prevented unauthorised users from accessing the data so this PR just stops them from navigating to the page.

PR #1541, opened 2026-05-27
feat: add baseURL option for dynamic http question

zacharyjhankin:
## Description

This adds an option to dynamically set the baseURL of the dynamic question to the domain that is currently deployed. NOTE: needs http:// in the beginning of your BASE_URL environment variable. (this PR updates example.env and docker-compose.yml)

## Motivation and Context

We are using a graphQL endpoint to obtain the list of countries that the users can select as their country. This will always use the domain that the UOC is deployed on, and different environments will have different domains. This lets the code dynamically call its own domain. We also don't want users getting confused. 

## How Has This Been Tested

I made a dynamic http question in the develop branch and checked that in the proposal it would work to make the countries. using: 

http://localhost:3000/graphql?query=query%20ExampleQuery%20%7B%20countries%20%7B%20value%20%7D%20%7D

$.data.countries[*].value

 content-type:application/json

then I switched to my branch and ran npm run generate:sdk

I confirmed my conversion worked to add baseURL to the dynamic question.

I confirmed that the existing question would still work without changing anything in a new proposal

I then edited my proposal template HTTP question to use my baseURL setting.

I confirmed that the country list would still be shown in the proposal.

image

image

image

image

NOTE:

There is an existing bug that means the URL that the proposal uses when someone is filling it out to get the countries is the parent question. Not the question that is edited for the specific template. This means you may encounter this error when testing. So make sure to edit both the parent question and the template question when testing this to see it actually being used.

Added backend unit test


apps/backend/src/queries/TemplateQueries.ts
```diff
@@ -140,3 +141,11 @@ export default class TemplateQueries {
     return [];
   }
 }
+
+function constructDynamicURL(url: string, useBaseURL: boolean): string {
+  if (useBaseURL) {
+    return `http://${process.env.BASE_URL}/${url}`;
```
GrantDLS: I wonder how much work we should do here to deal with how the base url is set? It's a bit like when we were setting the UAS_INSTANCE environment variable - trailing slash, already includes http or https?

In most application, a double forward slash would usually not break anything (so mydomain.com//my-endpoint would still go to mydomain.com/my-endpoint), so that's not much of an issue, so it might just be that we don't include the http bit.

apps/frontend/src/components/questionary/questionaryComponents/DynamicMultipleChoice/QuestionTemplateRelationDynamicMultipleChoiceForm.tsx
```diff
@@ -103,17 +146,94 @@ export const QuestionTemplateRelationDynamicMultipleChoiceForm = (
...
+              <div style={{ display: 'flex', gap: '8px' }}>
+                {useBaseDomain && (
+                  <span
+                    style={{
+                      whiteSpace: 'nowrap',
+                      color: 'grey',
+                      paddingTop: '20px',
+                    }}
+                  >
+                    {`http://${serverConfig}/`}
```
GrantDLS: Another bit where we potentially don't need the http part.

zacharyjhankin:
Updated to work with http:// in the base_url environment variable. Will not work if your BASE_URL is just localhost:3000

ellen-wright (approved):

GrantDLS (approved):

apps/backend/src/queries/TemplateQueries.ts
zacharyjhankin: Thanks for the comment, that is a very good spot. The standard way for what is meant by "base_url" would include the HTTP, but in the existing .env file there is not http. I updated my code to depend on http being include in the BASE_URL and changed the example .env. I left a comment in the general chat so people know of this change. @ellen-wright  is aware too.

The code now reads 

`${process.env.BASE_URL}/${url}`;

So neither the baseURL of the URL path should contain a trailing or leading slash. I have that in my regex to show users and make it obvious.

apps/frontend/src/components/questionary/questionaryComponents/DynamicMultipleChoice/QuestionTemplateRelationDynamicMultipleChoiceForm.tsx
zacharyjhankin: Has been removed now

PR #691, opened 2024-08-07
feat: add FAP review templates

gnyiri:
## Description
This PR introduces a new feature that adds FAP review templates to the application.

## Motivation and Context
The addition of FAP review templates is a necessity for the project. This feature aids in the creation, management, and use of FAP review templates, enhancing the versatility and functionality of the system.

## Changes
1. A new SQL script has been added (`0158_AddFAPReviewTemplate.sql`) that creates new template categories, template groups, question data types, and questions related to FAP reviews. It also modifies the `fap_reviews` and `call` tables to include the new `questionary_id` and `fap_review_template_id` fields respectively.
2. Changes have been made to `QuestionaryAuthorization.ts` and a new file `FapReviewQuestionaryAuthorizer.ts` has been created to handle authorization for FAP review templates.
3. The `dependencyConfigELI.ts` file has been created to handle dependency injection for FAP review templates.

## Fixes Jira Issue

[https://jira.esss.lu.se/browse/](https://jira.esss.lu.se/browse/)

## Tests included/Docs Updated?

- [x] I have added tests to cover my changes.
- [x] All relevant doc has been updated

PR #1552, opened 2026-06-01
fix: remove extra row in questionaries in inital db setup

zacharyjhankin:
## Description

When you start the database from scratch and apply all of the seeds and patches you are left with one row in the questionaries table. This row contains a FK reference to user 0, which is the service account. We want to delete this account when starting a new instance of UOP, and we get an error. The questionaries table row can't be easily? deleted in the gui. So this just makes initial startup easier. 

Error that is shown when you try to delete the service user in the gui after startup.
screenshot-1

This row is added here:
[this](https://github.com/UserOfficeProject/user-office-core/pull/691) pull request, apps/backend/db_patches/0158_AddFAPReviewTemplate.sql line 45.

picture of DB just after starting up
image

## How Has This Been Tested

Locally ran though starting the database and tested deleting the service account with my user in it. 
image

## Changes

Adds a patch to delete one row from questionaries.

### DIFF
--- a/apps/backend/db_patches/0212_removeQuestionariesEntryFromSeed.sql
+++ b/apps/backend/db_patches/0212_removeQuestionariesEntryFromSeed.sql
@@ -1,0 +1,14 @@
+DO
+$$
+BEGIN
+	IF register_patch('RemoveQuestionariesEntryFromSeed.sql', 'Zachary Hankin', 'Removes entry in questionaries that is left over from initialising db', '2026-06-01') THEN
+	BEGIN
+        DELETE FROM questionaries
+        WHERE questionary_id = 1
+        AND template_id = 2
+        AND creator_id = 0;
+    END;
+	END IF;
+END;
+$$
+LANGUAGE plpgsql;


gnyiri:
As I wrote in my comment on Teams, this change might delete the only (default) fap review template. Let us wait if other developers share their thoughts on that on Teams.

joeshannon:
For additional context this row in `questionaries` appears to be responsible for this:
Screenshot From 2026-05-28 17-34-53

Which is indicating that there is 1 use of the template but clicking on it results in an empty result list.

As far as I can see the other template types: proposal,  technical review, are initialised with 0 proposals/reviews.

jekabs-karklins:
@zacharyjhankin I am little puzzled on when you say "We want to delete this account when starting a new instance of UOP, and we get an error." Are you talking about prod environment starting it up for the first time or some dev/test environment that is always recreated?

For the PROD environment normally we have transient backend service connecting to non-transient database instance, meaning that database is instantiated once, and it is more than okay to modify the DB to get it into the state manually.

Since we have already had some data in the datbase the added patch would not do anything because the questionaire id is 402.
image

zacharyjhankin:
@jekabs-karklins 

I am specifically talking about when you clear the entire database in dev/test envs, and run a "npm run dev". All the seeds and patch files run, and you have some data to play around with. Which is good, except this one row in the questionaries table that shouldn't be there. Because it shows that user 0 created something that uses the FAP template, but they have not. This row in the questionaries table only exists because it is manually added to the DB in one patch file: apps/backend/db_patches/0158_AddFAPReviewTemplate.sql line 45.

I don't want to touch anyones production environment. 



An alternative approach would be to delete line 45 in apps/backend/db_patches/0158_AddFAPReviewTemplate.sql so this line is never created. Maybe that is less risky, but I didn't want to change any of the existing patch files as that felt like bad practice.

jekabs-karklins:
Thank you for clarifying then I think the patch is good.

jekabs-karklins (approved):

gnyiri (approved):

PR #1553, opened 2026-06-01
fix: allow instrument contacts to filter by their assigned experimental areas

### LINKED ISSUE #1526: Instrument contact does not have that instrument come up on the tech review page
When someone is assigned as a instrument contact the instrument filter on the proposals page does not include that instrument, we should check if there are other places the instrument contact is not checked while fetching proposals/instruments

ChongJiaChua:
Closes UserOfficeProject/issue-tracker/issues/1526
## Description
This PR allows "instrument contacts" to have the correct visibility and access to their assigned experimental areas without being assigned as an "instrument scientist" as well.

## Motivation and Context

When a user is assigned as an "instrument contact" but not explicitly as an "instrument scientist", they were unable to see their assigned instruments in any of the experimental area filters.

## How Has This Been Tested

Added e2e tests.

## Changes
- Updated the backend database queries to correctly include instruments where the user is listed as the instrument contact.
- Fixed an issue with how the database filters were being combined to prevent other active filters from overriding the user's access rights.


apps/e2e/cypress/e2e/instruments.cy.ts
```diff
@@ -1885,4 +1885,131 @@ context('Instrument tests', () => {
       cy.should('not.contain', proposal1.title);
     });
   });
+
+  describe.only('Instrument contact visibility tests', () => {
```
jekabs-karklins: breaking: We should remove the "only"

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
```diff
@@ -588,15 +588,25 @@ export default class PostgresExperimentDataSource
...
+        .leftJoin(
           'instrument_has_scientists',
           'experiments.instrument_id',
           'instrument_has_scientists.instrument_id'
         )
-        .where(
-          'instrument_has_scientists.user_id',
-          filter.instrumentScientistUserId
-        );
+        .join(
```
jekabs-karklins, line 604: There is something wrong with this query I get duplicate rows back if I sign in as an instrument scientist

```
query DemoInstrumentContactDuplication {
  allExperiments(filter: {}, first: 20, offset: 0) {
    totalCount
    experiments {
      experimentPk
      experimentId
      instrumentId
      proposal { proposalId title }
    }
  }
}
```

Image

jekabs-karklins (changes requested):

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
mutambaraf: Also it is nice to have a test for when user is both a scientist AND a manager on different instruments

apps/e2e/cypress/e2e/instruments.cy.ts
ChongJiaChua: Thanks for spotting that

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
ChongJiaChua: @jekabs-karklins Good catch. The issue was that the query joined all scientist rows for an instrument, so when the user was the instrument contact, the same experiment was returned once per joined scientist. I have fixed it now.

ChongJiaChua: @mutambaraf I have added a test case for when a user who is an instrument contact on one instrument and an instrument scientist on another.

jekabs-karklins (approved):

PR #1563, opened 2026-06-08
feat: shift as an option for allocation time period

zacharyjhankin:
## Description

This PR adds a shift as a new unit for calls. This ability is vital for diamond as we allocate beam time from the synchrotron based on 8 hour units, 3 per day. All existing proposals base the scientists requests for beam time in shift units, so this allows us to be consistent with the existing infrastructure. It also prevents requests for 3 hours. 

The way I have coded this allows other organisations to customise the shift unit to be any length of time they want. So it could be useful to them.

## Motivation and Context

At diamond we use shift as a time unit for our calls which is 8 hours. This PR allows us to have shifts and lets other companies edit it. 

## How Has This Been Tested

image

image
I set the time period to shift, then edited the code to execute getProposalMessageData when a proposal is submitted. Then verified that it correctly calculates the time. from the shift time period as you can see above. I manually set the allocated time to 1 so it multiplies 1 by the shift time in seconds. 
This is actually the only place in code that references how long an allocation period is in reality.

## Changes

A new allocation period unit.


apps/backend/src/config/base/allocationTimeUnitConverter.ts
```diff
@@ -0,0 +1,24 @@
+import { AllocationTimeUnits } from '../../models/Call';
+
+export type AllocationTimeUnitConverter = (
+  time: number,
+  unit: AllocationTimeUnits
+) => number;
+
+export const getSecondsFromAllocationTimeUnits: AllocationTimeUnitConverter = (
```
jekabs-karklins, line 8: suggestion: maybe it is a good idea to call the filename the same as the exported function

apps/backend/src/config/base/allocationTimeUnitConverter.ts
```diff
@@ -0,0 +1,24 @@
...
+  timeAllocation: number,
+  unit: AllocationTimeUnits
+) => {
+  // NOTE: Default AllocationTimeUnit is 'Day'.
+  // 'Shift' is configurable per organisation; it defaults to 8 hours here.
+  switch (unit) {
```
jekabs-karklins, line 15: nitpick: adding a quick unit test would be nice

apps/backend/src/config/base/getSecondsFromAllocationType.ts
```diff
@@ -0,0 +1,24 @@
...
+  // NOTE: Default AllocationTimeUnit is 'Day'.
+  // 'Shift' is configurable per organisation; it defaults to 8 hours here.
+  switch (unit) {
+    case AllocationTimeUnits.Hour:
+      return timeAllocation * 60 * 60;
+    case AllocationTimeUnits.Week:
+      return timeAllocation * 7 * 24 * 60 * 60;
+    case AllocationTimeUnits.Shift:
+      return timeAllocation * 60 * 60 * 8;
+    default:
```
jekabs-karklins:
```suggestion
    case AllocationTimeUnits.Day:
    default:
```
Can we have this? so it's more intutitve by looking at the code what is the default?

apps/backend/src/config/base/getSecondsFromAllocationType.ts
```diff
@@ -0,0 +1,24 @@
...
+  time: number,
+  unit: AllocationTimeUnits
+) => number;
+
+export const getSecondsFromAllocationTimeUnits: AllocationTimeUnitConverter = (
+  timeAllocation: number,
+  unit: AllocationTimeUnits
+) => {
+  // NOTE: Default AllocationTimeUnit is 'Day'.
+  // 'Shift' is configurable per organisation; it defaults to 8 hours here.
```
jekabs-karklins:
```suggestion
// NOTE: Shift duration defaults to 8 hours. 
// To customize this per organization, inject a configuration provider 
// that supplies the specific shift duration for the facility.
```

jekabs-karklins (commented):
Just some minor comments

apps/backend/src/config/base/getSecondsFromAllocationType.ts
zacharyjhankin: Good suggestion

apps/backend/src/config/base/getSecondsFromAllocationType.ts
zacharyjhankin: Good idea, I didn't actually know that syntax

apps/backend/src/config/base/allocationTimeUnitConverter.ts
zacharyjhankin: I changed the file name to allocationTimeUnitConverter.ts‎ because it can export two objects so I didn't want to call it exactly after the specific implementation of the converter. In the future then we might add more implementations for converting these units, as different instruments might want to convert them differently.  So this way we're flexible in the future.

apps/backend/src/config/base/allocationTimeUnitConverter.ts
zacharyjhankin: I added a .test.ts file which seems to have run with the pipeline and passed. Though I am not completely sure I did it correctly. I didn't know the difference between a .spec. file and a .test. file.

jekabs-karklins (approved):
👍🏻

apps/backend/src/config/base/allocationTimeUnitConverter.ts
mutambaraf: I think should be a .spec

mutambaraf: I think we use https://github.com/UserOfficeProject/user-office-core/blob/2c761157ae4653cb7f62e956357415ebe4bf683b/apps/backend/jest.unit.config.js#L11 to pick the tests,

mutambaraf (approved):
Looks good have added some few comments

apps/backend/src/config/base/allocationTimeUnitConverter.ts
zacharyjhankin: I changed it to spec, though could you explain what the .test. files are? There seems to be lots of them in the code, I assume they would run in the pipelines too?

PR #1564, opened 2026-06-09
Use template config override for dynamic multi

GrantDLS:


apps/frontend/src/components/common/proposalFilters/QuestionaryFilter.tsx
```diff
@@ -159,6 +163,7 @@ function QuestionaryFilter({ onSubmit, callId }: QuestionaryFilterProps) {
               }}
               questionTemplateRelation={selectedQuestion}
               callId={callId}
+              templateId={templateId}
```
zacharyjhankin, line 166: I don't think we can do this. What I am thinking is that in QuestionTemplateRelationFragment there is reference to templateId though not in dynamicHTTP type. So maybe we could add the templateID here?

GrantDLS: Yeah, QuestionTemplateRelationFragment is a generated type, so I'll figure out how it's generated and see where we need to add the template id.

PR #1571, opened 2026-06-12
fix: remove the ability to change proposal status for proposer reader

ellen-wright:
## Description
refs: #https://github.com/UserOfficeProject/issue-tracker/issues/1608

## Motivation and Context
Maintains the proposal reader role as read only 

## Fixes
Removes access to the modal to change proposal status for the read only role 


simonfernandes (approved):

jekabs-karklins (approved):
Looks good, I can see the backend is already preventing the call just frontend that was showing the ability

PR #1391, opened 2026-03-02
chore: user office experiment workflow

yoganandaness:
## Description

This PR implements comprehensive workflow improvements for experiment safety management and integrates the new workflow data structures across the application. It includes support for manual status changes, enhanced workflow engine with guards, and updates to frontend components for better workflow status handling.

## Motivation and Context

The experiment safety workflow needs to support more sophisticated status transitions and manual interventions similar to the proposal workflow. This requires integrating the new workflow data structures (from SWAP-4949) into the experiment safety module and implementing proper authorization checks, event handling, and UI components for managing experiment safety workflow statuses.

**Related Issue:** SWAP-5222

## How Has This Been Tested

- Unit tests updated and fixed for new workflow structures
- E2E tests updated for:
  - Experiment safety review workflows
  - Instrument settings with new workflow statuses
  - Settings workflow management
- Integration tests for workflow guards and state transitions
- Manual testing of experiment safety status changes
- Fixes applied to resolve flaky tests

### Backend Changes
- **Database Migrations:**
  - `0208_MigrateOldWorkflowsToNewDataStructures.sql` - Migrates existing workflows to new data structures
  - `0211_ExperimentSafetyWorkflowStatusNotNull.sql` - Makes experiment_safety.workflow_status_id non-nullable with AWAITING_ESF backfill

- **Workflow Engine & Guards:**
  - Added experiment workflow engine (`src/workflowEngine/experiment.ts`)
  - Implemented new guards for experiment ESF status checking
  - Enhanced workflow state machine with proper transitions

- **Data Sources:**
  - Updated `ExperimentDataSource` to support workflow status management
  - Enhanced `WorkflowDataSource` for experiment safety workflows
  - Updated mock implementations for testing

- **Event Handlers:**
  - Implemented experiment workflow event handlers
  - Added email and RabbitMQ handlers for experiment status changes
  - Added status action engine for experiment workflows
  - Enhanced proposal workflow event handlers

- **Authorization:**
  - Updated `ExperimentSafetyAuthorization` for workflow-based access control
  - Enhanced authorization checks in workflow operations

- **Mutations:**
  - Added `ChangeExperimentsSafetyStatusMutation` for manual status changes
  - Enhanced experiment and proposal mutations

### Frontend Changes
- **Experiment Safety Components:**
  - `ChangeExperimentSafetyStatus.tsx` - New component for manual status changes
[...]

PR #1574, opened 2026-06-15
feat: clone proposal workflows

### LINKED ISSUE #1530: Clone proposal workflows
Add a button to clone a proposals workflow

ChongJiaChua:
Closes UserOfficeProject/issue-tracker/issues/1530
## Description

This PR introduces a clone button to each workflow entry. Cloning a workflow creates a direct copy of it that can be opened and customised independently.
## Motivation and Context

It is complicated and time consuming to create a workflow from scratch. This change allows users to duplicate an existing workflow as a starting point and make the necessary changes in a shorter time frame. 
## How Has This Been Tested

Added unit and e2e tests. 
## Changes

1. Added a method that copies a workflow and all related records in a database transaction.
2. Added `cloneWorkflow` mutation with role-based access control and input validation.
3. Added a new GraphQL resolver and validation schema to support the `cloneWorkflow` mutation.
4. Added a clone action button to the workflows table with a confirmation dialog.
5. Added a `cloneWorkflow` GraphQL mutation query on the frontend.


TCMeldrum (approved):
lgtm tho we might want to wait for https://github.com/UserOfficeProject/user-office-core/pull/1391 to go in first before we merge

apps/backend/src/datasources/postgres/WorkflowDataSource.ts
```diff
@@ -176,6 +177,160 @@ export default class PostgresWorkflowDataSource implements WorkflowDataSource {
...
+      for (const connection of sourceConnections) {
+        const newPrevStatusId = statusIdMap.get(
+          connection.prev_workflow_status_id
+        );
+        const newNextStatusId = statusIdMap.get(
+          connection.next_workflow_status_id
+        );
+
+        if (newPrevStatusId === undefined || newNextStatusId === undefined) {
+          continue;
```
jekabs-karklins: suggestion: can we throw here? This should never happen, but if happens is an indication of an issue somewher upstream and would be nice to know about it so we can fix it. Othwerise the clonde will appear as successful but in fact will have missing connections.

apps/e2e/cypress/e2e/settings.cy.ts
```diff
@@ -722,6 +722,20 @@ context('Settings tests', () => {
...
+      cy.visit('/');
+
+      cy.contains('Settings').click();
+      cy.contains('Proposal workflows').click();
+
+      cy.contains(workflowName).parent().find('[aria-label="Clone"]').click();
+
+      cy.contains('Yes').click();
+
+      cy.contains(`Copy of ${workflowName}`);
```
jekabs-karklins: suggestion: would it be possble to inspect also the contents of the cloned workflow? The real work in this PR is done in the database layer so changes cant effectivelly be targegeted by Unit Tests. So adding some checks here in e2e could effectivelly check if the DB operations went though fine.

apps/backend/src/datasources/postgres/WorkflowDataSource.ts
ChongJiaChua: Yeap good point. I've changed it to throw a GraphQLError so no incomplete clone gets saved.

apps/e2e/cypress/e2e/settings.cy.ts
ChongJiaChua: I've updated the test to set up a status and connection before cloning, then open the cloned workflow and verify the metadata, statuses and connection edge are all present.

jekabs-karklins (approved):
LGTM

PR #1576, opened 2026-06-15
fix: column sort in experiments table

shivoomiess:
## Description
This PR fixes the issue of column sorting not working properly in the Experiments Table.

## Motivation and Context
The column sorting feature in the Experiments Table was not functioning as expected, leading to a less intuitive and potentially confusing user experience. This fix is necessary to ensure that users can sort data in a logical and predictable manner, improving the usability of the table.

## Changes
1. Debugging capabilities were added to React and the local backend to better facilitate future troubleshooting efforts.
2. The backend's `dev` script was updated to include the `--inspect` flag, allowing for easier debugging.
3. The `ExperimentDataSource.ts` file was updated to include a map of fields that can be sorted. This provides explicit control over which columns can have their sorting functionality enabled.
4. The `experiments.cy.ts` e2e test file was updated to include tests for the sorting functionality of each column in the visit table. This ensures that our sorting feature works as expected in the future.

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5630](https://jira.ess.eu//browse/SWAP-5630)


apps/backend/package.json
```diff
@@ -19,7 +19,7 @@
     "build:validation": "npm --prefix ../../validation ci && npm --prefix ../../validation run build",
     "build": "npm run build:validation && npm run generate:uows && rm -rf ./build && tsc",
     "postinstall": "npm run build:validation",
-    "dev": "docker compose up -d --remove-orphans && ts-node-dev -r tsconfig-paths/register --respawn -T ./index.ts",
+    "dev": "docker compose up -d --remove-orphans && ts-node-dev --inspect=9229 -r tsconfig-paths/register --respawn -T ./index.ts",
```
yoganandaness: Do we need to update readme about this? @jekabs-karklins

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
```diff
@@ -83,8 +83,17 @@ function generateExperimentId(
   return `${proposalNumber}-${sequence ?? 0}`;
 }
 
+// Which fields we are allowed to sort on
 const fieldMap: { [key: string]: string } = {
+  // These fields are in ExperimentsTable.tsx (frontend),
```
yoganandaness: would prefer the backend to be frontend agnostic.

apps/backend/src/datasources/postgres/ExperimentDataSource.ts
```diff
@@ -83,8 +83,17 @@ function generateExperimentId(
...
+  // refer Column[] and field
   experimentId: 'experiment_id',
+  'proposal.proposalId': 'proposal_id',
+  startsAt: 'starts_at',
+  endsAt: 'ends_at',
+  // We don't sort on the fields below, create a view later if sorting needed
```
yoganandaness: Would be nice to have an Enum of permissible sortField in the Resolver layer.

apps/e2e/cypress/e2e/experiments.cy.ts
yoganandaness, line 1: Would prefer having an e2e for Instrument Scientist and Experiment Safety Reviewer as well.

yoganandaness (changes requested):

apps/e2e/cypress/e2e/experiments.cy.ts
```diff
@@ -1,296 +1,392 @@
...
+      cy.get('span[aria-sort="descending"]').click();
+
+      cy.get(
+        'div[data-rfd-draggable-id="4"] [data-testid="mtableheader-sortlabel"]'
+      ).click();
+      cy.get('span[aria-sort="ascending"]').click();
+      cy.get('span[aria-sort="descending"]').click();
+    });
+
+    it('Columns in Experiments Table are sortable - Experiment Safety Reviwer', () => {
```
jekabs-karklins: typo: This should be Reviewer

apps/backend/package.json
jekabs-karklins: I think we should not add it here, because we already have the "debug" script. Does that work for you @shivoomiess

shivoomiess: I think that runs a separate instance of the frontend dev server while reusing existing docker containers? Haven't tried using the debug script separately; we could:
- either add a port number for the inspect in `debug`
- or make the dev server debuggable (should be fine since we're already in dev?)

jekabs-karklins: We can add 9229 to the `--inspect`, however it is the default port, but we can be explicit.

the "debug" script just starts the index.ts with the debug port open, you do need to run `docker compose up` before running `npm run debug`

yoganandaness: @shivoomiess FYI

apps/backend/src/resolvers/queries/ExperimentsQuery.ts
```diff
@@ -24,6 +25,18 @@ class TimeSpan {
...
 }
 
+export enum ExperimentTableSortField {
+  experimentId = 'experimentId',
+  proposalId = 'proposalId',
+  startsAt = 'startsAt',
+  endsAt = 'endsAt',
+}
+
+registerEnumType(ExperimentTableSortField, {
```
yoganandaness: Can we do this in apps/backend/src/resolvers/registerEnums.ts

apps/backend/src/resolvers/queries/ExperimentsQuery.ts
```diff
@@ -24,6 +25,18 @@ class TimeSpan {
   to?: Date;
 }
 
+export enum ExperimentTableSortField {
```
yoganandaness: Would prefer to have this in Experiment Model, so it can be used in other queries in the future. Let me know your thoughts.

yoganandaness (changes requested):

apps/e2e/cypress/e2e/experiments.cy.ts
```diff
@@ -1,296 +1,392 @@
...
+        'div[data-rfd-draggable-id="3"] [data-testid="mtableheader-sortlabel"]'
+      ).click();
+      cy.get('span[aria-sort="ascending"]').click();
+      cy.get('span[aria-sort="descending"]').click();
+
+      cy.get(
+        'div[data-rfd-draggable-id="4"] [data-testid="mtableheader-sortlabel"]'
+      ).click();
+      cy.get('span[aria-sort="ascending"]').click();
+      cy.get('span[aria-sort="descending"]').click();
```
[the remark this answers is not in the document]
shivoomiess: if there are any issues at that level, the button becomes unclickable

apps/backend/package.json
yoganandaness: Can we not have this change to stick to context?

yoganandaness (changes requested):
Requesting to rename to a branch, which holds the context of the problem statement

.zed/debug.json
yoganandaness, line 1: @shivoomiess Can this also be removed and added as new PR for Zed support.

yoganandaness (changes requested):

.zed/debug.json
shivoomiess: okay

yoganandaness (approved):

simonfernandes (approved):

jekabs-karklins (approved):

PR #1582, opened 2026-06-16
feat: use template config for dynamic multiple choice

GrantDLS:
## Description

Currently, when a dynamic multiple choice question is added to a proposal, the template level config options are being ignored, i.e. only the configuration values set on the question are used.

This is because, the component uses a hook `useGetDynamicMultipleChoiceOptions`, which takes a question ID.

This PR solves this by obtaining the template ID, and if it is present, passes it to the GraphQL query `getDynamicMultipleChoiceOptions`. Within this function, `getQuestionTemplateRelation` is then called with the question ID and template ID. If the template ID isn't present, it falls back to using `getQuestion` with just the question ID parameter.

It's worth noting that I tried to solve this by passing the config that's available to the react component through to the backend, but since the http headers are stripped for security purposes, this wasn't a viable solution.

## How Has This Been Tested

I've added a test to `TemplateQueries.spec.ts` that ensures it calls the `getQuestionTemplateRelation` method if the template ID is provided. I've also manually tested it.

## Changes

The changes flow from the front end (getting the template ID from the context within the react component), through to the hook, and then the GraphQL query.


apps/frontend/src/components/common/proposalFilters/QuestionaryFilter.tsx
```diff
@@ -159,6 +163,7 @@ function QuestionaryFilter({ onSubmit, callId }: QuestionaryFilterProps) {
               }}
               questionTemplateRelation={selectedQuestion}
               callId={callId}
+              templateId={templateId}
```
zacharyjhankin: I think this change has the same problem as the previous PR, are we sure we can't use the selectedQuestion object to pass the templateId in it?  If you check out the SDK definition of the object it already contains three references to templateId. 

I could be missing something if you think that isn't the right approach

GrantDLS: Yeah you're right - it would be good if it was part of that question template relation but I couldn't quite figure it out. Hopefully when someone has time to review the PR they might have an idea on how to achieve this.

jekabs-karklins: @GrantDLS I have raised a small PR to return the templateID and utilize in the `QuestionaryFilter.tsx`  #1611

jekabs-karklins (approved):
LGTM

apps/frontend/src/components/common/proposalFilters/QuestionaryFilter.tsx
GrantDLS: Thank you @jekabs-karklins . I've merged your PR into the branch for this PR.

PR #1583, opened 2026-06-16
fix: generic template title max length error

### LINKED ISSUE #1514: Generic template title max length error
There is no user friendly UI error for exceeding the max length in a generic template title

Image

mohamedhassanmn:
## Description

This PR improves error handling when creating a generic template by catching database errors caused by values exceeding the allowed field length and returning a more meaningful GraphQL error message.

## Motivation and Context

Previously, when attempting to insert a generic template with a value exceeding the database field limit, the backend returned the raw database error, making it difficult to understand the actual issue.

This change ensures that such cases are handled gracefully by detecting the specific database error and returning a cleaner and more understandable error message.

## How Has This Been Tested

Tested locally by:

- Attempting to create a generic template with input values exceeding the allowed database length.
- Verifying that the backend correctly catches the database error.
- Confirming that a user-friendly GraphQL error message is returned instead of the raw database error.

## Fixes

Fixes [#1514](https://github.com/UserOfficeProject/issue-tracker/issues/1514)

## Changes

- Added error handling in the generic template creation flow.
- Added a conditional check for database errors containing `value too long`.
- Return a custom `GraphQLError` with a clearer error message when field length validation fails.
- Preserved existing error handling behavior for all other errors.

## Depends on

No dependencies.


apps/backend/src/datasources/postgres/GenericTemplateDataSource.ts
```diff
@@ -125,6 +125,12 @@ export default class PostgresGenericTemplateDataSource
         return createGenericTemplateObject(records[0]);
       })
       .catch((err) => {
+        if (err.message.includes('value too long')) {
+          return new GraphQLError(
```
zacharyjhankin: Shouldn't this be:

"throw new GraphQLError("

?

zacharyjhankin: We want to throw the error, not return the error object. I have no idea what that would cause

mohamedhassanmn: Good catch, you’re right. Returning the error object here would resolve the promise instead of propagating the failure. I’ll update it to throw so the error is handled correctly upstream.

mohamedhassanmn: Updated this as suggested to throw the error from the datasource. I found that the resolver catch was overriding the custom `GraphQLError` with the generic rejection message, so I added a check to preserve `GraphQLError` instances and only wrap unexpected errors.

apps/backend/src/mutations/GenericTemplateMutations.ts
```diff
@@ -92,6 +93,10 @@ export default class GenericTemplateMutations {
         );
       })
       .catch((error) => {
+        if (error instanceof GraphQLError) {
+          throw error;
```
zacharyjhankin, line 97: I am not sure we can throw an error here. Because the previous code returns a rejection which doesn't propagate the error up the call stack. Throwing an error halts execution until its caught. Looks like the rejection already returns a string th user sees?

mohamedhassanmn: Throwing inside a .then() or .catch() callback just rejects the promise, it doesn't halt execution in a surprising way. The GraphQLError flows into the mutation's .catch(), where the instanceof check re-throws it so it reaches the GraphQL layer with the meaningful message intact. This is necessary because rejection() returns a Rejection object rather than throwing, so without the check, the GraphQLError would be caught and silently replaced by the generic rejection message, losing the 'value too long' context entirely.

mohamedhassanmn: Alternatively, to stay consistent with the existing pattern, we could handle it entirely through rejection():

.catch((error) => {
  if (error instanceof GraphQLError) {
    return rejection(error.message, { agent, args });
  }
  return rejection(
    'Can not create genericTemplate because an error occurred',
    { agent, args },
    error
  );
});

This avoids throwing altogether and keeps everything consistent with how other errors are handled in the mutation. Happy to go with whichever approach you prefer.

zacharyjhankin: Thank you very much for teaching me about this error handling. If you throw an error in a async function in JS then you're just rejecting the promise which is handled later. It's a good change so happy to let you make the call on syntax you want to use :)

zacharyjhankin (approved):
Good change

apps/backend/src/datasources/postgres/GenericTemplateDataSource.ts
```diff
@@ -125,7 +125,13 @@ export default class PostgresGenericTemplateDataSource
         return createGenericTemplateObject(records[0]);
       })
       .catch((err) => {
-        return err;
+        if (err.message.includes('value too long')) {
+          throw new GraphQLError(
+            'Failed to insert genericTemplate as the value is too long'
```
TCMeldrum: I think it might be more helpful to say something like `title too long` as a user does not know the name `genericTemplate` as it is never used in the frontend

mohamedhassanmn: That makes sense, thanks. I’ll update the error message to make it more user-friendly and avoid exposing the internal `genericTemplate` terminology.

TCMeldrum (approved):

PR #1584, opened 2026-06-16
feat: add new column for FAP review of time requested

zacharyjhankin:
## Description

This pull request adds the details of how much allocation time that a proposal requested so the FAP reviewers know how much time that they should allocate. 

image

## Motivation and Context

When you're doing a FAP review it is not clear how many hours that the proposal wants for an instrument so this PR makes it very obvious to them

## How Has This Been Tested

Manual testing with proposals in multiple states

- not entered the instrument time requested
- Has two instruments assigned
- has two instrument questions both with the same instrument and different times

## Changes

Key is adding a new graphql endpoint for requesting the time requested of a proposal

- [ ] All relevant doc has been updated. 
 No need for docs as its very intuative what the new info means.


GrantDLS:
Looks good to me :+1:

apps/backend/src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -312,6 +312,32 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
       });
   }
 
+  async getRequestedTime(
+    proposalId: number,
```
simonfernandes: This looks like it's meant to be proposalPk

simonfernandes (commented):
Nice feature 👍

apps/backend/src/datasources/postgres/ProposalDataSource.ts
zacharyjhankin: Oh yes, good catch thank you :)

simonfernandes (approved):

PR #1585, opened 2026-06-16
fix: add sort by enum to GQL endpoint

jekabs-karklins:
## Description
This PR introduces an improvement to the GraphQL sorting functionality by adding an enum for sortable fields.

## Motivation and Context
The change was needed to ensure type safety and allow for more precise sorting in the GraphQL endpoint.

## Changes
- Replaced the type of sortable fields from string to enum, which provides type safety.
- Registered the enum type in the ExperimentTableSortField.
- Updated the frontend to map the column field to the correct enum value.
- Updated the GraphQL query to accept the enum type for the sort field.

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5630](https://jira.ess.eu//browse/SWAP-5630)

### DIFF
--- a/apps/backend/src/datasources/postgres/ExperimentDataSource.ts
+++ b/apps/backend/src/datasources/postgres/ExperimentDataSource.ts
@@ -90,18 +90,17 @@ type SortableExperimentTableColumnName =
   | 'starts_at'
   | 'ends_at';
 
-// Which fields we are allowed to sort on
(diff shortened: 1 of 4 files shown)


yoganandaness (approved):
Good one. Should we do the same for other entities like Proposal ,People etc.,

shivoomiess (approved):
looks good

PR #1601, opened 2026-06-24
feat: technique proposals page use tags

### LINKED ISSUE #1554: Xpress feature enhancement - Implement Tags in Xpress flows
Given that the Tags framework was introduced with https://github.com/UserOfficeProject/user-office-core/pull/1068 and dynamic roles PR is merged. We need to start using the Tags. 

Confirmed with Tom earlier that R&A has not started using the tags yet and also have no plans of using the tags for any other purpose other than for differentiating facilities. 

- [ ] Analyse and decide on what all tags we need to configure
- [ ] Setup tags in dev and prod
- [ ] Update the Xpress flows to consider the tags as facility while applying logic

Scott-James-Hurley:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1554

## Description

If the user is logged in as a derived User Officer role with tags assigned to it, they will only see calls and instruments, and in turn, proposals and techniques, that are associated with those tags on the technique proposals page. They will also only be able to assign proposals to instruments that have the same tags as their derived role. This allows for siloed User Officers, who can only see and edit technique proposals for their facility.

## Motivation and Context

Another facility (Central Laser Facility) at STFC has expressed interest in using the technique proposals feature. A requirement is that ISIS and CLF User Officers should not be able to see entities (calls, instruments, proposals, etc) from the other facility. Currently, there is no way to limit what User Officers can see on the technique proposals page.

## How Has This Been Tested

Added E2E tests

## Fixes

Fix PR also fixes an issue where trying to create a derived role of certain roles would throw an error.


apps/backend/src/queries/InstrumentQueries.ts
```diff
@@ -40,8 +40,15 @@ export default class InstrumentQueries {
...
     if (!callIds || callIds.length === 0) {
-      return await this.dataSource.getInstruments();
+      return await this.dataSource.getInstruments(
+        undefined,
+        undefined,
+        agentRoleId
+      );
     } else {
       const instrumentsByCallIds =
         await this.dataSource.getInstrumentsByCallId(callIds);
```
zacharyjhankin, line 54: This might be a stupid question, but surely if .getInstrumentsByCallId( is called instead of .getInstruments( then your new filter on tag won't apply. Wouldn't this contradict the idea of the PR to limit what these users can see?

Scott-James-Hurley: From what I can see, `getInstrumentsByCallId` is not called when [populating the instrument filter](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx#L967) on the technique proposals page. The [call](https://github.com/UserOfficeProject/user-office-core/blob/develop/apps/frontend/src/components/techniqueProposal/useTechniqueProposalInstrumentsData.ts#L50) to fetch instruments doesn't contain a call id so it will take the path starting `line 47`, not `line 54`. Happy to corrected if I'm mistaken, though.

zacharyjhankin: I see you're right, but what about        

.getInstrumentsMinimal({ callIds }) in useInstrumentsMinimalData.ts,

 should that one also be using your filter?

apps/backend/src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -134,31 +134,54 @@ export default class PostgresInstrumentDataSource
...
+        .whereIn('ti.tag_id', tagIds)
+        .orderBy('i.instrument_id', 'desc')
+        .modify((query) => {
+          if (first) {
+            query.limit(first);
+          }
+          if (offset) {
+            query.offset(offset);
+          }
+        });
```
jekabs-karklins: breaking: If there is two roles AAA/BBB and both assigned to the derived root role and also both assigned to the instrument, then I get duplicates

```suggestion
      instruments = await database
        .select(['i.*', database.raw('count(*) OVER() AS full_count')])
        .from('instruments as i')
        .whereIn('i.instrument_id', function () {
          this.select('ti.instrument_id')
            .from('tag_instrument as ti')
            .whereIn('ti.tag_id', tagIds);
        })
        .orderBy('i.instrument_id', 'desc')
        .modify((query) => {
          if (first) {
            query.limit(first);
          }
          if (offset) {
            query.offset(offset);
          }
        });
```

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -85,11 +86,36 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
+      return techniques.flatMap((t) => t.instruments.map((i) => i.id));
+    }
+  }, [
+    loadingTechniques,
+    techniques,
+    loadingInstrumentsMinimal,
+    instrumentsMinimal,
+  ]);
+
+  const relevantTechniques =
```
jekabs-karklins, line 111: question: Do we also want to enforce the authorization in the backend?

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -427,6 +453,12 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
           instruments = [missingInst, ...instruments];
         }
 
+        if (currentRole === UserRole.USER_OFFICER) {
+          instruments = instruments.filter((i) =>
+            techniqueInstruments.map((ti) => ti.id).includes(i.id)
+          );
+        }
```
jekabs-karklins, line 460: suggestion: I am not that familiar with Technique proposals but have you tried givig tag to instrument, create technique propoal, and then remove the tag from the instrument. I wonder if you are not breaking whats added in :439

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
Scott-James-Hurley: Is this with regards to the role check on line `112`? If so, my intention there is determine whether techniques should be filtered by the instruments available to the user. If they're a user officer, filtering needs to be done as the method to fetch technique returns all of them - including ones not relevant to them if they're using a derived role. If they're a technique scientist, only the relevant techniques are returned by the backend - no filtering needed.

apps/backend/src/datasources/postgres/InstrumentDataSource.ts
Scott-James-Hurley: Great catch, thanks!

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
jekabs-karklins: Thanks for the explanation - so the filtering here is ux improvement? That makes sense.

One thing I'd flag: deriving the selectable set from the current instrument - technique associations feels a little fragile. If a technique is disassociated from an instrument later on, a User Officer would no longer be able to select it - which could block legitimate retrospective actions on proposals submitted while that association still existed.

Not blocking on this: if it's an requirement from UOS, I'm happy to keep it as is. Just worth being aware of the coupling or maybe just show all techniques

Scott-James-Hurley: Good point... I hadn't considered that. Could a solution be to include all techniques and instruments of proposals in the calls a derived role has access to instead? That should cast a wider net.

Scott-James-Hurley: With regards to relying on the instrument - technique association, I'm not sure how else it could be done. Techniques have no direct relationship with tags and all techniques can't be shown as that could expose techniques that should be hidden from the user.

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
Scott-James-Hurley: It's a good point and I think it highlights the topic of what tags should be seen as and how they should behave in UOP. At STFC, we are looking to use them as a proxy for facilities (e.g. ISIS and CLF), organisations that share infrastructure and functionality but whose data (calls, instruments, proposals) are entirely separate. In our case, I think it's unlikely that an instrument would have its tag removed. For other collaborators, this may be a scenario that needs to be considered. I'm not sure how the issue you highlight here could be addressed in a satisfactory manner right now. There's currently no concept of tag history for instruments or calls but maybe that would be helpful. I'd find having a chat about this really helpful, would you be up for that?

Scott-James-Hurley: I had a chat with @deepaksftc and I understand tags are generic groupings of calls and instruments. I'll try to make the logic more generic and account for scenarios like the one you highlighted here.

apps/backend/src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -134,31 +134,57 @@ export default class PostgresInstrumentDataSource
...
-      })
-      .then((instruments: InstrumentRecord[]) => {
-        const result = instruments.map((instrument) =>
-          this.createInstrumentObject(instrument)
-        );
+    let instruments: InstrumentRecord[];
+    const tags = agentRoleId
+      ? (await this.getTagsByRoleId(agentRoleId)) ?? []
+      : [];
+    if (tags.length > 0) {
```
simonfernandes: suggestion: could this if statement go inside the `.modify()`, so that the `whereIn()` is only executed when there are tags? That would avoid having two separate queries separated in an if/else

apps/backend/src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -134,31 +134,57 @@ export default class PostgresInstrumentDataSource
...
-        if (offset) {
-          query.offset(offset);
-        }
-      })
-      .then((instruments: InstrumentRecord[]) => {
-        const result = instruments.map((instrument) =>
-          this.createInstrumentObject(instrument)
-        );
+    let instruments: InstrumentRecord[];
+    const tags = agentRoleId
```
simonfernandes, line 140: suggestion: it might be a cleaner separation of concerns for this to be at query level, so that `tagIds` are passed into the datasource instead of `agentRoleId`.

simonfernandes (commented):
Just to be aware, technique proposals are currently included on the main User Officer proposals page, which I think undermines this feature. This was only meant as an initial thing to handle any teething issues and I believe it can safely be removed now (unless @deepaksftc can think of any reason why not). Sorry that you may end up having to do it as part of this but hopefully it's only a small change.

apps/backend/src/datasources/postgres/InstrumentDataSource.ts
Scott-James-Hurley: There's a few places where where this is done in datasources. If it's changed here, it should probably be changed in those instances as well. In order to keep this PR's scope down, I suggest I open an issue to address all instances. What do you think?

simonfernandes: Sure, it's only a minor thing.

Scott-James-Hurley: https://github.com/UserOfficeProject/issue-tracker/issues/1641

Scott-James-Hurley:
I think excluding Xpress proposals from the main proposals page warrants its own PR. That PR and this one will rely on each and will be marked as "do not merge" until they're both ready.

Scott-James-Hurley:
> I think excluding Xpress proposals from the main proposals page warrants its own PR. That PR and this one will rely on each and will be marked as "do not merge" until they're both ready.

We talked with the ISIS User Officer and are going to hold off on this as they use features on the main proposals page that aren't available on the technique proposals page.

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
Scott-James-Hurley: Apologies for taking so long to get back to you. Would you mind having a look again? The technical proposal's current instrument is now displayed, even when that instrument doesn't have the same tag as the user's derived role. Does that address the concern you raised?

jekabs-karklins: Looks better and nice comment

jekabs-karklins (approved):

simonfernandes (approved):

PR #1603, opened 2026-06-24
feat: date range picker in Visit Registration replaces two date pickers

shivoomiess:
## Description

Replaces the two separate "Visit start" and "Visit end" date pickers in the Visit Registration form with a single date-range picker. Selecting a range now populates both `startsAt` and `endsAt` from one calendar UI.

Screenshot 2026-06-24 at 13 27 23

## Motivation and Context

Registering a visit previously required interacting with two independent MUI date pickers, which was clunky and made it easy to pick an end date before the start date. A single range picker is more intuitive, keeps start/end selection in one place, and visually communicates the selected span. It also drops the MUI `@mui/x-date-pickers` / `LocalizationProvider` boilerplate that only existed to work around Cypress/Material-UI datepicker testing issues.

## Changes

1. Added a new reusable `FormikUIDayRangePicker.tsx` component built on `react-day-picker`, rendering the calendar inside an MUI `Popover` (so it isn't clipped by the modal) with a read-only text field, a calendar icon button, and a "Done" button. It supports `minDate`, custom Luxon display `format`, required/disabled/error states, and themed range styling.
2. Rewrote `QuestionaryComponentVisitBasis.tsx` to use the new `DayRangePicker` instead of the two `Field`/`DatePicker` instances wrapped in `LocalizationProvider`.
3. Added a `toDateTime` helper to normalize `startsAt`/`endsAt`, which arrive as Luxon `DateTime`s once edited but as ISO strings when a registration is loaded from the backend.
4. Wired Formik `touched`/`errors` (via `getIn`) into the range picker so start/end validation errors surface on the single field.
5. Removed the now-unused `useTheme` and `desktopModeMediaQuery` Cypress workaround.

## How Has This Been Tested?

Visually, we can see that the functionality works as intended. Since there are no changes on the backend and the frontend uses the same flow to send / receive data, we confirm the changes have been tested sufficiently.

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5412](https://jira.ess.eu//browse/SWAP-5412)


apps/frontend/src/components/common/FormikUIDayRangePicker.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const displayValue = value?.from
+    ? `${formatDate(value.from, format)} - ${formatDate(value.to, format)}`
+    : '';
+
+  const disabledDays: Matcher | undefined = minDate
+    ? { before: minDate.toJSDate() }
+    : undefined;
+
+  return (
+    <div style={{ display: 'flex', alignItems: 'flex-start' }}>
```
jekabs-karklins:
```suggestion
    <div style={{ display: 'flex' }}>
```

jekabs-karklins: I think this looks better without flex-start

apps/frontend/src/components/common/FormikUIDayRangePicker.tsx
```diff
@@ -0,0 +1,166 @@
...
+          label={label}
+          fullWidth
+          required={required}
+          disabled={disabled}
+          placeholder="dd/mm/yyyy - dd/mm/yyyy"
+          value={displayValue}
+          error={!!error}
+          helperText={error ?? helperText}
+          InputLabelProps={{ shrink: true }}
+          InputProps={{ readOnly: true }}
```
jekabs-karklins, line 121: suggestion: I think it feels more natural to open the date selector when user clicks on a date, rather that requiring to click the button. Especially because user can't edit the date by entering numbers with keyboard

apps/frontend/src/components/common/FormikUIDayRangePicker.tsx
```diff
@@ -0,0 +1,166 @@
...
+        aria-label="Open date range picker"
+        sx={{ marginLeft: 1, marginTop: 1 }}
+      >
+        <DateRangeIcon />
+      </IconButton>
+      <Popover
+        open={open}
+        anchorEl={anchorEl}
+        onClose={() => setAnchorEl(null)}
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
```
jekabs-karklins, line 137: suggestion: change required from comment on line :121

jekabs-karklins (approved):
LGTM

apps/frontend/src/components/common/FormikUIDayRangePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const displayValue = value?.from
+    ? `${formatDate(value.from, format)} - ${formatDate(value.to, format)}`
+    : '';
+
+  const disabledDays: Matcher | undefined = minDate
+    ? { before: minDate.toJSDate() }
+    : undefined;
+
+  return (
+    <div style={{ display: 'flex' }}>
```
yoganandaness: Can we use MUI box instead of div

<Box sx={{ display: 'flex' }}>

apps/frontend/src/components/common/FormikUIDayRangePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+      <div style={{ flex: 1 }}>
```
yoganandaness: Same here

yoganandaness (approved):

apps/frontend/src/components/questionary/questionaryComponents/VisitBasis/QuestionaryComponentVisitBasis.tsx
```diff
@@ -36,63 +51,32 @@ function QuestionaryComponentVisitBasis({ answer }: BasicComponentProps) {
...
+        from: toDateTime(state.registration.startsAt),
+        to: toDateTime(state.registration.endsAt),
+      }}
+      error={startsAtError || endsAtError || undefined}
+      onChange={({ from, to }: DayRange) => {
+        dispatch({
+          type: 'ITEM_WITH_QUESTIONARY_MODIFIED',
+          itemWithQuestionary: { startsAt: from, endsAt: to },
+        });
+      }}
```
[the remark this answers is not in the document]
shivoomiess: False Positive.
Refuted on every sub-claim:

- **Stale display?** No. `DayRangePicker`'s `value` reads `toDateTime(state.registration.startsAt)` straight from reducer state, so the field always reflects the latest selection.
- **Failing validation after selecting?** No. Selection → `dispatch` → `state.registration` updated → `enableReinitialize` copies it into Formik `values` → Yup validates the fresh values. (Yup coerces the Luxon `DateTime` fine — `DateTime.valueOf()` returns epoch ms, so `Yup.date()`'s `new Date(value)` cast is valid.)
- **Submit?** Reads `state.registration` directly, so it's correct regardless of Formik.

The only real nuance underneath the comment is minor and cosmetic: nothing calls `setFieldTouched` on interaction, so the inline error text only appears after a submit attempt (Formik's `setNestedObjectValues(values, true)` on submit populates `touched[id].startsAt/endsAt`, which is what the `getIn(touched, …)` wiring reads). That's normal "validate-on-submit" UX and doesn't block or corrupt anything. No change needed.

apps/frontend/src/components/questionary/questionaryComponents/VisitBasis/QuestionaryComponentVisitBasis.tsx
```diff
@@ -21,8 +20,24 @@ import { useFormattedDateTime } from 'hooks/admin/useFormattedDateTime';
...
+// `startsAt`/`endsAt` are Luxon DateTimes once edited, but come back as ISO
+// strings when the registration is loaded from the backend.
+const toDateTime = (value: unknown): DateTime | undefined => {
+  if (value instanceof DateTime) {
+    return value;
+  }
+
+  if (typeof value === 'string') {
+    return DateTime.fromISO(value);
+  }
```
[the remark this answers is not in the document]
shivoomiess: empty/malformed strings are not actually sent here in practice due to backend types

apps/frontend/src/components/common/FormikUIDayRangePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+    });
+  };
+
+  const displayValue = value?.from
+    ? `${formatDate(value.from, format)} - ${formatDate(value.to, format)}`
+    : '';
+
+  const disabledDays: Matcher | undefined = minDate
+    ? { before: minDate.toJSDate() }
+    : undefined;
```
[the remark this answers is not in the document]
shivoomiess: False Postive. 

Claude:
This repo is on `react-day-picker@10.0.0`. I read its matcher implementation — `node_modules/react-day-picker/dist/cjs/utils/dateMatchModifiers.js:54-55`:

```js
if (isDateBeforeType(matcher)) {
  return differenceInCalendarDays(matcher.before, date) > 0;
}
```

`{ before: minDate.toJSDate() }` is a `DateBefore` matcher, and `differenceInCalendarDays` **compares calendar days and discards the time-of-day**. For today's cell vs. `DateTime.now()` (today 14:30): `differenceInCalendarDays(today@14:30, today@00:00) === 0`, and `0 > 0` is `false` → today is **not** disabled. Copilot's premise (that the raw time makes midnight "before" now) holds for v8's time-sensitive comparison but not for the v10 API in use here. This also correctly matches the validation schema's `TODAY_MIDNIGT` `.min()`. No fix needed; normalizing would be harmless but is dead weight.

zacharyjhankin:
According to the international standard: ISO-8601 the week starts on a monday. Its just in the USA that they do not adhere to the ISO standard and start it on a Sunday. 

https://help.tableau.com/current/pro/desktop/en-us/dates_calendar.htm

image

I would like us to use the international standard if possible.

You can easily change it in code by using ISOWeek:
```


<DayPicker ISOWeek
            mode="range"
          />
```

PR #1605, opened 2026-06-24
fix: stfc remove roles bug

ellen-wright:
## Description

refs: #https://github.com/UserOfficeProject/issue-tracker/issues/1610

Fixes bug that stopped the removal of the all UOS given roles.

Now when removing a role if it is the last role given in UOS and not from STFC UOWS then all UOS roles are removed from the role_user table for that user and only roles from UOWS are left.

## Motivation and Context
When wanting to remove all roles given in the UI and not from UOWS there was a bug where the role would not be deleted as it was the last roles associated with that user in the role_user table, however an STFC user will always have roles given by UOWS  and therefore can have all UOS given roles removed without error. 

### DIFF
--- a/apps/backend/src/datasources/mockups/UserDataSource.ts
+++ b/apps/backend/src/datasources/mockups/UserDataSource.ts
@@ -475,6 +475,9 @@ export class UserDataSourceMock implements UserDataSource {
   async setUserRoles(id: number, roles: number[]): Promise<void> {
     // Do something here or remove the function.
   }
+  async removeUserRoles(id: number): Promise<void> {
+    return;
+  }
   async getUserRoles(id: number): Promise<Role[]> {
     if (id == dummyUserOfficer.id) {
       return [
--- a/apps/backend/src/datasources/postgres/UserDataSource.ts
+++ b/apps/backend/src/datasources/postgres/UserDataSource.ts
@@ -151,6 +151,12 @@ export default class PostgresUserDataSource implements UserDataSource {
     });
   }
 
+  async removeUserRoles(id: number): Promise<void> {
+    return database.transaction(async (trx) => {
+      await trx<RoleUserRecord>('role_user').where('user_id', id).del();
+    });
+  }
+
   async me(id: number): Promise<User | null> {
     return database
       .select()
--- a/apps/backend/src/datasources/stfc/StfcUserDataSource.ts
+++ b/apps/backend/src/datasources/stfc/StfcUserDataSource.ts
@@ -416,7 +416,7 @@ export class StfcUserDataSource implements UserDataSource {
       );
 
       if (newRolesToAssign.length == 0) {
-        postgresUserDataSource.removeUserRoles(id);
+        this.removeUserRoles(id);
         this.stfcRolesCache.remove(String(id));
         this.uopRolesCache.remove(String(id));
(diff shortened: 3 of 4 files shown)


deepaksftc (approved):

apps/backend/src/datasources/mockups/UserDataSource.ts
```diff
@@ -475,6 +475,9 @@ export class UserDataSourceMock implements UserDataSource {
   async setUserRoles(id: number, roles: number[]): Promise<void> {
     // Do something here or remove the function.
   }
+  async removeUserRoles(id: number): Promise<void> {
+    return;
+  }
```
mutambaraf, line 480: We are not mocking any thing here

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -522,6 +526,10 @@ export class StfcUserDataSource implements UserDataSource {
     throw new Error('Method not implemented.');
   }
 
+  async removeUserRoles(id: number): Promise<void> {
+    throw new Error('Method not implemented.');
+  }
```
mutambaraf, line 533: We may need to pass this to the super class or implement it.

apps/backend/src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -415,7 +415,11 @@ export class StfcUserDataSource implements UserDataSource {
         (roleId) => !assignedRoleIds.has(roleId) && roleId !== userRole?.id
       );
 
-      if (newRolesToAssign.length > 0) {
+      if (newRolesToAssign.length == 0) {
+        postgresUserDataSource.removeUserRoles(id);
```
mutambaraf: We may need await here

apps/backend/src/datasources/postgres/UserDataSource.ts
```diff
@@ -151,6 +151,12 @@ export default class PostgresUserDataSource implements UserDataSource {
     });
   }
 
+  async removeUserRoles(id: number): Promise<void> {
+    return database.transaction(async (trx) => {
```
mutambaraf, line 155: We may be missing await

apps/backend/src/datasources/postgres/UserDataSource.ts
```diff
@@ -151,6 +151,12 @@ export default class PostgresUserDataSource implements UserDataSource {
     });
   }
 
+  async removeUserRoles(id: number): Promise<void> {
+    return database.transaction(async (trx) => {
+      await trx<RoleUserRecord>('role_user').where('user_id', id).del();
```
mutambaraf, line 156: This may leave some one with out any assigned role do not we need to leave at least the user role

mutambaraf (commented):
Some few comments.

YufanKambang (approved):

apps/backend/src/datasources/postgres/UserDataSource.ts
ellen-wright: This will only remove the roles assigned in the role)user table all UOWS assigned roles will still be there and we give everyone at least one role from UOWS right?

mutambaraf (approved):

PR #1606, opened 2026-06-25
fix: improve sample error messaging

### LINKED ISSUE #1378: Generic templates validation error
This came through support 16/04/2025.
"Hi,  I have just had a user contact me because they couldn’t finish their proposal submission due to the very unhelpful error “All generic templates must be completed.” on the (5) Samples page.
 
After they gave me edit rights I went in and found out that this was because they had entered too much information in the “Please give at least one of the following with appropriate units”.  The character limit for this is only 50, but they had entered more, and could still Save the data and carry on – the warning message appears too high up the screen to be seen on my particular setup, and on the (5) Samples page the subtle red colouring of the icon next to the sample name doesn’t exactly leap out (if it was an actual warning symbol, perhaps it would be more easily interpreted).
 
To give a better UX in future I would strongly suggest that either the text edit box is made fully limited to 50 characters, or even better, just provide specific entries for each of the quantities that can be selected. I also wonder why this type of information is even required at submission time? For myself in Disordered Materials it doesn’t tell me, the instrument scientist, anything useful."
Full discussion in in [teams](https://teams.microsoft.com/l/message/19:fbe05f2b9db74c42972d9d9188711a5a@thread.skype/1744792538025?tenantId=3f66361c-a87e-4158-8f61-99e82db3cac8&groupId=403a668a-985f-445b-a4e3-b3a3070bb25b&parentMessageId=1744792538025&teamName=ISIS%20Computing%20Division%20Staff&channelName=Reviews%20and%20Allocations&createdTime=1744792538025)

ChongJiaChua:
Closes [#1378](https://github.com/UserOfficeProject/issue-tracker/issues/1378)
## Description
This PR improves validation feedback for generic templates on proposal sample pages. When one or more sample entries are incomplete or contain answers that violate constraints, the parent page now lists the affected sample names instead of showing the generic "All genericTemplates must be completed" message.

## Motivation and Context
Users could save draft progress in a generic template with an answer that violated constraints, such as text longer than the configured character limit. This draft save is allowed, but when they later tried to continue from the parent Samples page, they only saw the ambiguous error "All genericTemplates must be completed".

## How Has This Been Tested

Added e2e tests.

## Changes
1. Updated the error message to list the affected sample(s).
2. Shows exact validation errors when reopening an invalid saved sample.
3. Changed incomplete generic-template list entries to use a warning icon instead of a red document icon.

### DIFF
--- a/apps/frontend/src/components/questionary/questionaryComponents/GenericTemplate/GenericTemplateDefinition.tsx
+++ b/apps/frontend/src/components/questionary/questionaryComponents/GenericTemplate/GenericTemplateDefinition.tsx
@@ -10,6 +10,25 @@ import QuestionaryComponentGenericTemplate from './QuestionaryComponentGenericTe
 import { QuestionGenericTemplateForm } from './QuestionGenericTemplateForm';
 import { QuestionTemplateRelationGenericTemplateForm } from './QuestionTemplateRelationGenericTemplateForm';
 import { QuestionaryComponentDefinition } from '../../QuestionaryComponentRegistry';
+
+type GenericTemplateValidationValue = {
+  title?: string | null;
+  questionary?: {
+    isCompleted?: boolean | null;
+  } | null;
+};
+
+const getIncompleteGenericTemplatesMessage = (
+  genericTemplates: GenericTemplateValidationValue[]
+) => {
+  const genericTemplateTitles = genericTemplates
+    .filter((genericTemplate) => !genericTemplate?.questionary?.isCompleted)
+    .map((genericTemplate) => genericTemplate.title || 'Untitled sample')
+    .join(', ');
+
+  return `The following sample(s) are violating constraints: ${genericTemplateTitles}. Fix the red sections and click "Save and continue" in each sample before continuing.`;
+};
+
 export const genericTemplateDefinition: QuestionaryComponentDefinition = {
   dataType: DataType.GENERIC_TEMPLATE,
   name: 'Sub Template',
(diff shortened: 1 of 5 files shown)


apps/frontend/src/components/questionary/questionaryComponents/GenericTemplate/GenericTemplateDefinition.tsx
```diff
@@ -10,6 +10,25 @@ import QuestionaryComponentGenericTemplate from './QuestionaryComponentGenericTe
...
+
+const getIncompleteGenericTemplatesMessage = (
+  genericTemplates: GenericTemplateValidationValue[]
+) => {
+  const genericTemplateTitles = genericTemplates
+    .filter((genericTemplate) => !genericTemplate?.questionary?.isCompleted)
+    .map((genericTemplate) => genericTemplate.title || 'Untitled sample')
+    .join(', ');
+
+  return `The following sample(s) are violating constraints: ${genericTemplateTitles}. Fix the red sections and click "Save and continue" in each sample before continuing.`;
```
ACLay: This looks like it'd work for the samples case in the issue, but generic templates can be used for other things in the questionnaires too, so I don't think we should be hardcoding the text to be about samples.

ChongJiaChua: Yea that's a good point. I've updated the message to be generic rather than specific to samples.

ACLay (approved):

PR #1613, opened 2026-06-30
fix: getProposal api error

ellen-wright:
## Description
refs: https://github.com/UserOfficeProject/issue-tracker/issues/1620
There has been an error when calling the getProposal api call that the isComplete cannot be null. 
This can be resolved by allowing it to be null. 

### DIFF
--- a/apps/backend/src/resolvers/types/Questionary.ts
+++ b/apps/backend/src/resolvers/types/Questionary.ts
@@ -38,7 +38,7 @@ export class QuestionaryResolver {
     );
   }
 
-  @FieldResolver(() => Boolean)
+  @FieldResolver(() => Boolean, { nullable: true })
   async isCompleted(
     @Root() questionary: Questionary,
     @Ctx() context: ResolverContext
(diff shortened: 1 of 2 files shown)


zacharyjhankin:
I don't understand the purpose of this PR. This API ultimately calls  QuestionaryDataSource.ts (getIsCompleted) which returns a promise(boolean) So why do we need to allow the API to return null if it can't return null given its implementation? What is the issue you are facing?

jekabs-karklins:
I also believe that the problem is somewhere else. By inspecting code I can see that if it is `undefined` or `null` in the datasource it should be set to `false` upstream

https://github.com/UserOfficeProject/user-office-core/blob/f1225fdc84702a8f103eee8d42a83faecdb62f6f/apps/backend/src/datasources/postgres/QuestionaryDataSource.ts#L444

I think we would want to investigate to see how does `isComplete` become `null` and fix it there, maybe?

ellen-wright:
Thanks for the help, I will close this PR and do some more looking into it

simonfernandes:
@zacharyjhankin @jekabs-karklins 

FYI I have opened https://github.com/UserOfficeProject/user-office-core/pull/1734 to address this.

PR #1615, opened 2026-06-30
feat: allow inviting Data Access Users to Proposal with email

shivoomiess:
## Description
This PR introduces the capability to invite data access users to a proposal by creating a new data access claim.

## Motivation and Context
This change is required to allow data access users to be linked directly to proposals, enabling them to view and manipulate data related to the proposal they are invited to. This enhances collaboration and data management in the project.

## Changes
1. A new table 'data_access_claims' is created in the database to store the data access user claims.
2. The 'DataAccessClaimDataSource' token is added to various configuration files to enable the data access claim functionality across different environments.
3. The 'DataAccessClaimDataSource' is mapped to 'PostgresDataAccessClaimDataSource' in the dependency configuration files to tie the functionality with the PostgreSQL database.
4. GraphQL queries and mutations have been added on the frontend and backend for getting / setting DAU email invites.
6. The read/write path for creating invites has been mirrored from the existing code for Co Proposer Invites and corresponding mutations have been added.
7. On the frontend Proposal Summary page, we now see invited emails inline for both Co-Proposers and Data Access Users instead of a separate row for only co-proposers.

## Visual Changes

The invite pop-ups have been updated to look clean and show the type of proposal invite (Co-Proposer / Data Access). 

### Previous
Screenshot 2026-07-06 at 15 48 27

### Current
Screenshot 2026-07-06 at 16 02 22

### Adding and Deleting Invited Users 

1. Changes are saved to the DB

https://github.com/user-attachments/assets/8a7a68a8-c8b3-4e18-8a9d-bd50c0bb2549

https://github.com/user-attachments/assets/f25b1ce0-0109-427e-8696-3cf81d75fb2b

2. Emails are sent correctly 

https://github.com/user-attachments/assets/2a429b68-baf0-4501-bebd-69c03c3fac17

### Invite Popups for New Sign-Ups

https://github.com/user-attachments/assets/ffceded6-bb7e-42f7-933c-a4e2eb224ead

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5408](https://jira.ess.eu//browse/SWAP-5408)


apps/backend/src/datasources/postgres/records.ts
```diff
@@ -1463,6 +1464,17 @@ export interface CoProposerClaimRecord {
 export const createCoProposerClaimRecord = (invite: CoProposerClaimRecord) =>
   new CoProposerClaim(invite.invite_id, invite.proposal_pk);
 
+// TODO: create a similar claim record for Data Access User
```
yoganandaness: Should we remove this?

apps/backend/src/datasources/postgres/DataAccessUsersDataSource.ts
```diff
@@ -90,6 +90,53 @@ export default class PostgresDataAccessUsersDataSource
...
+          user_id: userId,
+        };
+
+        await trx('data_access_user_has_proposal')
+          .insert(insertData)
+          .onConflict(['proposal_pk', 'user_id'])
+          .ignore();
+      });
+
+      const addedUser = await database
```
yoganandaness: Do we need this?

apps/backend/src/datasources/postgres/DataAccessUsersDataSource.ts
```diff
@@ -90,6 +90,53 @@ export default class PostgresDataAccessUsersDataSource
...
+    userId: number
+  ): Promise<BasicUserDetails | Rejection> {
+    try {
+      await database.transaction(async (trx) => {
+        const insertData = {
+          proposal_pk: proposalPk,
+          user_id: userId,
+        };
+
+        await trx('data_access_user_has_proposal')
```
yoganandaness: Recommend to test

apps/backend/src/datasources/DataAccessUsersDataSource.ts
yoganandaness, line 1: Remove the todo and deprecated comments.

yoganandaness (commented):
@shivoomiess Kindly attach relevant video or screenshot

apps/backend/src/eventHandlers/logging.ts
```diff
@@ -81,6 +81,20 @@ export default function createLoggingHandler() {
 
           break;
         }
+        case Event.PROPOSAL_DATA_ACCESS_INVITE_SENT: {
```
yoganandaness, line 94: requesting to test this.

apps/backend/src/eventHandlers/email/essEmailHandler.ts
```diff
@@ -350,6 +350,51 @@ export async function essEmailHandler(event: ApplicationEvent) {
       break;
     }
 
+    case Event.PROPOSAL_DATA_ACCESS_INVITES_UPDATED: {
```
yoganandaness, line 363: test this as well.

apps/backend/src/mutations/InviteMutations.ts
yoganandaness, line 1: Can we have a clean diff

yoganandaness (changes requested):
1. kindly test all pieces of change
2. Attach video or pic if possible
3. Pls have clean diff

apps/backend/src/queries/InviteQueries.ts
```diff
@@ -37,6 +37,24 @@ export default class InviteQueries {
     return invites;
   }
 
+  @Authorized()
+  async getDataAccessInvites(agent: UserWithRole | null, proposalPk: number) {
+    const hasReadRights =
+      this.userAuth.isApiToken(agent) ||
+      this.proposalAuth.hasReadRights(agent, proposalPk);
```
jekabs-karklins: fix: we should await the hasReadRights invocation otherwise the check always is ´true`

apps/backend/src/eventHandlers/email/essEmailHandler.ts
```diff
@@ -350,6 +350,51 @@ export async function essEmailHandler(event: ApplicationEvent) {
...
+            event,
+          });
+
+          return;
+        }
+
+        await sendInviteEmail(
+          invite,
+          inviter,
+          EmailTemplateId.USER_OFFICE_REGISTRATION_INVITATION_CO_PROPOSER,
```
jekabs-karklins: fix: I think we should create a new constant in EmailTemplateId for creating specialized email. While email text should not be part of this PR creating a constant should be

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -1,429 +1,561 @@
...
+
+    for await (const claim of dataAccessClaim) {
+      const isDataAccessUser =
+        await this.dataAccessUsersDataSource.isDataAccessUserOfProposal(
+          claimerUserId,
+          claim.proposalPk
+        );
+      // already a data access user
+      if (isDataAccessUser) {
+        return;
```
jekabs-karklins: suggestion: although not a real world use case for now, can we have a `continue` here? If at some point there are two data access claims attached to the invite and for the first claim isDataAccessUser is already true, the second will not be avaluated

jekabs-karklins (commented):
Over all looks very good :raisedhands:
As this is a bigger PR I think you will get a bit more comments :). 

I have checked the PR and left some comments, and Happy to chat if any help or clarification needed.

shivoomiess (commented):
incorporated feedback

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -72,9 +83,14 @@ export default class InviteMutations {
...
-    await this.processAcceptedCoProposerClaims(agent.id, invite);
-    await this.processAcceptedVisitRegistrationClaims(agent.id, invite);
+    try {
+      await this.processAcceptedRoleClaims(agent.id, invite);
+      await this.processAcceptedCoProposerClaims(agent.id, invite);
+      await this.processAcceptedDataAccessClaims(agent.id, invite);
+      await this.processAcceptedVisitRegistrationClaims(agent.id, invite);
+    } catch (error) {
+      logger.logException('Error during claim processing', error);
+    }
```
jekabs-karklins: breaking: the problem is that catch will capture the error, and flow will continue, and it will mark invite as claimed. We need to re-throw or simply remove the try...catch block all together. 
The any exception occurring are handled globally under apps/backend/src/middlewares/graphql.ts see `errorLoggingPlugin`

apps/backend/src/datasources/postgres/DataAccessUsersDataSource.ts
```diff
@@ -90,6 +90,31 @@ export default class PostgresDataAccessUsersDataSource
...
+          .ignore();
+      });
+    } catch (error) {
+      return new Rejection('Failed to add data access user', {
+        proposalPk,
+        userId,
+        error: error instanceof Error ? error.message : String(error),
+      });
+    }
+  }
```
jekabs-karklins, line 109: suggestion: there is no need for transaction here as we only do one database mutation
```suggestion
  async addDataAccessUser(proposalPk: number, userId: number): Promise<undefined | Rejection> {
  try {
    await database('data_access_user_has_proposal')
      .insert({ proposal_pk: proposalPk, user_id: userId })
      .onConflict(['proposal_pk', 'user_id'])
      .ignore();
  } catch (error) {
        return new Rejection('Failed to update data access users', {
            proposalPk,
            userIds,
            error: error instanceof Error ? error.message : String(error),
        });
  }
}
```

apps/backend/src/eventHandlers/email/essEmailHandler.ts
```diff
@@ -350,6 +350,51 @@ export async function essEmailHandler(event: ApplicationEvent) {
...
+          invite.createdByUserId
+        );
+
+        if (!inviter) {
+          logger.logError('No inviter found when trying to send email', {
+            inviter,
+            event,
+          });
+
+          return;
```
jekabs-karklins: breaking: should be `continue` instead of the `return`, so that in case the user is missing for one invite we do not skip the rest of the invites. I can see this bug originates by taking the code example from above. Could  you also address the issue in the example code?

apps/backend/src/datasources/postgres/InviteDataSource.ts
```diff
@@ -152,6 +176,47 @@ export default class PostgresInviteDataSource implements InviteDataSource {
...
+          if (filter.isClaimed) {
+            query.whereNotNull('claimed_at');
+          } else {
+            query.whereNull('claimed_at');
+          }
+        }
+
+        if (filter.isExpired) {
+          query.where('expires_at', '<', new Date());
+        }
```
jekabs-karklins: breaking: this currently checks if isExpired is a truthy value then apply the filter.
In VisitMutations we actually use this param to filter out expired invites by specifying { isExpired: false }, which looks right from outside but as this is implemented will skip the check alltogether, meaning you can accept expired invites.
Again I see this originating from the example code, the fix would look something like this


```suggestion
        if (filter.isExpired !== undefined) {
            if (filter.isExpired) {
                query.where('expires_at', '<', new Date());
            } else {
                query.where((qb) =>
                qb.whereNull('expires_at').orWhere('expires_at', '>=', new Date())
                );
            }
        }

```

jekabs-karklins (changes requested):

apps/backend/db_patches/0213_AddDataAccessClaims.sql
```diff
@@ -0,0 +1,14 @@
+DO
+$$
+BEGIN
+  IF register_patch('0213_AddDataAccessClaims', 'Shivam K', 'Adding data access user claims', '2026-06-26') THEN
```
jekabs-karklins: nitpick: can this be 214?

apps/frontend/src/components/proposal/DataAccessUsersModal.tsx
```diff
@@ -25,6 +25,7 @@ const DataAccessUsersModal = ({
     useDataAccessUsersData(proposalPk);
   const { proposalData } = useProposalData(proposalPk);
```
jekabs-karklins: suggestion: Can we add loadingProposal and add it besides the `loadingDataAccessUsers` which prevents view from rendering until proposal is loaded. 
```suggestion
  const { proposalData, loading:loadingProposal } = useProposalData(proposalPk);
```

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -202,6 +259,77 @@ export default class InviteMutations {
...
+      this.userAuth.isApiToken(agent) ||
+      (await this.dataAccessUsersAuth.hasWriteRights(agent, proposalPk));
+
+    if (!hasWriteRights) {
+      return rejection(
+        'User is not authorized to create data access invites for this proposal'
+      );
+    }
+
+    const existingInvites = await this.getDataAccessInvites(proposalPk);
```
jekabs-karklins: breaking: getDataAccessInvites will get also claimed and expired invites. Since frontend will send only active invites the claimed and expired ones will be deleted. We should only get the active invites from getDataAccessInvites method

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -334,6 +438,49 @@ export default class InviteMutations {
...
+      );
+
+      // Rejection extends GraphQLError, so throwing aborts the accept before
+      // the invite is stamped as claimed.
+      if (failure) {
+        throw failure;
+      }
+
+      this.eventBus.publish({
+        type: Event.PROPOSAL_DATA_ACCESS_INVITE_ACCEPTED,
```
jekabs-karklins, line 491: issue: we need to map this event in the messageBroker to make sure the event PROPOSAL_UPDATED is sent out to the RabbitMQ so that scicat gets the update. see DATA_ACCESS_USERS_UPDATED in the messageBroker

apps/backend/src/eventHandlers/email/essEmailHandler.ts
```diff
@@ -350,12 +350,8 @@ export async function essEmailHandler(event: ApplicationEvent) {
       break;
     }
```
jekabs-karklins, line 362: issue(maybe): did you mean to delete the Event.DATA_ACCESS_USERS_UPDATED ? This is still being dispatched.

apps/backend/src/mutations/InviteMutations.ts
```diff
@@ -334,6 +438,49 @@ export default class InviteMutations {
...
+    const dataAccessClaim =
+      await this.dataAccessClaimDataSource.findByInviteId(inviteId);
+
+    for await (const claim of dataAccessClaim) {
+      const isDataAccessUser =
+        await this.dataAccessUsersDataSource.isDataAccessUserOfProposal(
+          claimerUserId,
+          claim.proposalPk
+        );
+      // already a data access user
```
jekabs-karklins, line 457: suggestion: shall we also check if user is not already a member of proposal the same as https://github.com/UserOfficeProject/user-office-core/blob/dafb880b027bcf181bd562f761e23bc292a29aa3/apps/backend/src/mutations/DataAccessUsersMutations.ts#L56 ?

apps/backend/src/eventHandlers/email/essEmailHandler.ts
```diff
@@ -366,43 +362,37 @@ export async function essEmailHandler(event: ApplicationEvent) {
...
-            });
+          continue;
+        }
+
+        await sendInviteEmail(
+          invite,
+          inviter,
+          EmailTemplateId.USER_OFFICE_REGISTRATION_INVITATION_DATA_ACCESS_USER,
+          { proposalTitle: proposal.title, proposalId: proposal.proposalId }
+        ).then(async () => {
```
jekabs-karklins: suggestion: would you mind to do a bit more general fix for re-throwing on :581, because if sendInviteEmail fails, it catches and resolves, which will cause the then clause to execute and publish PROPOSAL_DATA_ACCESS_INVITE_SENT which is wrong due to an error. So we get incorrect audit trail

jekabs-karklins:
I thiks otherwise this looks very good, just let me know your thoughts on the comments I mentioned

shivoomiess:
@zacharyjhankin @simonfernandes we need one more review before we merge this, could you please take a look? my summer engagement at ESS is concluding next week

jekabs-karklins (approved):
Looks good now, looking forward to have this feature on production 👍🏻

apps/backend/src/datasources/mockups/DataAccessUsersDataSource.ts
```diff
@@ -46,6 +46,7 @@ export const dummyInstitution2 = new Institution(11, 'Research Center', 2);
 
 export const dummyCountry = new Country(1, 'Denmark');
 export const dummyCountry2 = new Country(2, 'United Kingdom');
+export const dummyCountry3 = new Country(3, 'Belarus');
```
mutambaraf, line 49: We can remove dummyCountry3  because it is not being used.

apps/backend/src/datasources/mockups/InviteDataSource.ts
```diff
@@ -2,37 +2,36 @@ import { inject, injectable } from 'tsyringe';
...
 
 @injectable()
 export class InviteDataSourceMock implements InviteDataSource {
   private invites: Invite[];
-  private coProposerClaims: CoProposerClaim[];
 
   constructor(
     @inject(Tokens.CoProposerClaimDataSource)
-    private coProposerDataSource: CoProposerClaimDataSource
+    private coProposerDataSource: CoProposerClaimDataSourceMock,
```
mutambaraf, line 20: We can improve naming to  coProposerClaimDataSource

apps/backend/src/datasources/mockups/InviteDataSource.ts
```diff
@@ -2,37 +2,36 @@ import { inject, injectable } from 'tsyringe';
...
 
-  async findCoProposerInvites(proposalPk: number): Promise<Invite[]> {
-    const coProposerClaims =
-      await this.coProposerDataSource.findByProposalPk(proposalPk);
-
-    const invites = await Promise.all(
-      coProposerClaims.map((claim) => this.findById(claim.inviteId))
-    );
+  async findPendingCoProposerInvites(proposalPk: number): Promise<Invite[]> {
+    return this.getCoProposerInvites({ proposalPk, isClaimed: false });
```
mutambaraf, line 28: Maybe lets avoid hard coding this variable by actually passing in through findPendingCoProposerInvites

apps/backend/src/datasources/postgres/InviteDataSource.ts
```diff
@@ -2,36 +2,19 @@
...
-        if (isClaimed !== undefined) {
-          if (isClaimed) {
-            query.whereNotNull('claimed_at');
-          } else {
-            query.whereNull('claimed_at');
-          }
-        }
-      })
+  findPendingCoProposerInvites(proposalPk: number): Promise<Invite[]> {
+    return this.getCoProposerInvites({ proposalPk, isClaimed: false });
```
mutambaraf, line 13: We can also avoid hard coding isClaimed: false

apps/backend/src/datasources/postgres/records.ts
```diff
@@ -1464,6 +1464,11 @@ export interface CoProposerClaimRecord {
 export const createCoProposerClaimRecord = (invite: CoProposerClaimRecord) =>
   new CoProposerClaim(invite.invite_id, invite.proposal_pk);
 
+export interface DataAccessClaimRecord {
+  readonly invite_id: number;
+  readonly proposal_pk: number;
+}
+
```
mutambaraf, line 1471: We can remove this if we are not using it.

apps/backend/src/datasources/postgres/DataAccessClaimDataSource.ts
```diff
@@ -0,0 +1,26 @@
...
+      .then((rows) =>
+        rows.map((row) => new DataAccessClaim(row.invite_id, row.proposal_pk))
+      );
+  }
+  async create(inviteId: number, proposalPk: number): Promise<DataAccessClaim> {
+    return database('data_access_claims')
+      .insert({ invite_id: inviteId, proposal_pk: proposalPk })
+      .returning('*')
+      .then((rows) => {
+        const row = rows[0];
```
mutambaraf, line 21: We may need to check if rows is an array and I think this is where we where supposed to use DataAccessClaimRecord

apps/backend/src/datasources/mockups/InviteDataSource.ts
```diff
@@ -185,56 +199,112 @@ export class InviteDataSourceMock implements InviteDataSource {
...
-          if (invite.createdAt >= filter.createdBefore) {
-            return false;
-          }
+
+  async getCoProposerInvites(
+    filter: GetProposalInvitesFilter
+  ): Promise<Invite[]> {
+    const inviteIdsOnProposal = filter.proposalPk
+      ? (
+          await this.coProposerDataSource.findByProposalPk(filter.proposalPk)
```
mutambaraf, line 208: It may be more clear if we assign the value of  await this.coProposerDataSource.findByProposalPk(filter.proposalPk) to a variable an check if it is returned.

mutambaraf (commented):
Looks good and left few comments

yoganandaness:
@mutambaraf Thanks for your feedback. Your comments have been resolved and requesting you to re-review and approve. Thanks.

apps/backend/db_patches/0214_AddDataAccessClaims.sql
```diff
@@ -0,0 +1,14 @@
+DO
+$$
+BEGIN
+  IF register_patch('0214_AddDataAccessClaims', 'Shivam K', 'Adding data access user claims', '2026-06-26') THEN
```
mutambaraf: Before we merge we can update the date here

mutambaraf (approved):
LGTM

yoganandaness (approved):

PR #1618, opened 2026-07-02
fix: add dependency to dynamic http question type

zacharyjhankin:
This adds the option for dependencies for the dynamic http question as we need it here at diamond. Tested locally.

image

image

image


GrantDLS (approved):
LGTM

Scott-James-Hurley (approved):

PR #1619, opened 2026-07-02
feat: diamond production/staging settings no longer gets overwritten on deployments

zacharyjhankin:
This PR lets diamond settings match the same logic as ESS where on prod/staging deployments the configuration does not get re-applied. This means that when someone in the user office role changes a setting it will not be overwritten the next time that we deploy.

However, the colours in staging and development are going to be set on deployment so it is clear what environment you are in. The settings in development will be auto set up for you. Staging and prod will be left the same so they can be configured as you like. 

Tested locally.

Development colour:
image

staging colour:

image

I went with the dark purple in staging and the red in development because purple is closer to blue which is production so it reflects that you are closer to production.


GrantDLS (approved):

zacharyjhankin:
https://jira.diamond.ac.uk/browse/UAS-14546

PR #1626, opened 2026-07-06
feat: send email notifications to DAU newly added to a proposal

shivoomiess:
## Description
This PR adds the feature to send email notifications to Data Access User (DAU) when they are newly added to a proposal.

## Motivation and Context
This change is required to ensure that newly added DAUs are promptly notified about their addition to a proposal, thereby improving communication and efficiency.

## Changes
1. An enum for 'DATA_ACCESS_USER_ADDED' has been added to the email template IDs.
2. A case for 'DATA_ACCESS_USERS_UPDATED' event has been added to the email handler. It sends emails to only the newly added users and skips if no new users are added.
3. An 'invitedUserIds' field has been added to the 'UserDataAccessUpdatedEvent' interface to keep track of newly invited users.
4. 'DataAccessUsersMutations' class has been updated to capture the current members before updating the data access users, to work out who is genuinely newly invited.

## How Has This Been Tested
This has been tested by verifying that email notifications are sent out accurately to newly added DAUs, and no notifications are sent out in case of no new additions. Error handling has been implemented to log an error message if an email could not be sent.

The logs confirm this is the case:
Proof of Email sent in logs

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5655](https://jira.ess.eu//browse/SWAP-5655)


apps/backend/src/events/applicationEvents.ts
```diff
@@ -415,6 +415,9 @@ interface VisitRegistrationCancelledEvent extends GeneralEvent {
 interface UserDataAccessUpdatedEvent extends GeneralEvent {
   type: Event.DATA_ACCESS_USERS_UPDATED;
   proposalPKey: number;
+  // Users newly added as data access users in this update (empty on removals
+  // or no-op updates). Used to notify only the freshly invited users by email.
+  invitedUserIds: number[];
```
jekabs-karklins: suggestion: can we name this `newlyAddedUserIds` . `invited...` is a bit misleading as this name relates to email invites

jekabs-karklins (approved):
LGTM - one minor change request.

Non-blocking thought: wiring this only into essEmailHandler means any other facility wanting the same notification has to reimplement a near-identical handler. Might be worth rethinking the email-handling structure so common cases live in one place while facilities can still add their own. Follow-up, not for this PR.

simonfernandes (approved):

PR #1634, opened 2026-07-08
refactor: replace old PeopleSelectorModal with ProposalPeopleSelector in FAP

shivoomiess:
## Description
This PR refactors and replaces the old PeopleSelectorModal with ProposalPeopleSelector in the Facility Access Panel (FAP).

## Motivation and Context
The change was necessary to streamline the user selection process in the FAP and provide a more cohesive user interface. This improves user experience by making the process more intuitive and efficient.

## Changes
1. Replaced instances of 'PeopleSelectorModal' with 'ProposalPeopleSelectorModal' in the FAP.
2. Simplified the 'addUserThroughModal' function, removing redundant lines of code.
3. Updated Cypress tests to reflect these changes, ensuring the functionality remains intact.

## How Has This Been Tested?

Screenshot 2026-07-08 at 11 52 48

Screenshot 2026-07-08 at 11 52 53

Screenshot 2026-07-08 at 11 53 00

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5652](https://jira.ess.eu//browse/SWAP-5652)


yoganandaness (changes requested):
@shivoomiess I have few thoughts and req. you to validate it. 

1. I can see that now the existing PeopleSelector is only used in the place of Instruments table. Can we also convert that into the new Selector for adding instruments as well(Without invite as of now). If you agree, this could be done in the same PR with e2e. 
2. The component name reads `ProposalPeopleSelector`. Is there any specific reason behind naming it after Proposal as it is Proposal specific or can it be removed and just be renamed People Selector after removing the existing People Selector.

shivoomiess:
> @shivoomiess I have few thoughts and req. you to validate it.
> 
> 1. I can see that now the existing PeopleSelector is only used in the place of Instruments table. Can we also convert that into the new Selector for adding instruments as well(Without invite as of now). If you agree, this could be done in the same PR with e2e.
> 2. The component name reads `ProposalPeopleSelector`. Is there any specific reason behind naming it after Proposal as it is Proposal specific or can it be removed and just be renamed People Selector after removing the existing People Selector.

1. It is possible and I will look at this further. 
2. `ProposalPeopleSelector` is a legacy name. I believe when the component was designed, it was only with Proposals in mind and placed in the folder with components relating to proposals. Since it can fairly easily be re-used in other tables for People Selector, we can factor it out into a common folder (like common or utils) and do some cleanup.

apps/frontend/src/components/fap/Members/FapMembers.tsx
```diff
@@ -66,16 +70,12 @@ const FapMembers = ({
   onFapUpdate,
   confirm,
 }: FapMembersProps) => {
-  const [modalOpen, setOpen] = useState(false);
-  const [fapChairModalOpen, setFapChairModalOpen] = useState(false);
-  const [fapSecretaryModalOpen, setFapSecretaryModalOpen] = useState(false);
+  const [openModal, setOpenModal] = useState<OpenModal | null>(null);
```
jekabs-karklins: nitpick: because this is a state variable the name modalOpen reads better that openModal

shivoomiess: Good catch, renamed to `modalOpen` (and the setter to `setModalOpen`) in 75c8f51.

shivoomiess:
@yoganandaness I've done them in #1683, which is on top of this branch so this PR stays focused:

1. **Instruments converted.** Done, without invites as you suggested. While doing it I found the Techniques table used the old selector for the same "assign scientist" flow, so I converted that too — otherwise the old component couldn't actually be deleted. Cypress specs for both are updated.
2. **Renamed.** The old selector has been removed and `ProposalPeopleSelectorModal` is now just `PeopleSelectorModal`. It has also been moved into the common components folder.

Two things there worth your attention: 
- converting these screens costs some discoverability (the old table listed every instrument scientist; the new one needs 3 characters before it searches), and 
- `PeopleTable` is left with a few now-unreachable props that I deliberately left for a follow-up. Both are called out in the PR description.

One note for this PR: the FAP member step in `eventLogs.cy.ts` has been fixed.

jekabs-karklins (approved):

shivoomiess:
fixed one flaky E2E test

apps/e2e/cypress/support/peopleSelector.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Picks a user in the ProposalPeopleSelectorModal autocomplete and submits the
+ * dialog. `search` needs at least 3 characters, otherwise the selector does not
+ * query.
+ */
+export function addUserThroughModal(search: string) {
+  cy.get('[data-cy="invite-user-autocomplete"]').type(search);
+  // Matching on the option itself re-queries the DOM on every retry. Chaining
+  // off `[role="presentation"]` does not: that selector also matches the
+  // "No options" node rendered between the debounced search firing and the
+  // results arriving, and the retries then stay inside that stale node.
```
jekabs-karklins: suggestion: I think the code is quiet readable so no need for the extensive comment. sometimes claude adds comment where something does not work the first time. Maybe good for claude but makes it harder to read for humans

jekabs-karklins (approved):

shivoomiess:
@TCMeldrum we need one more review before we merge this, could you please take a look? my summer engagement at ESS is concluding next week

TCMeldrum (approved):

PR #1637, opened 2026-07-08
feat: allow Visit Team Lead to Edit Visitor List

shivoomiess:
## Description
This PR allows the Visit Team Lead to edit the Visitor List. There are significant changes to permissions. We disallow the Co-Proposer from accessing the visitor list. Proposers can create / read / write the list. Visitors can now view other visitors. Team Leads can view / edit the list and a popup checks for confirmation when changing the team lead.

## Motivation and Context
Previously, only the Proposal Investigator (PI) or Co-Proposer had the ability to edit the Visitor List. This change was required to enhance collaboration and efficiency by enabling the Visit Team Lead to also make necessary changes to the Visitor List. A policy change has been made, which has been updated in the documentation.

## Changes
- Updated the authorization check in `VisitAuthorization.ts` to also allow Visit Team Leads to make changes to the Visitor List.
- Adjusted the conditional check in `useActionButtons.tsx` to include Visit Team Leads in the list of roles that are allowed to edit the Visitor List.

## How Has This Been Tested?

https://github.com/user-attachments/assets/d41cb283-1434-46df-9c9c-1adc276195a5

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5660](https://jira.ess.eu//browse/SWAP-5660)


yoganandaness (changes requested):
@shivoomiess Requesting an e2e and unit(if possible).

zacharyjhankin (approved):
Diamond is happy with this change

shivoomiess:
> @shivoomiess Requesting an e2e and unit(if possible).

added E2E tests for both Team Lead (who should be able to view) and Regular Visitor (who should NOT be able to view the button)

apps/frontend/src/hooks/proposalBooking/useActionButtons.tsx
```diff
@@ -79,7 +79,7 @@ export function useActionButtons(args: UseActionButtonsArgs) {
     let buttonState: ActionButtonState;
     let stateReason: string | null = null;
 
-    if (isPiOrCoProposer(user, event)) {
+    if (isPiOrCoProposer(user, event) || isTeamlead(user, event)) {
```
jekabs-karklins: suggestion: If you are TL and assign someone else to be TL you will loose the ability to change back, I wonder if we should warn the user that the aciton is irreversable by him?

shivoomiess: that's true. shall we send a notification popup or ...?

jekabs-karklins: I think you can use `withConfirm` HOC in the apps/frontend/src/components/proposalBooking/CreateUpdateVisit.tsx and then call confirm if the user is teamlead and not PI or Coproposer.

apps/backend/db_patches/0000_init.sql
```diff
@@ -550,7 +550,57 @@ INSERT INTO users (
                       );
 
     INSERT INTO role_user (role_id, user_id) VALUES (1, 6);
-    
+
+    INSERT INTO users (
```
yoganandaness: Pls put into seeds

shivoomiess:
@zacharyjhankin @simonfernandes we need your eyes on this as well since my summer engagement at ESS is concluding next week

apps/backend/src/queries/VisitQueries.ts
```diff
@@ -91,8 +100,45 @@ export default class VisitQueries {
...
+    let readable = false;
+    let writeable = false;
+    let createable = false;
+
+    const visit = await this.dataSource.getVisitByExperimentPk(experimentPk);
+    const experiment =
+      await this.experimentDataSource.getExperiment(experimentPk);
+
+    if (experiment) {
+      const isPi = await this.visitAuth.hasCreateRights(
```
jekabs-karklins: issue: we should use proposalAuth.isPrincipalInvestigator()

apps/backend/src/auth/VisitAuthorization.ts
```diff
@@ -117,4 +109,28 @@ export class VisitAuthorization {
 
     return true;
   }
+
+  // NOTE: Keep constraints in sync with VisitDataSource.ts
+  async hasCreateRights(
```
jekabs-karklins, line 140: issue: we should use this method in the visit.create mutation. Currently visit.create piggybacks on proposalAuth.hasReadRights

    const hasReadRights =
      this.userAuth.isApiToken(agent) ||
      (await this.proposalAuth.hasReadRights(agent, proposal));
   

   We should replace it with visitAuth.hasCreateRights

apps/backend/src/auth/VisitAuthorization.ts
```diff
@@ -117,4 +109,28 @@ export class VisitAuthorization {
 
     return true;
   }
+
+  // NOTE: Keep constraints in sync with VisitDataSource.ts
```
jekabs-karklins: nitpick: I think this comment does not bring much value

apps/backend/src/datasources/postgres/VisitDataSource.ts
```diff
@@ -1,230 +1,234 @@
...
+      .where('visits_has_users.visit_id', visitId)
+      .andWhere('visits_has_users.user_id', visitorId)
+      .then((results) => results.length > 0);
+  }
+
+  // current user -> get related visits -> get related users for each of those visits
+  async getRelatedUsersOnVisits(id: number): Promise<number[]> {
+    // Visits the user is related to: as creator, team lead, visitor,
+    // or as a member (PI or co-proposer) of the visit's proposal
+    const relatedVisitIds = database
```
jekabs-karklins, line 219: issue:comment says     (PI or co-proposer) but the co-proposers that are stored in proposal_user is never queried. I think this query should not include co-prposers nor PI. I can that were it is being used it already has pi, and co-pi on the list. so we can keep this method solely focusing on `creator, team lead, visitor`, i guess

apps/backend/src/queries/VisitQueries.ts
shivoomiess: created a separate helper to separate out the Visits-related checks from Proposal related checks at the Query level. If you think that's not a good idea, I can implement your suggestion above.

apps/backend/src/datasources/postgres/VisitDataSource.ts
shivoomiess: if the query deliberately does not include PI, that causes problems with the service account, which can be a creator but later the PI faces trouble in reading the user details and the null is propagated upto the frontend

apps/backend/src/queries/VisitQueries.ts
jekabs-karklins: ellaboration: what looks wrong here is the variable name with the method name which destroys separation of concerns
isPi = hasCreateRights

You are right that in the VisitQuery we should use visitAuth 💯 
But then the problem is variable name, Canyou simply write this out as?

 ```
createable = await this.visitAuth.hasCreateRights(
        agent,
        experiment!.proposalPk
      );
```

and remove
```
      if (isPi) {
        createable = true;
      }
```

File: apps/backend/src/datasources/postgres/VisitDataSource.ts (hunk 6 above)
Comment (jekabs-karklins) [reply apps/backend/src/datasources/postgres/VisitDataSource.ts:219]: ellaboration: TBH I'm not sure if we still use the service account for anything because we have API tokens, so I think that is a legacy thing. If you say we need PI there is not a big harm having it. One thing I would like to flag is the function signature; 
`async getRelatedUsersOnVisits(id: number): Promise<number[]>`
Can we update it to 
`async getRelatedUsersOnVisits(userId: number): Promise<number[]>`

File: apps/backend/src/auth/VisitAuthorization.ts [hunk 7]
```diff
@@ -100,13 +91,14 @@ export class VisitAuthorization {
...
+    const isPI = await this.proposalAuth.isPrincipalInvestigatorOfProposalPk(
       agent,
-      proposal
+      visit.proposalPk
     );
 
-    if (isMemberOfProposal === false) {
+    const isTeamLead = agent.id === visit.teamLeadUserId;
+
+    if (!isPI && !isTeamLead) {
```
simonfernandes, line 99: question: Is it correct that team leads can also delete the visit? Since `hasWriteRights` is also used for by the delete mutation. Didn't see a mention of it in the description.

simonfernandes (commented):
Sorry about the delay in reviewing, it looks great generally

apps/backend/src/auth/VisitAuthorization.ts
shivoomiess: removing the comment, maybe we should have some pointers which suggest that the underlying datasource implementation should match the desired constraints in the Authorizer. For example, the user data source is used to retrieve user information about visit members depending on auth. A drift here had created some bugs earlier.

apps/backend/src/auth/VisitAuthorization.ts
shivoomiess: you're correct. changing it to `hasCreateRights` in the delete mutation.

shivoomiess: `hasCreateRights` is not a good drop-in solution, since user officers don't have create rights and the authorizer uses proposal identifiers to check. So I am adding a `hasDeleteRights`, with checks specific to the `visitId`.

apps/backend/src/auth/VisitAuthorization.ts
shivoomiess: done

apps/backend/src/datasources/postgres/VisitDataSource.ts
shivoomiess: done, also fixed suggestions for async functions while at it

apps/backend/src/queries/VisitQueries.ts
shivoomiess: thanks for point out. done

shivoomiess:
@jekabs-karklins @yoganandaness @simonfernandes we should be good to merge now

apps/backend/src/mutations/VisitMutations.spec.ts
simonfernandes, line 1: suggestion (test): Adding a test for "User Officer can delete a visit" would also be nice here.

simonfernandes (approved):
I have just one comment if you have time but I don't want to hold up the PR over it. Good luck with what comes next after your summer engagement. 🙂

jekabs-karklins (approved):

yoganandaness (approved):

PR #1643, opened 2026-07-14
feat: add date range picker question type

zacharyjhankin:
## Description

This PR adds a date time range picker question. Warning that date time ranges will never be compatible with date ranges. 

image

## Motivation and Context

We need to add a way of entering good/bad dates for scheduling purposes so needed a new question type. 

## How Has This Been Tested

Locally, and with cypress. 

## Changes

Adds new question type


validation/src/Questionary/date.ts
```diff
@@ -1,7 +1,7 @@
 import { DateTime } from 'luxon';
 import * as Yup from 'yup';
 
-function normalizeDate(date: string, includeTime: boolean) {
+export function normalizeDate(date: string, includeTime: boolean) {
```
joeshannon: I can't see this function being called externally in this PR?

apps/backend/src/models/questionTypes/DateRange.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    return dateRangeQuestionValidationSchema(field).isValid(value);
+  },
+  createBlankConfig: (): DateRangeConfig => {
+    const config = new DateRangeConfig();
+    config.small_label = '';
+    config.required = false;
+    config.tooltip = '';
+    config.readPermissions = [];
+    config.includeTime = false;
```
joeshannon: I think `includeTime` is not implemented yet?

apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx
```diff
@@ -0,0 +1,50 @@
...
+        from: Yup.date().required(),
+        to: Yup.date().required(),
+      })
+    )
+    .required();
+  if (config.required) {
+    dateRangeSchema = dateRangeSchema.min(1, 'A daterange is required.');
+  }
+
+  return Yup.object({ dateRanges: dateRangeSchema });
```
joeshannon: The structure looks like e.g.
```json
{
  "value": {
    "dateRanges": [
      {
        "to": "2026-07-16T23:00:00.000Z",
        "from": "2026-07-15T23:00:00.000Z"
      }
    ]
  }
}
```
implying eventually there could be a list of date ranges in each question?
Is that correct or should we only have one range per question?


An unrelated note - do you know why the "to" key is first? Obviously this doesn't cause any functional change but I was curious as to where that is coming from.

joeshannon (commented):
It looks pretty good to me in general!

I do have a question about if we're storing the ranges correctly though, for example the following range:

Image

is stored as `{"to": "2026-08-11T23:00:00.000Z", "from": "2026-08-10T23:00:00.000Z"}`
which is actually only a 24 hour interval but I *think* you would expect this to be 48 hours (so the "to" would be 2026-08-12T23:00:00.000Z).

In a similar way you can't make/persist a single day selection.

Supporting both date only and date+time might be tricky, or perhaps just requires more thinking on my comprehension!

GrantDLS (approved):

apps/backend/src/models/questionTypes/DateRange.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable quotes */
+import { GraphQLError } from 'graphql';
+import * as Yup from 'yup';
+
+import { DateRangeConfig } from '../../resolvers/types/FieldConfig';
+import { DataType, QuestionTemplateRelation } from '../Template';
+import { Question } from './QuestionRegistry';
+
+//TODO add validation script to duo-validation
```
GrantDLS: Is this still todo or can this comment be removed?

apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx
GrantDLS: Regarding Joe's comment about 11th to 12th August being stored as:

```{"to": "2026-08-11T23:00:00.000Z", "from": "2026-08-10T23:00:00.000Z"}```

This definitely looks like a daylight savings time issue - we're currently in BST, which is UTC+1, so what I think is happening is whatever date is reported by the frontend UI is assumed to be in the local timezone, and then to store it it must be converting that to UTC, which is why we end up with 1 hour _before_ midnight of whatever day is chosen in the picker.

So 2026-08-12T00:00:00.000Z  *BST* = 2026-08-11T23:00:00.000Z *UTC*

I'm happy to help debug this if you need.

Comment (GrantDLS) [reply apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx]: Regarding Joe's comment about selecting 2 consecutive dates and it only storing a 24 hour time range - I don't think there's a right or wrong way to do this and so it just needs to be made clear in the UI whether the final date selected is inclusive or exclusive. i.e. does the range you're selecting end at the start of the "to" date, or at the end of that day?

Unfortunately, this further complicates how things are stored internally. For example, if we're _including_ the "to" date, do we store 23:59:59 of that day, or midnight of the next day? If we're _excluding_ the "to" date, do we store 23:59:59 of the _previous_ day? As I've said, there are no right or wrong answers with any of this and the complication is coming from the fact there's no pure Date object in JS that doesn't also store a time value.

My suggestion would be to make it inclusive. That seems the most intuitive. And then when `{"to": "2026-08-12T00:00:00.000Z", "from": "2026-08-11T00:00:00.000Z"}` is stored in the DB, it might only be a 24 hour period, but since it's a date picker, and not a datetime picker, we can see from the data it's the 11th to the 12th.

Comment (zacharyjhankin) [reply apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx]: To respond to the fact that I have stored the dateRanges as a list, the answer is yes. I was expecting in the future that one of the config options will allow us to get multiple inputs from the user. Specifically because when we use this question for good/bad dates, then the user might want to enter any number of good dates. We couldn't support that if each dateRange only took one entry

File: apps/backend/src/models/questionTypes/DateRange.ts (hunk 2 above)
Comment (zacharyjhankin) [reply apps/backend/src/models/questionTypes/DateRange.ts]: It is not implemented yet. But the dateTimePicker is storing times with the dates. And we will add support with time in the future. So it feels like its worth having it set to false here at least for now. Though I do understand that usually you want to not add something you're not actively using.

File: validation/src/Questionary/date.ts (hunk 1 above)
Comment (zacharyjhankin) [reply validation/src/Questionary/date.ts]: removed export

File: apps/backend/src/models/questionTypes/DateRange.ts (hunk 4 above)
Comment (zacharyjhankin) [reply apps/backend/src/models/questionTypes/DateRange.ts]: It was still a todo, but I did it. I learned that the validation repo has been moved into this repo, so I could do it now.

File: apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx (hunk 3 above)
Comment (zacharyjhankin) [reply apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx]: I have changed the UI to make the second date white, so it is clear that the backend is storing just one 24 hour block. (starting with the 10th). In this image you can see my dateRange option is starting on 23rd hour of the 9th. zulu time. Which is the the 0th hour of 10th. Going on for 24 hours. So I think that is all correct. With timezones and UI. 

image

@joeshannon @GrantDLS

Comment (GrantDLS) [reply apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx]: But 23:00 on the 9th is only 00:00 on the 10th for us in the UK at the moment. When the clocks go back in autumn and we're back on UTC, what will get stored then? I feel like dates should always be stored in UTC and converted to to local timezone when needed.

Comment (zacharyjhankin) [reply apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx]: When the UK is back on the same time as UTC then the database will store that as 00:00 on the 10th.   But that is really what needs to happen, I guess it might appear like the dates have shifted by an hour when you read through the list in the database. But that is just because you need to convert the dates to BST, and then they will all align. 

If we stored all the dates in UTC and all of the dates were starting on the same hour across the year, then when you convert them to BST then they would go out of alignment. 

We are in fact always storing the dates as UTC and converting them to BST when we need them. 

The thing is that the timezone of our project is always going to stay in BST, so the frontend will always convert to the backend UTC to BST. So don't worry about the database UTC times changing through the year.

Comment (zacharyjhankin) [reply apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx]: About the from and to order, I looked into it a bit, and its not my code. If I overwrite the order when I save it from the UI it changes the order before putting it in the database. So I think its something to do with:

apps/frontend/src/components/questionary/QuestionaryStepView.tsx

But yeah maybe not worth worrying too much about. Though I do agree its not optimal

@joeshannon

Comment (joeshannon) [reply apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx]: > I have changed the UI to make the second date white, so it is clear that the backend is storing just one 24 hour block.

I think this is better but to me as a user it's not immediately obvious that it would be an exclusive end just based on the colour. 

I'm still not completely convinced that the above is the correct UI for a single day selection. Having said that it doesn't preclude iterating on the design later. Let's have a look in person tomorrow.

File: apps/frontend/src/components/questionary/questionaryComponents/DateTimeRangePicker/DateTimeRangePickerDefinition.tsx [hunk 5]
```diff
@@ -0,0 +1,30 @@
...
+  questionForm: () => QuestionDateTimeForm,
+  questionTemplateRelationForm: () => QuestionTemplateRelationDateForm,
+  readonly: false,
+  creatable: true,
+  icon: <TodayIcon />,
+  renderers: {
+    questionRenderer: defaultRenderer.questionRenderer,
+    answerRenderer: DateTimeRangeAnswerRenderer,
+  },
+  createYupValidationSchema: null,
```
zacharyjhankin, line 27: My understanding of this is that this validation schema is sent as a formik prop to the actual place where QuestionaryComponentDateTimeRangePicker.tsx or more like (QuestionaryComponentDateRangePicker) is actually rendered. Which is usually fine and how it is supposed to work, so when you press "continue" on the UI the red text complaining that something is wrong. (such as it being required)  Can be shown without any reference to the backend. But in this specific case the UI is going through QuestionaryComponentDateTimeRangePicker which calls FormikUIDayTimeRangePicker.tsx which calls a normal DayPicker.  Which I (((think))) might be using the same validation schemas, so you can't select anything because when you enter the start date, then the second one is empty so it gets annoyed and clears your input. But I don't really know what is going on. But at any rate, I can't get this to work. So if I just set it to null, then there is no frontend validation, but there is backend validation that pops up when  you press the submit option. I think this is adequate as it is only used when the question is required and it is not entered. And I think we can fix this in future development if we need.

apps/frontend/src/components/questionary/questionaryComponents/DateRangePicker/DateRangePickerDefinition.tsx
zacharyjhankin: Resolving conversation because I re-wrote it to work on dateTimes and not dates.

apps/frontend/src/components/questionary/QuestionaryStepView.tsx
```diff
@@ -55,6 +55,11 @@ export const createFormikConfigObjects = (
         answer,
         state,
       });
+    } else {
```
yoganandaness: is this to be removed?

apps/backend/src/models/Template.ts
```diff
@@ -17,6 +17,7 @@ export class FieldDependency {
 export enum DataType {
   BOOLEAN = 'BOOLEAN',
   DATE = 'DATE',
+  DATE_TIME_RANGE_PICKER = 'DATE_TIME_RANGE_PICKER',
```
yoganandaness:
```suggestion
  DATE_TIME_RANGE = 'DATE_TIME_RANGE',
```

apps/backend/src/resolvers/types/FieldConfig.ts
```diff
@@ -94,6 +94,20 @@ export class DateConfig extends ConfigBase {
   includeTime: boolean;
 }
 
+@ObjectType()
+export class DateTimeRangeConfig extends ConfigBase {
+  @Field(() => String, { nullable: true })
+  defaultDate:
+    | {
+        from: Date | undefined;
+        to?: Date | undefined;
```
yoganandaness: are we going to allow a value without `to`?

validation/src/Questionary/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from './boolean';
 export * from './date';
+export * from './dateTimeRange';
```
yoganandaness: Is this a new file that is not added yet?

yoganandaness (changes requested):

validation/src/Questionary/index.ts
zacharyjhankin: Yeah sorry, deleted this change. That was just left over from dev. Now I only have backend validation

apps/backend/src/resolvers/types/FieldConfig.ts
zacharyjhankin: removed thanks

apps/backend/src/models/Template.ts
zacharyjhankin: updated them all to be this

apps/frontend/src/components/questionary/QuestionaryStepView.tsx
zacharyjhankin: I actually did originally intend for this to be kept in. But removed it as out of scope of PR. I think it could be a good change

apps/e2e/cypress/types/template.d.ts
```diff
@@ -174,6 +174,16 @@ declare global {
         }
       ) => void;
 
+      /**
+       * Creates dynamic multiple choice question. You have to be in edit template view to call this method
```
joeshannon, line 178: nit:t this docstring referrers to a different question type.

joeshannon (approved):
Leaving an approval as this looks good from DLS's perspective but we would request approvals from the other facilities before merging.

yoganandaness (approved):

PR #1649, opened 2026-07-16
feat: role based translations

Scott-James-Hurley:
Ref: https://github.com/UserOfficeProject/issue-tracker/issues/1519

## Motivation

Another facility at STFC would like to use the Xpress functionality but need uses of the term "technique" to be changed to "facility". ISIS would like to retain the use of the term "technique", so a way to change translations for different users is needed.

## How It Works

The first tag associated with the user's current role (we're assuming the role only has one tag) is used to determine what translation to use for the term "technique". A nested object in the translation file with the same name as the tag allows the correct translation to be used.

For example, given tag,name = CLF

and the following in translation.json:

```
"CLF": {
    "Technique": "Facility"
}
```

The term "Technique" will be translated to "Facility"

This PR introduces a new environment variable to point to the translation file in the backend. If it's not set, it could cause the "Technique" column header in the technique proposals export to appear incorrectly.

## How Has This Been Tested

Tested manually

## Changes

- Adds i18n to backend
- Adds tags associated with roles in user context
- Translation added for "Technique" header name for technique proposals spreadsheet export
- Translation added for instances of "Technique" on technique proposals page


apps/frontend/src/components/common/proposalFilters/TechniqueFilter.tsx
```diff
@@ -35,18 +36,24 @@ const TechniqueFilter = ({
...
+  const currentRoleFirstTagName = currentRoleTags?.length
+    ? currentRoleTags[0].name
+    : '';
+
   return (
     <>
       <FormControl fullWidth>
         <InputLabel id="technique-select-label" shrink>
-          {t('Technique')}
+          {t(`${currentRoleFirstTagName}.Technique`)}
```
jekabs-karklins: suggestion: We should take into account the scenario where the user has no tags. So that it does not render literally `.Technique`

apps/backend/src/models/Role.ts
```diff
@@ -23,6 +25,7 @@ type RoleBase = {
   title: string;
   description: string;
   isRootRole: boolean;
+  tags?: Tag[];
```
jekabs-karklins, line 28: question: I can see this is populated only in STFCDataSource, do you know if this has a potential to break something for other facitlities?

Scott-James-Hurley: Currently, the tags are only used in the user context for deciding what translation to use so I can't find a scenario where the tags property not being set breaks something. There's more work to be done relating to roles and tags that I'm hoping to use this property for. I can move the logic of fetching tags into the postgres user datasource if other facilities are interested in the functionality, or for the sake of consistency. I'd love to hear you opinion on this.

jekabs-karklins (approved):

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -694,7 +699,7 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
 
   const techniquesColumns = () => [
     {
-      title: 'Technique',
+      title: i18n.t(`${currentRoleFirstTagName}.Technique`),
```
ACLay: The i18n call in `TechniqueFilter.tsx` was changed to not have the `.` after Jekab's review, but it's still here in this use.

ACLay (approved):

PR #1656, opened 2026-07-20
feat: diamond specific FAP spreadsheet columns

GrantDLS:
## Description

We want to include some different fields in the FAP spreadsheet download, so this change adds in the new fields and renames some others. It uses dependency injection to introduce this functionality so that existing configurations are unaffected.

## Motivation and Context

The `getDataRow` function was changed from positional parameters to a single `FapDataRowInput` object because the function is selected through dependency injection and different facility implementations require different inputs.

The facility-specific implementations require different subsets of proposal data, and the positional API became fragile as new values such as the instrument ID were introduced. A named object prevents argument-order errors, provides a stable interface across implementations, and allows each facility to consume only the fields it requires.

The reason this worked before is because the function was being called with 14 parameters. The default implementation only needed 11 params and so would just ignore the additional 3. Following this same approach meant that the DLS implementation wasn't able to only accept 12 params - it needed to accept the extra three from the STFC implementation, plus the extra instrument ID that was needed.

## How Has This Been Tested

Extra cypress fixtures have been added to test the inclusion of the new columns and the renamed ones.


apps/backend/src/middlewares/factory/xlsx.ts
```diff
@@ -21,6 +21,11 @@ import {
 } from '../../factory/xlsx/proposal';
 
 const fapDataColumns = container.resolve<string[]>(Tokens.FapDataColumns);
+const callExtraFapDataColumns = container.isRegistered(
+  Tokens.CallExtraFapDataColumns
+)
+  ? container.resolve<string[]>(Tokens.CallExtraFapDataColumns)
+  : CallExtraFapDataColumns;
```
zacharyjhankin: I think CallExtraFapDataColumns should be renamed to DefaultCallExtraFapDataColumns just to make it clear what is going on here.

apps/backend/src/factory/xlsx/fap.ts
zacharyjhankin, line 105: Just want to note that Diamond is inconsistent with our naming. I guess in this context we should use their firstname, and in emails we should use their preferred name. So the code is working how we want. But it feels like something that we need to be weary of. Especially if we ever need support importing in the future. (though obviously that is a very bad idea)

https://github.com/DiamondLightSource/user-office-core/pull/25/changes

apps/backend/src/middlewares/factory/xlsx.ts
GrantDLS: Yes, I agree. I've changed the exported const from callFaps.ts to be called DefaultCallExtraFapDataColumns and in the code you've referenced above it uses this new name for if there is no custom resolver for the token.

apps/e2e/cypress/fixtures/exampleCallFapExportDLS.json
TCMeldrum, line 1: Do you ever run the cypress tests in `DEPENDENCY_CONFIG === 'dls'`? if not i don't think we need this file?

zacharyjhankin (changes requested):
There is a bug in the code where the comments to users and the comments to mangement are mixed up

image

apps/backend/src/factory/xlsx/dls/DLSFapDataColumns.ts
```diff
@@ -0,0 +1,20 @@
+export const DLSFapDataColumns = [
+  'Proposal Reference Number',
+  'Proposal Title',
+  'Principal Investigator',
+  'Institution',
+  'Instrument',
+  'Instrument available Time',
+  'TA Recommended Time',
```
zacharyjhankin: This needs to be changed to "Technical Review Recommended Time", as confirmed with Marcos. This is to match the UI as "TA" makes no sense.

zacharyjhankin:
I think Average Score can't be zero when there is no data, it should be <missing>. At least for our implementation of it

apps/backend/src/factory/xlsx/dls/DLSFapDataRow.ts
```diff
@@ -0,0 +1,72 @@
...
+export function populateDLSRow(row: DLSFapRowObj): (string | number)[] {
+  return [
+    nullFieldHelper(row.propShortCode),
+    nullFieldHelper(row.propTitle),
+    nullFieldHelper(row.principalInv),
+    nullFieldHelper(row.piOrg),
+    nullFieldHelper(row.instrName),
+    nullFieldHelper(row.instrAvailTime),
+    nullFieldHelper(row.techReviewTimeAllocation),
+    nullFieldHelper(row.fapTimeAllocation ?? row.techReviewTimeAllocation),
```
zacharyjhankin: Surely this is not right, they are completely different values

apps/backend/src/factory/xlsx/dls/DLSFapDataRow.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export async function getDLSDataRow(
+  input: FapDataRowInput
+): Promise<DLSFapRowObj> {
+  const { proposalPk, proposerId, instrumentId } = input;
+  const userDataSource = container.resolve<UserDataSource>(
+    Tokens.UserDataSource
+  );
+  const pi = proposerId
```
zacharyjhankin: I don't like pi here, its not clear if they are a principleInvestigator, or a propserID. I think they are the same thing. If I am right should be principleInvestigator

apps/backend/src/factory/xlsx/FapDataRow.ts
```diff
@@ -1,19 +1,41 @@
...
-): RowObj {
+export type FapDataRowInput = {
+  proposalPk: number;
+  piName: string;
+  proposalAverageScore: number;
+  instrumentName: string;
+  instrumentAvailabilityTime: number;
+  fapTimeAllocation: number | null;
+  proposalTitle: string;
+  proposalId: number | null;
```
zacharyjhankin, line 14: Does this need to be nullable?

apps/backend/src/factory/xlsx/dls/DLSFapDataRow.ts
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+  return {
+    ...getDataRow(input),
+    piOrg: pi?.institution,
+    instrumentRequestedTime,
+  };
+}
+
+export function populateDLSRow(row: DLSFapRowObj): (string | number)[] {
```
zacharyjhankin: We don't need to pass the DLSFapRowObj type here because you're not even using the new instrumentRequestedTime value. That is being used here callFapDLSPopulateRow

apps/e2e/cypress/fixtures/exampleCallFapExportDLS.json
GrantDLS: Good catch. I added it with the intention but we don't currently run dls specific tests so I'll remove it.

apps/backend/src/factory/xlsx/dls/DLSFapDataColumns.ts
GrantDLS: Thanks, updated. I've resolved the ordering of the FAP meeting comments fields, too.

apps/backend/src/factory/xlsx/dls/DLSFapDataRow.ts
GrantDLS: Yeah, this is how the default one is set up, and I don't fully understand the difference so I left it as it is.

However, it's clear from our column naming that it should be the FAP time allocation, so I've updated it to be just that, especially since the column before it is the tech review time allocation.

apps/backend/src/factory/xlsx/dls/DLSFapDataRow.ts
GrantDLS: On a scale of 1 to 10, I like `pi` 3.14159 out of 10. Which isn't a lot. So I've renamed it to `principleInvestigator`.

apps/backend/src/factory/xlsx/FapDataRow.ts
GrantDLS: You're right there does seem to be some inconsistencies with the `FapDataRowInput` type. Some fields are nullable, whilst others aren't. I think we should leave it as since it's not causing any issues right now.

apps/backend/src/factory/xlsx/dls/DLSFapDataRow.ts
GrantDLS: Yeah, true! Reverted to `RowObj`.

TCMeldrum (approved):

zacharyjhankin (approved):
Nice

PR #1657, opened 2026-07-20
chore: graphql schema change check

### LINKED ISSUE #1613: Create GitHub action to check for schema changes
As part of #1612  

A GitHub action needs to be set up that will check for breaking schema changes. 
This is any edits or deletions to the existing schema.
The breaking changes label (or another way to tag a PR) should be added to any PR that contains any of these schema changes to a reviewer can check if any downstream systems will be affected.

TCMeldrum:
Closes https://github.com/UserOfficeProject/issue-tracker/issues/1613

## Description

This will check and show developers where there are breaking schema changes if there are the job will fail. Once all team are aware and good with the changes we can add the label breaking:changes and the job will pass.

When a developer makes schema changes when the backend is brought up it should automatically updated the schema.gql file. They will then need to commit it for the check to take place.

See a example of it working here: https://github.com/TCMeldrum/user-office-core/pull/1

The test will always fail on this PR as the schema does not exist on develop yet for the job to compare too.


yoganandaness (commented):
This is good but I sense some limitations. Requesting your thoughts as well @TCMeldrum 

1. The gql file is >70kb. Is it optimal to include? I understand that the graphql inspector action requires the gql file to be available to perform comparison. Is it possible to do without the gql file but based on changes on the resolvers(query|mutation|type).
2. If the PR comes with breaking:change, then this action will pass even though there is a graphql change. Correct me if I am wrong.

TCMeldrum:
> This is good but I sense some limitations. Requesting your thoughts as well @TCMeldrum
> 
>     1. The gql file is >70kb. Is it optimal to include? I understand that the graphql inspector action requires the gql file to be available to perform comparison. Is it possible to do without the gql file but based on changes on the resolvers(query|mutation|type).
> 
>     2. If the PR comes with breaking:change, then this action will pass even though there is a graphql change. Correct me if I am wrong.

1. It only seems to work of the `gql` file which is why I included it. Not use the action and spin up containers to dynamically generate the file but then we loss the report made by the file. I thought this way was less error prone and more maintainable. 

2. It would fail until we add the label `breaking:schema-all-facilities-approved` (or at least that is how it meant to work) that way all facilities must have approved it. It is a bit of extra red tape but would mean all facilities will know of any breaking change.

ACLay (commented):
I definitely like the idea of having a check for schema changes that operates on PR's like this, but I'm not sure about this implementation. It feels like there's multiple ways the check could end up not being run. Like if the code's not run directly, or not run directly from the checked out directory (being built and run with the production docker image, or if someone is just running test suites and not running the app to test their changes) the updated schema file won't be generated, or alternatively if the schema changes aren't committed then the check won't flag anything.

PR #1665, opened 2026-07-22
fix: deeplinks to Proposal and Experiment

shivoomiess:
## Description
Implemented a fix for deeplinks to Proposal and Experiment in the frontend for the User Officer.

## Motivation and Context
Before this change, a Proposal or Experiment not already loaded in the relevant table rows did not open up the details modal if directly accessed from a link because the data was never fetched. 

The change was required to fix the issue where deeplinks to Proposal and Experiment were not functioning as expected. This was breaking links from the Experiments page to Proposals that were not already loaded in the Proposals Table (due to pagination) on the ESS dev instance because it has much more data than a local dev environment.

## Changes
1. In `ExperimentsTable.tsx`, 
- the experiment identifier in the URL is now parsed correctly to ensure that the correct experiment data is loaded when accessed via a deeplink. 
- A fallback mechanism is added to fetch experiment data by primary key if it is not found in the currently loaded page of results.
2. In `ProposalTableOfficer.tsx`, similar changes are made to correctly parse the proposal identifier from the URL and fetch the proposal data. A fallback mechanism is also added here to fetch proposal data by primary key if it is not found in the current page of results.
3. Both changes ensure that the correct data is loaded when accessed via deeplinks, improving the user experience.

## How Has This Been Tested?

A dump of the Dev DB was taken today and we ensured that all Proposals were clickable and redirected to the correct Proposal View. For experiments, since there are currently no links on the frontend redirecting to a particular experiment, we tried to access an experiment's details using a deep link.

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5680](https://jira.ess.eu//browse/SWAP-5680)


jekabs-karklins (approved):
👍🏻

shivoomiess:
@zacharyjhankin @TCMeldrum this one too please, alongside #1615 #1634 and #1637

mutambaraf (approved):
LGTM

mutambaraf:
We could also had an e2e to test "opening a proposal via reviewModal parameter when it's not in the current page" if its not there already

PR #1680, opened 2026-07-27
feat: add validation logic to config of questions

zacharyjhankin:
## Description

Adds a new way of adding validation before any configs are written to the database. Before now you could put anything in without any validation being done.  This PR adds the architecture needed to add validation to all question configs easily moving forwards. 

## Motivation and Context

Before via graphQL queries using the config update query there was no logic in the backend to prevent any form of object being written to the database. This means that we could get invalid data and cause problems down the line when it is read. 

## How Has This Been Tested

Adds unit testing. Tested manually saving the config and it still gets written to database.

How to test my code:

Remember to add auth bearer token.

image


joeshannon:
I don't know much about the validation in this project, but how does this relate to the code in the validation module? E.g. `validation/src/Questionary/instrumentPicker.ts`

joeshannon:
> I don't know much about the validation in this project, but how does this relate to the code in the validation module? E.g. `validation/src/Questionary/instrumentPicker.ts`

Think I've answered my own question - the validation there is for the answers but this is the config itself. Not sure if the specific validation code should be there or 

The idea in general seems useful though.

zacharyjhankin:
> > `validation/src/Questionary/instrumentPicker.ts`



> > I don't know much about the validation in this project, but how does this relate to the code in the validation module? E.g. `validation/src/Questionary/instrumentPicker.ts`
> 
> Think I've answered my own question - the validation there is for the answers but this is the config itself. Not sure if the specific validation code should be there or
> 
> The idea in general seems useful though.


I _think_ that the validation folder is left over from when the backend and frontend were in different repos. So they needed one place where the validation could be re-used from the backend and the frontend. 

In this situation I have my validation logic exactly where it needs to be, so it it more simple and more readable than having it abstracted into the validation folder. If I needed the same validation logic in frontend and backend, then maybe it would make more sense.

@joeshannon  

https://github.com/UserOfficeProject/user-office-validation
https://github.com/UserOfficeProject/user-office-core/pull/1347

GrantDLS (approved):

apps/backend/src/datasources/postgres/TemplateDataSource.ts
```diff
@@ -682,6 +684,8 @@ export default class PostgresTemplateDataSource implements TemplateDataSource {
       dependenciesOperator,
     } = args;
 
+    await validateConfigBeforeWrite(config, questionId);
```
yoganandaness: Overall, the idea is great. Great to have a validation for Config being a PG json field. 

It is just a small concern that the data source layer does not usually contain the business logic, but just execute the queries. It is the business layer(Mutation) that the validation and processing predominantly take place. Having said that, would it be possible to move the validation to the Mutation layer?

yoganandaness (changes requested):

apps/backend/src/datasources/postgres/TemplateDataSource.ts
zacharyjhankin: Thanks yeah I see how the architecture works now. That makes a lot of sense. Thanks you :)

zacharyjhankin: should be good now

zacharyjhankin:
Resolved @yoganandaness comments. I think this is ready to be reviewed/merged. @ACLay

yoganandaness (approved):

PR #1681, opened 2026-07-27
feat: add allocation time unit to instrument picker question

### LINKED ISSUE #1629: Add call time unit to instrument picker

When a facility user is submitting a proposal that includes an instrument picker configured with "request time" the box does not show which time units the request is in:

Image

### Steps to reproduce (if it's a bug).

- Create template including instrument picker with request time
- Complete the template, selecting one or more instruments

joeshannon:
## Fixes

This fixes https://github.com/UserOfficeProject/issue-tracker/issues/1629

## Changes

Before:
619480665-ebb885fb-e920-4f6c-83d4-936b7bdee45a

After:
Screenshot From 2026-07-27 17-18-45


apps/backend/src/models/questionTypes/InstrumentPicker.ts
```diff
@@ -70,15 +72,25 @@ export const instrumentPickerDefinition: Question<DataType.INSTRUMENT_PICKER> =
           true
         );
 
+        const callDataSource = container.resolve<CallDataSource>(
+          Tokens.CallDataSource
+        );
+        const call = await callDataSource.getCall(callId);
+        if (!call) {
+          throw new Error('Call not found');
```
zacharyjhankin: I see this is in a try block, but if the try fails then it just errors out. Maybe we'd rather go back to  fallBackConfig, if we can't get the call? or still show the instrument picker on screen, just without a allocatedTimeUnit?

zacharyjhankin:
I am a bit confused why we actually need to store the allocationTimeUnit as a config with the question type. As one question could be used for multiple different calls, so could have multiple different allocationTimeUnits that we need to show to users. 

Surely it would make more sense to generate the allocationTimeUnit that is relevant to the user purely in the frontend by calling a graphQL query and fill in the string at that point in time. Then you wouldn't need to edit the config and put the value in there. 

What we are risking by putting the allocationTimeUnit in the database here, is that the value in the database is wrong. So we need to verify that it is correct at runtime when the user loads the screen. At that point we wouldn't need it in the database anyway

joeshannon:
> I am a bit confused why we actually need to store the allocationTimeUnit as a config with the question type. As one question could be used for multiple different calls, so could have multiple different allocationTimeUnits that we need to show to users.
> 
> Surely it would make more sense to generate the allocationTimeUnit that is relevant to the user purely in the frontend by calling a graphQL query and fill in the string at that point in time. Then you wouldn't need to edit the config and put the value in there.
> 
> What we are risking by putting the allocationTimeUnit in the database here, is that the value in the database is wrong. So we need to verify that it is correct at runtime when the user loads the screen. At that point we wouldn't need it in the database anyway

Note that modifications made by transformConfig do not get stored in the database:
```js
  /**
   * Function to transform the config value before sending back to the front end.
   * Ideal for changes in config at run time.
   */
  readonly transformConfig?: (
    config: QuestionDataTypeConfigMapping<T>,
    callId?: number
  ) => Promise<QuestionDataTypeConfigMapping<T>>;
```
I've used the same approach as is used for the instruments in the picker where the available instruments are dynamically set based on the call.

I agree with your comment in general, although I'm not sure what an alternative solution would look like. The complication arises due to the fact that a `question` doesn't directly have a relation to a specific call. Once there is an `answer` associated (i.e. the user presses save) it is possible - this is actually already implemented for `InstrumentPickerAnswerRenderer`. However it's entirely possible I've missed an alternative  approach!

joeshannon:
The latest version doesn't make any backend changes.

zacharyjhankin:
I think there is a bug where it doesn't seem to be working the first time you make a proposal, but then after saving it, and then entering in the second time it does show? 

I added logging:  

`state?.proposal?.call `

On  the line after:

` const callAllocatedTimeUnit = state?.proposal?.call?.allocationTimeUnit;`


You can see in the console in the video that the callAllocatedTimeUnit is missing on the first load of the screen, but is supplied the second time. So seems like the backend has some bug where the value is missing for some reason some of the time?
 
https://github.com/user-attachments/assets/dfbf5ca7-bf26-4b24-a158-0654afde5199

joeshannon:
Thanks for looking at this. Yes the structures are populated from different queries on creation compared to later on.

I've now added allocationTimeUnit to getCallCloseDetails which is the query to get the call details on proposal creation. When the proposal is viewed after creation the units are populated via
the getProposal query.

apps/frontend/src/components/questionary/questionaryComponents/InstrumentPicker/QuestionaryComponentInstrumentPicker.tsx
```diff
@@ -71,6 +73,8 @@ export function QuestionaryComponentInstrumentPicker(
   const [stateValue, setStateValue] = useState<
     Array<InstrumentIdAndTime> | InstrumentIdAndTime
   >(() => processInstrumentPickerValue(answer, config));
+  const { state } = useContext(QuestionaryContext) as ProposalContextType;
```
zacharyjhankin, line 76: Just as a note the other places in the codebase that use QuestionaryContext has this pattern:

`const { state } = useContext(QuestionaryContext) as ProposalContextType;
  if (!state) {
    throw new Error(createMissingContextErrorMessage());
  }`

zacharyjhankin (approved):
Passed my manual testing

yoganandaness (approved):

ACLay (approved):
Looks useful

PR #1682, opened 2026-07-27
chore: add details on how to follow the release process

zacharyjhankin:
## Description

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,10 @@ Please consider these guidelines when filing a pull request:
 
 ## Creating releases
 
+To create a release you must create a pull request from the develop branch into 'master' using the 'Rebase and merge' option. Once this is approved it will automatically run the pipeline to generate a new release. [example](https://github.com/UserOfficeProject/user-office-core/pull/1621)
+
+After this, create a new pull request merging 'Master' into 'Develop' using the 'Rebase and merge' option. There should be no files changed, but the commits listed. This is to ensure the branches remain in sync. [example](https://github.com/UserOfficeProject/user-office-core/pull/1623)
+
 UserOfficeProject/user-office-core uses [semantic-release](https://github.com/semantic-release/semantic-release)
 to release new versions automatically.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -12,6 +12,10 @@ Please consider these guidelines when filing a pull request:
 
 ## Creating releases
 
+To create a release, you must create a pull request from the develop branch into 'master' using the 'Rebase and merge' option. Once this has been approved it will automatically run the pipeline to generate a new release. [example](https://github.com/UserOfficeProject/user-office-core/pull/1621)
```
joeshannon: Are we sure about the 'Rebase and merge' option?
I would have thought it would be a normal merge. For example if there were hotfixes on `master` between releases this option would 'duplicate' the commits from `develop` by reapplying them on top of  `master`?

simonfernandes (commented):
This is a nice change generally, though I have the same query as Joe, as I would normally use the standard merge commit option for both the release and merge-back.

My understanding is also that we'd potentially end up with duplicate commits due to the rebase causing differing commit hashes, which could cause awkward merges.

It'd be good to get another opinion from someone on the ESS side. I think @jekabs-karklins might be a good person to ask. 🙂

CONTRIBUTING.md
zacharyjhankin: @simonfernandes 

I have now tested it and you are correct we do want to use the "create merge commit" option. 

image


The rebasing and merge technique should only be done if you are NOT merging from master back into develop post release. But as we are, github needs to identify the entire merge as one thing from the history so needs the merge commit. So that was my mistake, I should have tested it. 



But as to @joeshannon s comment, you're right. If you do any hotfixes on master, then cherry-pick to develop, then PR from develop into master you will get duplicate commit issues. But that happens with both types of merge technique. 

I would suspect that if you merge committed your hotfixes from master into dev it would work. But that's out of scope here.

jekabs-karklins:
@simonfernandes Although we currently are not using master branch for releases so this does not affect us directly, as discussed a merge commit in my opinion will serve the best so not only it saves commit history so that the version number for release can be derived from conventional commits messages but also sha is preserved.

CONTRIBUTING.md
simonfernandes: Thanks for doing this research and confirming 🙂

simonfernandes (approved):

PR #1683, opened 2026-07-28
refactor: unify people selectors into a single PeopleSelectorModal

shivoomiess:
## Description

Follow-up to #1634 addressing @yoganandaness's review. There were two people selectors in the codebase; this PR removes the old one, converts its last two consumers to the new selector, and drops the now-misleading `Proposal` prefix from the surviving component.

## Motivation and Context

@yoganandaness raised two points on #1634:

1. The old `PeopleSelectorModal` was only still used by the Instruments table, so it could be converted to the new selector as well (without invites for now).
2. `ProposalPeopleSelector` is not proposal-specific, so the name should lose the `Proposal` prefix once the old selector is gone.

Both are done here. While converting the Instruments table I found the Techniques table used the old selector for the same "assign scientist" flow, so it is converted too — otherwise the old component could not be deleted.

## Changes

1. Converted instrument and technique scientist assignment (`InstrumentTable.tsx`, `TechniqueTable.tsx`) to the new autocomplete selector. Email invites stay off, as requested. Already-assigned scientists are passed as `excludeUserIds` and the `INSTRUMENT_SCIENTIST` role filter is preserved.
2. Deleted the old table-based `PeopleSelectorModal.tsx`.
3. Renamed `ProposalPeopleSelectorModal` to `PeopleSelectorModal` (file, component and props type) and updated its imports in `FapMembers`, `UserManagementTable` and `PrincipalInvestigator`.
4. Fixed a bug where typing a valid email and pressing Enter added a chip even when `allowInviteByEmail` was off. The chip was silently dropped on submit, so the user got a success toast while nobody was actually assigned. This affected instruments, techniques, FAP members and the principal investigator picker, where it also produced an `undefined` PI.
5. Fixed the close-with-confirm path not clearing the query and selection, which meant a discarded selection reappeared the next time the dialog opened.
6. The selector is now only mounted while open in the instrument and technique tables, so those pages no longer fire user lookup queries on page load and on every table re-render.
7. Updated the Cypress tests for the new flow (`instruments`, `techniques`) and extracted the duplicated `addUserThroughModal` helper into `cypress/support/peopleSelector.ts`.
8. Fixed the FAP member assignment step in `eventLogs.cy.ts`, which still drove the old checkbox modal and would have failed on #1634 as it stands.

## Fixes Jira Issue

[https://jira.ess.eu//browse/SWAP-5652](https://jira.ess.eu//browse/SWAP-5652)

## Depends On

This PR targets `SWAP-5652-refactor-FAP-table` (#1634) as its base branch.

## Notes for reviewers

Two things worth a deliberate decision:

- **Discoverability.** The old selector showed a paginated, sortable table of every instrument scientist. The new one needs 3 characters before it queries, and with an empty box it falls back to the officer's previous collaborators, which for a User Officer assigning scientists is usually not a useful list. An officer who does not already know the scientist's name has a harder time finding them. **However,** we can safely assume that a User Officer knows exactly who the instrument scientists are.

- **Dead code in `PeopleTable`.** With the old selector gone, `PeopleTable`'s `emailSearch`, `selectedUsers` and `userRole` props are no longer passed by any caller, which makes the email search bar and its `getUserByEmail` lookup unreachable. I left that cleanup out to keep this PR reviewable.


jekabs-karklins (approved):
Looks very good 💯

shivoomiess:
@yoganandaness let us decide how to close this soon

PR #1685, opened 2026-07-28
feat: allow replaying multiple selected status action logs

### LINKED ISSUE #1487: Allow multiple select for status action reruns
Currently you can only rerun one status action at a time. 
There was a change to the proposal PDF and 56 actions needed to be rerun and we had to run these individually. 
Having multiple select available along with improved filtering (#1486 ) would have made this a very easy task.

mutambaraf:
## Description

Adds checkbox row selection to the Status Actions Logs table (used by both the Email and Proposal Download logs pages) so a User Officer can select multiple log rows and replay them all in a single bulk action, instead of replaying one row at a time.

The selected rows persist across pagination (stored in the URL's `selection` query param), so navigating between pages doesn't lose the selection. A badge on the bulk replay button always reflects the true total number of selected rows, even while viewing a page that has no rows checked locally. After a successful bulk replay, the selection is cleared and the table refreshes.

## Motivation and Context

Previously, User Officers could only replay one status action log at a time via the per-row replay icon. When many status actions failed (e.g. after an outage) and needed to be replayed, this meant repeating the same manual confirmation for every single row. Bulk selection removes that repetitive work.

## How Has This Been Tested

- Added new Cypress e2e tests in `apps/e2e/cypress/e2e/statusActions.cy.ts`:
  - Selecting multiple rows and running a bulk replay, verifying the confirmation dialog, success notification, and that the selection is cleared afterwards.
  - Selecting rows, paginating away and back, and verifying the selection (and the bulk replay button/badge count) persists correctly even on pages with no local selection.
- Manually verified in the browser against both the Email and Proposal Download status actions logs pages.

## Fixes

Closes [#1487](https://github.com/UserOfficeProject/issue-tracker/issues/1487)

## Changes

- `apps/frontend/src/components/statusActionsLogs/StatusActionsLogsTable.tsx`:
  - Enabled row selection (`options.selection: true`) with `onSelectionChange` persisting selected log IDs to the `selection` URL search param, merged correctly across page changes.
  - Added a bulk "Replay selected status actions" action (registered under both `toolbar` and `toolbarOnSelect` positions so it stays visible regardless of whether the current page has any rows checked), showing a badge with the live cross-page selected count.
  - Overrode the `nRowsSelected` toolbar text to reflect the true persisted total selection rather than material-table's per-page count.
  - Always set `tableData.checked` explicitly per row (instead of only when a selection exists) to avoid stale checked state bleeding through after the selection is cleared.
- `apps/frontend/src/graphql/statusActionsLogs/replayStatusActionLogs.graphql` renamed to `replayStatusActionsLogs.graphql` for consistency with its sibling `replayStatusActionsLog.graphql`.
- `apps/e2e/cypress/e2e/statusActions.cy.ts`: added e2e coverage for bulk replay and cross-page selection persistence.

## Depends on

N/A


Scott-James-Hurley (approved):

apps/frontend/src/components/statusActionsLogs/StatusActionsLogsTable.tsx
```diff
@@ -263,6 +275,64 @@ const StatusActionsLogsTable = ({
...
+      return;
+    }
+
+    confirm(
+      () =>
+        api({
+          toastSuccessMessage: 'Status action replay successfully sent.',
+        })
+          .replayStatusActionsLogs({
+            statusActionsLogIds: selectedStatusActionsLogIds.map(
```
yoganandaness: I could not find the graphql variable(statusActionsLogIds) in the apps/frontend/src/graphql/statusActionsLogs/replayStatusActionsLogs.graphql

Correct me if I am wrong.

yoganandaness:
@mutambaraf LGTM. 

Should we remove `replayStatusActionsLog` from `ReplayStatusActionsLogMutation` as a cleanup?

yoganandaness (approved):

simonfernandes (commented):
Question and potential problem I just thought of: what happens if you select multiple of the same failed status actions and replay them? Would it send duplicate emails in this example?

Image

apps/frontend/src/components/statusActionsLogs/StatusActionsLogsTable.tsx
mutambaraf: Looks like it there replayStatusActionsLogs(statusActionsLogIds: $statusActionsLogIds) {

mutambaraf:
> Question and potential problem I just thought of: what happens if you select multiple of the same failed status actions and replay them? Would it send duplicate emails in this example?
> 
> Image

Will add backend fix

simonfernandes (commented):
If I have understood the fix correctly, it will stop status actions with the same status action log id executing, but duplicate failures will usually have different status action log ids since we use a composite key of status action log id, connection id and action id. For example this

image

will actually look like this in the database (different status action log ids)

image

mutambaraf:
> If I have understood the fix correctly, it will stop status actions with the same status action log id executing, but duplicate failures will usually have different status action log ids since we use a composite key of status action log id, connection id and action id. For example this
> 
> image
> will actually look like this in the database (different status action log ids)
> 
> image

@simonfernandes Had a look at this and here is how Email and PDF work during replay, and why selecting all logs on the frontend is safe from duplicate emails/downloads. PDFs have always passed statusActionsLogId: null. That means PDFs have always called .create() to insert a new log row on replay (which is why you were seeing new replay_numbers for PDFs).Emails we pass previous statusActionsLogId  and just update the message and also update if it is succesfull or failed. We  may need to make it consistent between the two. Also we should remember this feature is not for audit is just always to retry an failed action the trail is not that inportant since it is recorded on the proposal events.

simonfernandes (approved):

PR #1686, opened 2026-07-28
refactor: improve logging on Factory download failure

simonfernandes:
## Description

- Fixes errors being swallowed when tracing was enabled. The catch block would never throw and so the code continued until it errored trying to read a non existent stream which caused the ambiguous `ERR_INVALID_ARG_TYPE` error.
- Improves the logging by including the error message from the factory.

Before:
> [2026-07-23T12:02:59.395+01:00] ERROR - Failed to generate the requested file(s) 
>  {"err":{"error":{"code":"ERR_INVALID_ARG_TYPE"},"message":"Could not download generated pdf/proposal"},"originalUrl":"/download/pdf/proposal/31754"

After: 
> [2026-07-28T17:19:27.448+01:00] ERROR - Factory service failed to generate pdf/proposal: Could not create readable stream
> error: large object 628393 does not exist undefined 
>  {"exception":{"name":"Error","message":"Factory service failed to generate pdf/proposal: Could not create readable stream \nerror: large object 628393 does not exist undefined",

## Motivation and Context

We see this vague error occasionally in logs at STFC and it is useful for diagnosing problems.

## How Has This Been Tested

Manual testing checking logs with tracing enabled and disabled.

## Fixes

https://github.com/UserOfficeProject/issue-tracker/issues/1636

## Changes

- A new function `throwFactoryError` is introduced to handle Factory download failures. It provides more detailed error messages, including the download type and the exact error message from the Factory service.
- The error handling in `callFactoryService` function has been updated to use the new `throwFactoryError` function.
- The error logging in the Factory middleware has been enhanced to log the underlying error along with its message, providing more context for debugging.

### DIFF
--- a/apps/backend/src/factory/service.ts
+++ b/apps/backend/src/factory/service.ts
@@ -110,6 +110,7 @@ export default async function callFactoryService<TData, TMeta extends MetaBase>(
               });
               span.setStatus({ code: 2, message: String(err) });
             }
+            throw err;
           } finally {
             span.end();
           }
--- a/apps/backend/src/factory/service.ts
+++ b/apps/backend/src/factory/service.ts
@@ -38,6 +38,27 @@ export type XLSXMetaBase = MetaBase & { columns: string[] };
 
 const ENDPOINT = process.env.USER_OFFICE_FACTORY_ENDPOINT;
 
+async function throwFactoryError(
+  factoryResp: globalThis.Response,
+  downloadType: DownloadType,
+  type: PDFType | XLSXType | ZIPType
+) {
+  const body = await factoryResp.text();
+
+  let parsed: { message?: string } | undefined;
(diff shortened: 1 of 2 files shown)


deepaksftc (approved):

jekabs-karklins:
This is a good change, but I wanted to highlight that recently we merged 
https://github.com/UserOfficeProject/user-office-core/pull/1645

To get to the root cause of the problem we need to look at the logs from the factory itself because I don't think factory returns sufficient logs upon error in many cases.

jekabs-karklins (approved):

simonfernandes:
Thanks @jekabs-karklins. There are a few more errors logged as part of https://github.com/UserOfficeProject/user-office-factory/pull/916 which should also help.

PR #1690, opened 2026-07-30
feat: add configurable decorator to email substitution data

GrantDLS:
## Description

When an email is triggered from a proposal workflow action, there is a default set of data available to the email template. This PR introduces a new dependency config token that can be configured to decorate this substitution data with additional fields.

It also includes a handler for Diamond to decorate with additional call data and fap meeting comments, amongst other things.

## Motivation and Context

This is required by DLS because we need additional information (like the call, for example) to be available for the proposal accepted template.


apps/backend/src/eventHandlers/email/DLS/DLSProposalEmailActionSubstitutionDataDecorator.ts
```diff
@@ -0,0 +1,346 @@
...
+  const technicalAssessments = getTechnicalAssessments(
+    technicalReviews,
+    instruments
+  );
+
+  return {
+    uos_instance: getBaseURL(),
+    proposal,
+    isRapidAccess,
+    machineRequired: true,
```
GrantDLS, line 219: I've defaulted this to true for now since I'm not sure we have an equivalent in UOS. When this is true, the template reads:

"Please ensure that you contact all your team members regarding this beamtime and scheduling."

When it is false, it reads:

"Please ensure that you contact all your team members regarding scheduling."

simonfernandes (approved):
Looks like a nice change 👍

apps/backend/src/eventHandlers/email/DLS/DLSProposalEmailActionSubstitutionDataDecorator.ts
```diff
@@ -0,0 +1,346 @@
...
+    awardedShifts,
+    commentsToUser,
+    technicalAssessments,
+    call: {
+      ...getFormattedAllocationPeriod(call),
+      referenceNumber: call.shortCode.replace(/^AP\s*/i, ''),
```
zacharyjhankin, line 225: I don't think it is a good idea to introduce the concept of allocation periods to UOS, so you should stick to the same logic in the proposalSubmittedHandler where it uses the dates to show the users. Note up to this time users will never have seen this value before, so might not make sense to show them now. Also I don't want us to add a dependency on making the shortcodes stick to a format of AP(ap number) that we want to get rid of in the future.

apps/backend/src/eventHandlers/email/DLS/DLSProposalEmailActionSubstitutionDataDecorator.ts
```diff
@@ -0,0 +1,346 @@
...
+  const isRapidAccess = isRapidAccessWorkflow(proposalWorkflow);
+  const commentsToUser = getCommentsToUser(fapMeetingDecisions);
+  const requestedFacilities = getRequestedFacilities(requestedInstruments);
+  const technicalAssessments = getTechnicalAssessments(
+    technicalReviews,
+    requestedInstruments
+  );
+
+  return {
+    proposal,
```
zacharyjhankin, line 278: This proposal is just the first proposal in the list of proposals that is already submitted to the mail service in mailServiceResponse. So we don't need to add it here, its just confusing. Use the existing proposals data that you have available

apps/backend/src/eventHandlers/email/DLS/DLSProposalEmailActionSubstitutionDataDecorator.ts
```diff
@@ -0,0 +1,346 @@
...
+    isRapidAccess,
+    requestedFacilities,
+    commentsToUser,
```
zacharyjhankin, line 281: Should dis-ambiguate this from proposals.commentForUser that already exists in email object.

yoganandaness:
@GrantDLS I have not reviewed this PR completely yet. But based on the description of the PR, I would like to propose that it is better to have the facility related logic as minimal as possible for the betterment of the project. 

Having said that, if you would like to add call, fap info to the substitution data of the email template, I dont see any issues and it can be added as it is to the common logic. It is upto the Emails to use those data based on how we configure the email template. Correct me if I am wrong. 

@gnyiri @TCMeldrum Kindly let me know if this will affect your facilities, just to be double-sure.

zacharyjhankin:
I think I would generally agree with @yoganandaness

As in at diamond we want to be contributing to the project and unifying our approach with the other collaborators instead of adding bespoke code that really doesn't need to be bespoke. I think this would be better to be written to be available to all members.

zacharyjhankin:
I also think that there is a lot of added complexity for limited benefit of adding the data on top of the email service after the email function has already been called. I think that it would be much more simple add to use the recipientsWithData object that already exists. Then we have the added benefit of letting anyone use the data that you are collecting for these emails.

apps/backend/src/eventHandlers/email/DLS/DLSProposalEmailActionSubstitutionDataDecorator.ts
```diff
@@ -0,0 +1,346 @@
...
+      : [];
+  });
+  const technicalAssessments = getTechnicalAssessments(
+    technicalReviews,
+    instruments
+  );
+
+  return {
+    uos_instance: getBaseURL(),
+    proposal,
```
zacharyjhankin, line 217: This data already exists in defaultSubstitutionData so you have just duplicated it. Which is super confusing when you consume it. It should be left out

zacharyjhankin (changes requested):

apps/backend/src/eventHandlers/email/DLS/DLSProposalEmailActionSubstitutionDataDecorator.ts
```diff
@@ -0,0 +1,346 @@
...
+      Tokens.InstrumentDataSource
+    ),
+    reviewDataSource: container.resolve<ReviewDataSource>(
+      Tokens.ReviewDataSource
+    ),
+  };
+}
+
+function isRapidAccessWorkflow(workflow: { name: string } | null): boolean {
+  return workflow?.name.toLowerCase().includes('rapid') ?? false;
```
zacharyjhankin, line 52: Not sure we can rely on this. Perhaps it is best to change the email to only show data that is already naively in UOS.

GrantDLS:
Abandoning this PR in favour of adding the necessary data to the default array of `EmailReadyType` that is passed to the `sendMail` function.

PR #1691, opened 2026-07-30
feat: add technique proposal rejection comment

ellen-wright:
## Description

refs: https://github.com/UserOfficeProject/issue-tracker/issues/1533

## Motivation and Context

Experiment Scientists would like to add a comment to emails when setting status to unsuccessful to explain their reasoning. Currently someone will reach out separately after this so this will give the user extra information on the proposal decision. 

## Changes

- New database table - similar to Internal Scientist Comment
- Adds new data source for the rejection comment 
- create mutation that checks if any previous comment was made for this proposal - this can only happen if a User Officer sets the status to a different status then back to unsuccessful and give a new comment, and adds the comment for the proposal to the db 
- get function to retrieve the comment and allow it to be shown in an email to the user.
- A new modal that appears when the status of a technique proposal is set to unsuccessful
    - This allows the option for no comment to be set
    - This modal appears before the confirmation to change the status - if the user does not confirm the status change the rejection comment will not be set
- The rejection comment is exposed to be able to be used in emails


apps/backend/src/eventHandlers/workflowEntities/proposal/emailActionHandler.ts
```diff
@@ -405,6 +405,7 @@ const sendMail = async (
               proposalTemplate: recipientWithData.proposalTemplate,
               samples: recipientWithData.samples,
               hazards: recipientWithData.hazards,
+              rejectionComment: recipientWithData,
```
mutambaraf, line 408: Do we not want to pass the `rejectionComment` only here

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -604,60 +667,19 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
-                        title: 'Change status',
-                        description:
-                          'Are you sure you want to change this status?',
-                        confirmationText: 'Yes',
-                        cancellationText: 'Cancel',
-                      }
-                    )();
+                  if (e.target.value === StatusCode.UNSUCCESSFUL) {
+                    // Do you want to add a comment
+                    console.log('HEREEEEEEE!!!!!!!!!!');
```
mutambaraf, line 672: We may need to remove this .

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -347,6 +361,55 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
+              workflowId: workflowId,
+            })
+            .then(({ workflowStatuses }) => {
+              const selectedWorkflowStatus = workflowStatuses?.find(
+                (s) => s.statusId === value
+              );
+              if (!selectedWorkflowStatus) {
+                throw new Error('Selected workflow status not found');
+              }
+              updateProposalStatus(
```
mutambaraf, line 384: Do not we need to await here

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -347,6 +361,55 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
+                primaryKey,
+                selectedWorkflowStatus.workflowStatusId
+              );
+            });
+
+          if (value === StatusCode.UNSUCCESSFUL && comment != '') {
+            api({
```
mutambaraf, line 391: We may also need to await here

apps/frontend/src/components/techniqueProposal/RejectionComment.tsx
mutambaraf, line 1: This looks like is not being used anywhere

PR #1689, opened 2026-07-30
chore: upgrade mui to v9

shivoomiess:
## Description
This PR upgrades Material-UI (Mui) to version 9, along with related dependencies, and adopts the new styling syntax introduced in Mui v9.

## Motivation and Context
The upgrade to Mui v9 is necessary to keep up with the latest improvements and features from the library, such as better performance, a new styling solution, and many more. This also helps in maintaining the quality and modernity of our codebase.

## Changes
- Upgraded `@mui/material`, `@mui/icons-material`, `@mui/system`, `@mui/x-date-pickers` packages to version 9.
- Upgraded `@emotion/react` and `@emotion/styled` to version 11.14.
- Refactored code in multiple files to adopt the new Mui v9 styling syntax.
- Replaced deprecated `useTheme` import with the new one from `@mui/material/styles`.

## Fixes Jira Issue

[https://jira.ess.eu//browse/](https://jira.ess.eu//browse/)

## Tests included/Docs Updated?

- [ ] I have added tests to cover my changes.
- [ ] All relevant doc has been updated

PR #1693, opened 2026-07-30
fix: restore picker theming and make the app shell responsive

shivoomiess:
## Description

Follow-up to the Material UI v5 → v9 and x-date-pickers v7 → v9 upgrade in #1689. That PR was a
faithful mechanical port; this picks up the layout and theming work it left behind, and fixes two
regressions the port introduced that neither `tsc` nor `vite build` can catch.

> **Stacked on #1689 — do not merge before it.**
>
> This branch sits on top of `chore/mui-upgrade`, so the diff below currently includes that PR's
> commits as well. It is targeted at `develop` only because `test-build.yml` and `lint.yml` are
> restricted to `branches: [develop]`, so a PR aimed at `chore/mui-upgrade` gets no build, lint or
> e2e run at all. Once #1689 merges, the merge base moves and this diff narrows to the 15 files
> listed below.

## Motivation and Context

Two of these are genuine v9 regressions rather than improvements, and they are invisible to the
compiler:

- Since pickers v8 the date fields render their own `PickersTextField` instead of a MUI
  `TextField`, so they stopped inheriting the `MuiTextField` `standard` variant and silently fell
  back to `outlined` — leaving every date field looking different to every other input in the app.
- `FapGradeGuide` passes `sm={25}`, which is out of range. Under the legacy Grid that was inert
  because it emitted a nonexistent class. Grid v2 computes width arithmetically, so it became a
  real 208% overflow.

The rest is the groundwork for the upcoming mobile-friendly work: the app shell had hardcoded
desktop assumptions that are wrong on phones.

## Changes

1. **theme:** add `MuiPickersTextField` defaults so date fields inherit the `standard` variant
   again, matching every other input.
2. **theme:** hoist `desktopModeMediaQuery` into `MuiDatePicker`/`MuiDateTimePicker`/`MuiTimePicker`
   defaults and drop the 20 duplicated call-site props — all 20 passed the identical value. Resolves
   the long-standing `NOTE` in `theme.tsx`.
3. **PageLayout:** the AppBar is 56px below `sm` and 64px from `sm` up, so the hardcoded 64px offset
   left an 8px gap on phones. Also drops the fixed `calc(100% - 250px)` width, which assumed the
   drawer was always expanded.
4. **FapGradeGuide:** fix the out-of-range `sm={25}` described above.
5. **useResponsive:** centralise the four hardcoded pixel media queries into
   `hooks/common/useResponsive.ts`. Values are unchanged, so this is not a behavioural change — it
   just gives the mobile work one place to retune.

## How Has This Been Tested?

- `npm run lint` (`tsc --noEmit` + eslint) and `npm run build` pass on top of the rebased
  `chore/mui-upgrade`.
- CI e2e on this PR.
- Still needs a visual pass on the layout changes — items 3 and 5 are layout-only and will not show
  up as test failures.

## Fixes Jira Issue

N/A

## Depends On

#1689 — this branch is stacked on `chore/mui-upgrade` and must not merge before it.

- [ ] Tests added
- [ ] Docs updated


shivoomiess:
Superseded by #1694 (chore/mui-upgrade-pt1). The layout commit remains available on chore/mui-upgrade (#1689).

PR #1703, opened 2026-08-03
feat: add technique proposal rejection comment

ellen-wright:
## Description

refs: https://github.com/UserOfficeProject/issue-tracker/issues/1533
## Motivation and Context
Experiment Scientists would like to add a comment to emails when setting status to unsuccessful to explain their reasoning. Currently someone will reach out separately after this so this will give the user extra information on the proposal decision. 
This PR allows a comment tot be given once the status of a technique proposal is set to 'Unsuccessful'. 

## Changes
- New database table - similar to Internal Scientist Comment
- create mutation that checks if any previous comment was made for this proposal - this can only happen if a User Officer sets the status to a different status then back to unsuccessful and give a new comment.
- get function to retrieve the comment and allow it to be shown in an email to the user.
- A new modal that appears when the status of a technique proposal is set to unsuccessful
    - This allows the option for no comment to be set
    - This modal appears before the confirmation to change the status - if the user does not confirm the status change the rejection comment will not be set
- The rejection comment is exposed to be able to be used in emails
- e2e tests added 


apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -604,60 +667,19 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
-                        title: 'Change status',
-                        description:
-                          'Are you sure you want to change this status?',
-                        confirmationText: 'Yes',
-                        cancellationText: 'Cancel',
-                      }
-                    )();
+                  if (e.target.value === StatusCode.UNSUCCESSFUL) {
+                    // Do you want to add a comment
+                    console.log('HEREEEEEEE!!!!!!!!!!');
```
mutambaraf: We may need to remove this.

apps/backend/src/eventHandlers/workflowEntities/proposal/emailActionHandler.ts
```diff
@@ -405,6 +405,7 @@ const sendMail = async (
               proposalTemplate: recipientWithData.proposalTemplate,
               samples: recipientWithData.samples,
               hazards: recipientWithData.hazards,
+              rejectionComment: recipientWithData,
```
mutambaraf: We need to pass rejectionComment only here

apps/backend/src/queries/ProposalQueries.ts
```diff
@@ -204,6 +204,21 @@ export default class ProposalQueries {
...
+  @Authorized([Roles.INSTRUMENT_SCIENTIST, Roles.USER_OFFICER])
+  async getProposalRejectionComment(
+    agent: UserWithRole | null,
+    proposalPk: number
+  ) {
+    return await this.proposalInternalCommentsDataSource
+      .getProposalRejectionComment(proposalPk)
+      .catch((error) => {
+        return rejection(
+          `Could not get proposal scientist comment proposal: '${proposalPk}'`,
```
mutambaraf: We might wan to say `Could not get proposal rejection comment`

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -347,6 +361,55 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
+
+          if (value === StatusCode.UNSUCCESSFUL && comment != '') {
+            api({
+              toastSuccessMessage:
+                'Proposal rejection comment successfully created',
+            }).createProposalRejectionComment({
+              proposalPk: unsuccessfullPK,
+              comment: comment ?? '',
+            });
+          }
```
mutambaraf: Here if `updateProposalStatus` fails we may need to skip updating the comment.

apps/backend/src/eventHandlers/workflowEntities/proposal/emailActionHandler.ts
jekabs-karklins:
```suggestion
              rejectionComment: recipientWithData.rejectionComment,
```

apps/backend/db_patches/0213_AddProposalRejectionCommentTable.sql
```diff
@@ -0,0 +1,23 @@
+DO
+$$
+BEGIN
+	IF register_patch('0213_AddProposalRejectionCommentTable.sql', 'Ellen Wright', 'Adding proposal rejection comment table', '2026-06-15') THEN
+		BEGIN
+			CREATE TABLE IF NOT EXISTS proposal_rejection_comments (
```
jekabs-karklins, line 6: maybe issue: can one proposal have multiple rejection comments, if not can we add UNIQUE on the proposal?

apps/backend/db_patches/0213_AddProposalRejectionCommentTable.sql
```diff
@@ -0,0 +1,23 @@
...
+				comment_id SERIAL,
+				proposal_pk INT REFERENCES proposals(proposal_pk) ON DELETE CASCADE,
+				comment TEXT NOT NULL,
+				updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+				PRIMARY KEY (comment_id, proposal_pk)
```
jekabs-karklins: suggestion: I suggest having the primary key on comment_id, since that is already unique for every comment

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -347,6 +361,55 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
+            .then(({ workflowStatuses }) => {
+              const selectedWorkflowStatus = workflowStatuses?.find(
+                (s) => s.statusId === value
+              );
+              if (!selectedWorkflowStatus) {
+                throw new Error('Selected workflow status not found');
+              }
+              updateProposalStatus(
+                primaryKey,
+                selectedWorkflowStatus.workflowStatusId
```
jekabs-karklins: issue: the updateProposalStatus accepts 3 arguments, the last one `statusActionsWorkflowConnectionId` is not passed here any more. The consequence is that status actions (if any are configured), will not be executed.

Some context.
When the updateProposalStatus is called on the backend it optionally gets an argument of on which connection the status actions needs to be run. The connection id needs to be specified because there might be multiple connection ids leading to the status (see f.x. here the FEASIBILITY_REVIEW)

if connection id is omitted then no status acctions will be run.

In previous implementation for the technique proposals table it automatically picked the first connection but now it does not pick any. 

I agree that the implemmentation on develop is hacky, and it should be improved to have the same as ProposalsTable where the user can choose which actions to run.


but for now I suggest keeping the existing behavior

apps/frontend/src/components/techniqueProposal/TechniqueProposalTable.tsx
```diff
@@ -347,6 +361,55 @@ const TechniqueProposalTable = ({ confirm }: { confirm: WithConfirmType }) => {
...
+  const handleStatusChange = (
+    value: string,
+    workflowId: number,
+    primaryKey: number,
+    comment: string
+  ) => {
+    if (value) {
+      confirm(
+        () => {
+          api()
```
jekabs-karklins: suggestion: can we await this first `getWorkflowStatuses` call, to make the code little more predictable?

apps/backend/src/datasources/postgres/ProposalInternalCommentsDataSource.ts
```diff
@@ -60,6 +60,43 @@ export default class PostgresProposalInternalCommentsDataSource
...
+      await database('proposal_rejection_comments')
+        .where('proposal_pk', args.proposalPk)
+        .del();
+      const [proposalRejectionComment]: ProposalInternalCommentRecord[] =
+        await database('proposal_rejection_comments')
+          .insert({
+            proposal_pk: args.proposalPk,
+            comment: args.comment,
+          })
+          .returning('*');
```
jekabs-karklins: issue: can we use transaction here to make the change atomic?

apps/backend/src/queries/ProposalQueries.ts
```diff
@@ -204,6 +204,21 @@ export default class ProposalQueries {
         );
       });
   }
+  @Authorized([Roles.INSTRUMENT_SCIENTIST, Roles.USER_OFFICER])
+  async getProposalRejectionComment(
```
jekabs-karklins, line 208: question: this seems to not be used at the moment, is it planned to be user somewhere in the future?

jekabs-karklins (changes requested):

apps/backend/src/queries/ProposalQueries.ts
ellen-wright: This is being used in the utils for the email handler so the rejection comment can be used in the emails

jekabs-karklins (approved):
looks good form my side

apps/e2e/cypress/e2e/techniqueProposals.cy.ts
```diff
@@ -1681,6 +1681,71 @@ context('Technique Proposal tests', () => {
         .should('not.exist');
     });
 
+    it('Scientist should be able to add a rejection comment when changing status to unsuccessful', function () {
```
mutambaraf, line 1684: I’m also aware that User Officers may be able to make these changes. Is it worth confirming whether this is actually the case? If so, we should add a test to cover that scenario as well.

mutambaraf (approved):

apps/backend/db_patches/0213_AddProposalRejectionCommentTable.sql
jekabs-karklins: questions: was intended to remove the UNIQUE?

PR #1715, opened 2026-08-06
feat: add submit management decision button

zacharyjhankin:
## Description

Adds a button to submit the final decision on proposals. 
Also adds an option to all api()s to surpress the snackbar popup.

https://github.com/user-attachments/assets/9477aeef-25ca-461c-9163-69437ee32ea6

## Motivation and Context

So it is easier for the user office to submit the final decision all at once. 
https://jira.diamond.ac.uk/browse/UAS-14586

## How Has This Been Tested

manual testing


TCMeldrum (approved):

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -966,6 +1004,22 @@ const ProposalTableOfficer = ({
...
+      icon: ApprovalIcon,
+      tooltip: 'Submit Final Decision',
+      onClick: () => {
+        confirm(
+          () => {
+            sendProposalsToFinalDecision();
+          },
+          {
+            title: 'Submit Proposals',
+            description: `This action submit ${selectedProposalsData.length} proposal(s).`,
```
joeshannon: I think this could be a bit more informative e.g.
```typescript
`This action will submit the management decision for ${selectedProposalsData.length} proposal(s).`
```

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -966,6 +1004,22 @@ const ProposalTableOfficer = ({
     .map((selectedProposal) => selectedProposal.instruments);
 
   const tableActions: Action<ProposalViewData>[] = [
+    {
+      icon: ApprovalIcon,
+      tooltip: 'Submit Final Decision',
```
joeshannon: I believe the process is termed "management decision" within this system so we should probably stick to that.

@TCMeldrum can you confirm?

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -666,6 +670,40 @@ const ProposalTableOfficer = ({
...
+          await suppressednackbarAPI().administrationProposal(inputvals);
+          successCount += 1;
+        } catch {
+          failCount += 1;
+        }
+      })
+    );
+    refreshTableData();
+    enqueueSnackbar(
+      `${successCount} management decisions were submitted and ${failCount} failed.`,
```
joeshannon: nit: "decision(s)"

joeshannon (commented):
Tested and seems to work well

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
zacharyjhankin: I can confirm, there is lots of references to management decision in codebase, I just didn't see it in the UI when I added this.

The value in the db is: managementDecisionSubmitted

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
zacharyjhankin: Updated

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
zacharyjhankin: Updated all my code to for the management decision terminology

apps/frontend/src/components/proposal/ProposalTableOfficer.tsx
zacharyjhankin: updated

joeshannon (approved):

yoganandaness:
@zacharyjhankin Would it be misleading to notify when some proposals failed and the User Officer will be confused what are those proposals and why did it fail.

yoganandaness:
@zacharyjhankin I dont have access to JIRA to traceback the Problem statement.

zacharyjhankin:
@yoganandaness 

Yes sorry I know you can't read the jira ticket. It basically just says that the user office needs a way to bulk change the management decision submitted button. Which makes sense. In the future we will move over to the public issue tracker, but not implemented yet.



I think that the popup is not _**misleading**_ because it is in fact correctly-leading the user to the understanding that not everything they selected has had their management decision submitted. Though if they are in that state they would probably be confused. I think the solution to this is to have a new column that shows the management decision submitted value, so its clear which ones have been submitted or not. Which we will add in future development.

PR #1717, opened 2026-08-07
fix: pdf omits answers dependent on instrument picker questions

simonfernandes:
## Description
This PR fixes an issue where answers dependent on instrument picker questions were being omitted in the generated PDF.

## Motivation and Context
The issue arose because the current condition evaluator did not properly handle the `INSTRUMENT_PICKER` datatype, causing dependent answers to be erroneously omitted. This fix is crucial to ensure that all relevant data is accurately represented in the generated PDFs.

## Changes
1. Modified `QuestionaryDataSource.ts` to include instrument picker questions and their dependent answers.
2. Updated `ConditionEvaluator.ts` to add an equality check for the `INSTRUMENT_PICKER` datatype.
3. Added a test case in `ProposalModelFunctions.spec.ts` to verify that instrument picker dependencies are correctly satisfied.

## How Has This Been Tested?

- Unit test added
- Manually tested that the question and answer shows up in PDFs, with both default and custom template.

## Fixes Jira Issue

https://github.com/UserOfficeProject/issue-tracker/issues/1645

### DIFF
--- a/apps/backend/src/datasources/mockups/QuestionaryDataSource.ts
+++ b/apps/backend/src/datasources/mockups/QuestionaryDataSource.ts
@@ -26,6 +26,7 @@ import {
   EmbellishmentConfig,
   FieldConfigType,
   FileUploadConfig,
+  InstrumentPickerConfig,
   SelectionFromOptionsConfig,
   TextInputConfig,
 } from '../../resolvers/types/FieldConfig';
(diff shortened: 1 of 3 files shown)


Scott-James-Hurley (approved):

yoganandaness:
@mutambaraf 

I did some test and noticed few strange things in my local machine. Attaching screenshot. 

1. When I select multiple status action items and click replay, the graphql request contains repeating IDs. In the screenshot, you can see that the ID 3 is being passed twice. 
Screenshot 2026-08-12 at 11 45 45

2. Pagination issue. there are totally 22 records. but the second page is packed full instead of just 2. 
Screenshot 2026-08-12 at 11 47 43

3. As you can see in both of the above screenshot, what is ProposalIDs representing and why is it long and repeating between different status action logs. 


I know, its a lot of question. We can talk on call, if that is easier.

simonfernandes:
@yoganandaness I think you may have meant to post this reply in https://github.com/UserOfficeProject/user-office-core/pull/1685 instead 🙂

TCMeldrum (approved):

mutambaraf:
> @mutambaraf
> 
> I did some test and noticed few strange things in my local machine. Attaching screenshot.
> 
> 1. When I select multiple status action items and click replay, the graphql request contains repeating IDs. In the screenshot, you can see that the ID 3 is being passed twice.
> 
> Screenshot 2026-08-12 at 11 45 45
> 2. Pagination issue. there are totally 22 records. but the second page is packed full instead of just 2.
> 
> Screenshot 2026-08-12 at 11 47 43
> 3. As you can see in both of the above screenshot, what is ProposalIDs representing and why is it long and repeating between different status action logs.
> 
> I know, its a lot of question. We can talk on call, if that is easier.
Will have a look at these in https://github.com/UserOfficeProject/user-office-core/pull/1685

yoganandaness:
> @yoganandaness I think you may have meant to post this reply in #1685 instead 🙂

Ah that's my bad. Sorry.

yoganandaness (approved):

PR #1719, opened 2026-08-10
feat: add new events to the message broker

gnyiri:
## Description

Add new events to the message broker. 

## Motivation and Context

At ELI we need these proposal update events. 

## How Has This Been Tested

Local tests. 

## Fixes

N/A

## Changes

messageBroker.ts

## Depends on

N/A

### DIFF
--- a/apps/backend/src/eventHandlers/messageBroker.ts
+++ b/apps/backend/src/eventHandlers/messageBroker.ts
@@ -121,7 +121,7 @@ export function createListenToQueueHandler() {
   );
 }
 
-export const getProposalMessageData = async (proposal: Proposal) => {
+const buildProposalMessageData = async (proposal: Proposal) => {
   const userDataSource = container.resolve<UserDataSource>(
     Tokens.UserDataSource
   );
@@ -231,7 +231,21 @@ export const getProposalMessageData = async (proposal: Proposal) => {
     );
   }
 
+  return messageData;
+};
+
+export const getProposalMessageData = async (proposal: Proposal) => {
+  const messageData = await buildProposalMessageData(proposal);
+
   return JSON.stringify(messageData);
+};
+
+export const getProposalsMessageData = async (proposals: Proposal[]) => {
+  const proposalsMessageData = await Promise.all(
+    proposals.map(getProposalMessageData)
+  );
+
+  return JSON.stringify(proposalsMessageData);
 };
 
 export const getExperimentMessageData = async (experiment: Experiment) => {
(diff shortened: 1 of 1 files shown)


apps/backend/src/eventHandlers/messageBroker.ts
```diff
@@ -327,6 +327,17 @@ export async function createPostToRabbitMQHandler() {
     }
 
     switch (event.type) {
+      case Event.PROPOSAL_STATUS_CHANGED_BY_USER: {
+        const jsonMessage = JSON.stringify(event.proposals);
```
yoganandaness: Can we have the same payload as that for other Events as well?
```suggestion
        const jsonMessage = await getProposalMessageData(event.proposal);
```

yoganandaness (dismissed):

apps/backend/src/eventHandlers/messageBroker.ts
gnyiri: thanks, modified the code accordingly

simonfernandes (approved):

apps/backend/src/eventHandlers/messageBroker.ts
```diff
@@ -231,9 +231,23 @@ export const getProposalMessageData = async (proposal: Proposal) => {
...
+
   return JSON.stringify(messageData);
 };
 
+export const getProposalsMessageData = async (proposals: Proposal[]) => {
+  const proposalsMessageData = await Promise.all(
+    proposals.map(getProposalMessageData)
+  );
+
+  return JSON.stringify(proposalsMessageData);
```
jekabs-karklins, line 248: issue: getProposalMessageData method already stringifys JSON. So this would mean we need to json parse twice.

Maybe you can use `buildProposalMessageData` in the map function?

jekabs-karklins (changes requested):

apps/backend/src/eventHandlers/messageBroker.ts
gnyiri: Right, I fixed it.

jekabs-karklins (approved):

gnyiri:
@jekabs-karklins could you pls resolve the requested changes? I guess thats the reason I cannot merge this pr. Thanks.

jekabs-karklins:
> @jekabs-karklins could you pls resolve the requested changes? I guess thats the reason I cannot merge this pr. Thanks.

I thinik this is due to @yoganandaness requesting changes, but I see you have addressed his comment, so I'll close it

PR #1720, opened 2026-08-12
feat: add button to export data for management decision

zacharyjhankin:
## Description

https://github.com/user-attachments/assets/3ee70de5-36c1-43be-8f7e-6788e9074659

This PR adds a new button that you can access through the calls page to export all the information that is needed for the managers to make a management decision. 

PLEASE REVIEW: https://github.com/UserOfficeProject/user-office-factory/pull/928
(the code works without this as I am using the call_fap string but not ideal)

## Motivation and Context

Managers want a way to export a list of all the data they have about the proposals so they can review them.
This requires the collation of extended data beyond that provided by the existing FAP spreadsheet export For example the comments to users are required to be sent to various people within the organisation for review.

## How Has This Been Tested

Added unit tests for the new complex functions. 
Manual testing


apps/backend/src/factory/xlsx/managementDecision.ts
```diff
@@ -0,0 +1,165 @@
...
+    fapMeetingData[0] && fapMeetingData[0].recommendation
+      ? ProposalEndStatusStringValue[fapMeetingData[0].recommendation]
+      : ProposalEndStatusStringValue[0];
+
+  return {
+    grade: fapreview.average_grade ?? 0,
+    instrumentAllocatedTime: instrumentAllocatedTime ?? 0,
+    xlsxrowdata: [
+      fapreview.proposal_id.toString() ?? '<missing>',
+      fapreview.proposal_pk.toString() ?? '<missing>',
```
DLSValcarcel, line 75: This is not needed for the export. Should be removed, as it has no meaning to the user

apps/backend/src/factory/xlsx/managementDecision.ts
```diff
@@ -0,0 +1,165 @@
...
+      fapreview.instrument_id.toString() ?? '<missing>',
```
DLSValcarcel, line 76: This should be replaced with the instrument name. The ID could be different from the displayed name

DLSValcarcel (commented):
I want to try it on the staging environment with more realistic data but for now it looks good to take to the collaboration

apps/backend/src/factory/xlsx/managementDecision.ts
zacharyjhankin: We need it for the import because the graphQL takes both the proposal_pk and instrument_id

apps/backend/src/factory/xlsx/managementDecision.ts
zacharyjhankin: We need it for the import because the graphQL takes both the proposal_pk and instrument_id. But added the instrument name as well for clarity

apps/backend/src/middlewares/factory/xlsx.ts
```diff
@@ -237,6 +238,62 @@ router.get(`/${XLSXType.TECHNIQUE}/:proposal_pks`, async (req, res, next) => {
...
+      singleFilename: filename,
+      collectionFilename: filename,
+      columns: managementDecisionColumns,
+    };
+
+    const userRole = req.user.currentRole;
+    callFactoryService(
+      DownloadType.XLSX,
+      //XLSXType.MANAGEMENT_DECISION,
+      XLSXType.CALL_FAP,
```
yoganandaness: I believe this must be put back to XLSXType.MANAGEMENT_DECISION? I can see that the Factory PR has been merged.

apps/backend/src/factory/xlsx/managementDecision.ts
```diff
@@ -0,0 +1,166 @@
...
+
+async function generateManagementDecisionRowsByInstrument(
+  agent: UserWithRole,
+  callId: number,
+  instrumentId: number
+) {
+  const faps = await baseContext.queries.fap.dataSource.getFapsByCallId(callId);
+  const rowsOfManagmentDecisions = [] as string[][];
+  for (const fap of faps) {
+    const fapReviewsRows = await collectManagementDecisionData(
```
yoganandaness: Possible to do a Promise.all here? Just curious.

yoganandaness (changes requested):

apps/backend/src/factory/xlsx/managementDecision.ts
zacharyjhankin: yes you can, that is much better fixed

apps/backend/src/middlewares/factory/xlsx.ts
zacharyjhankin: ah yes good call before merge thank you

yoganandaness (approved):

DLSValcarcel (approved):

PR #1721, opened 2026-08-13
fix: fix bug with pdf downloads for tokens with draft propsals

TCMeldrum:
## Description
This PR fixes a bug that was preventing the download of PDFs for tokens with draft properties.

## Motivation and Context
The change is required to allow users to download PDFs proposals that have a insturment picker question but have not answered that question, which was previously not possible due to a bug. 

## Changes
1. Added a conditional check to ensure that the `answerId` is not null before attempting to fetch the call by `answerId` proposal. 

## Fixes Jira Issue


j4mes-4rgent (approved):

ACLay (approved):

simonfernandes (approved):

apps/backend/src/factory/pdf/proposal.ts
```diff
@@ -222,14 +222,21 @@ const addTopicInformation = async (
...
 
-        const call = await callDataSource.getCallByAnswerIdProposal(
-          answer.answerId
-        );
-        const value = instrumentPickerAnswer(answer, instruments, call);
-        updatedAnswers.push({
-          ...answer,
-          value,
-        });
+        if (answer.answerId !== null) {
```
yoganandaness, line 225: Should we also handle the type as well? 

As per the [type definition](https://github.com/UserOfficeProject/user-office-core/blob/05a3b3ad53e10556c4f2ea781754b8b15a0a199c/apps/backend/src/models/Questionary.ts#L15), answerId is non nullable. In the case, I would recommend to refine the type definition as well.

yoganandaness (changes requested):

jekabs-karklins:
@TCMeldrum have you seen comment form @yoganandaness ?

TCMeldrum:
After looking at the Answer type I have made answer id nullable. It was already nullable here: https://github.com/UserOfficeProject/user-office-core/blob/pdf-token-fix-for-draft-proposals/apps/backend/src/resolvers/types/Answer.ts#L12

yoganandaness (approved):

PR #1726, opened 2026-08-17
Staging release 2026 08 17

jekabs-karklins:
## Description
Release 2026-08-17


yoganandaness (approved):

joeshannon:
Hi, we saw this PR and were curious as to how the `staging` branch fits into the release process? Is this an ESS specific procedure?

jekabs-karklins:
> Hi, we saw this PR and were curious as to how the `staging` branch fits into the release process? Is this an ESS specific procedure?

Hello @joeshannon , you are right that this is ESS speciffic procedure and we use the `staging` branch for releasing staging environment. 

That said, we are not happy with this flow but are limited by the state of our current infrastruture. We are working towards releasing from `tag` instead of `branch`

PR #1729, opened 2026-08-18
feat: implement Oauth2 transport creation in SMTPMailService

gnyiri:
At ELI we move from SMTP basic auth to Oauth2. 

Add Oauth2 transport creation in SMTP Mail Service. 

See above.

## How Has This Been Tested

Manual tests + backend tests. 

## Fixes

N/A

## Changes

SMTP Mail Service. 

## Depends on

N/A


apps/backend/package.json
```diff
@@ -36,6 +36,7 @@
     "@apollo/server": "^5.5.0",
     "@apollo/subgraph": "^2.13.2",
     "@as-integrations/express4": "^1.1.2",
+    "@azure/identity": "^4.13.1",
```
jekabs-karklins: issue: the `@azure/identity` package is larger than what we actually need and `@azure/msal-node` is included as phandom dependency
```
$ npm view @azure/identity dependencies
{
  open: '^10.1.0',
  tslib: '^2.2.0',
  '@azure/logger': '^1.0.0',
  '@azure/core-auth': '^1.9.0',
  '@azure/core-util': '^1.11.0',
  '@azure/msal-node': '^5.1.5',  <--- We only need this
  '@azure/core-client': '^1.9.2',
  '@azure/core-process': '^1.0.0',
  '@azure/core-tracing': '^1.0.0',
  '@azure/msal-browser': '^5.5.0',
  '@azure/abort-controller': '^2.0.0',
  '@azure/core-rest-pipeline': '^1.17.0'
}
$ npm view @azure/msal-node dependencies
{ jsonwebtoken: '^9.0.0', '@azure/msal-common': '16.13.0' }
```

can we switch this to @azure/msal-node dependencies ?

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
```diff
@@ -211,24 +251,57 @@ export class SMTPMailService extends MailService {
...
+
+    if (!template) {
+      logger.logError('Email template not found', {
         template: options.content.template,
       });
 
       return { results: sendMailResults };
     }
 
+    const transport = await this.createTransport();
```
jekabs-karklins, line 266: issue: the transport is created on every sendMail invocation and is never closed, which can exhaust the pool. We should add the
```
return Promise.allSettled(emailPromises)
       .then((results) => {
         ...
})
.finally(() => {
  transport.close();
});
```

jekabs-karklins (changes requested):
Hello @gnyiri ,

I have verified that the changes in this PR still works fine in our development environment and emails are properly sent out👍🏻 

I have added two comments that would be very nice to address before merging in.

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
gnyiri: Fixed

apps/backend/package.json
gnyiri: Fixed

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
```diff
@@ -172,22 +125,109 @@ export class SMTPMailService extends MailService {
...
+          type: 'OAuth2',
+          user: process.env.EMAIL_SENDER,
+          accessToken: accessToken,
+        },
+      });
+    }
+  }
+
+  private async createTransport() {
+    if (
```
mutambaraf: Maybe we can have a env like USE_SMTP_OAUTH_2 so that its easier and more straight forward to switch between the two.

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
```diff
@@ -172,22 +125,109 @@ export class SMTPMailService extends MailService {
     return result;
   }
 
-  private getSmtpAuthOptions() {
-    if (process.env.EMAIL_AUTH_USERNAME && process.env.EMAIL_AUTH_PASSWORD) {
-      return {
+  private isTokenExpired(): boolean {
+    if (!this.authToken?.expiresOn) {
+      throw new Error('Invalid token: Missing expiresOn property');
+    }
```
mutambaraf: I think we can do this check when we set authToken "    this.authToken = await cca.acquireTokenByClientCredential(tokenRequest);

    if (!this.authToken || !this.authToken.accessToken) {
      throw new Error('Failed to get access token');
    }"

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
mutambaraf: If we are creating a pool then closing it .My thinking was if possible we can reuse the transport across multiple instances of this class .Maybe by taking it out into a simgleton.

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
gnyiri: sure, good idea

mutambaraf (commented):
Just also did a manual email check and our emails are not being send on my local machine .Having error "{"error":{"code":"ESOCKET","command":"CONN"}".

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
gnyiri: Based on how mapClass is defined, I think that MailService class is a Singleton by default.

jekabs-karklins (approved):
Looks good from my side but @mutambaraf has some more good comments

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
gnyiri: In the recent update I only create a new transport if the token is expired.

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
gnyiri: I updated accordingly. Pls take a look.

apps/backend/src/eventHandlers/MailService/SMTP/SMTPMailService.ts
```diff
@@ -172,22 +126,113 @@ export class SMTPMailService extends MailService {
...
+      });
     }
+  }
 
-    return {
-      secure: false,
-      tls: {
-        rejectUnauthorized: false,
-      },
-    };
```
mutambaraf, line 190: @gnyiri  managed to find the issue,It is we used to set these as defaults and we are now not setting them when creating the basic auth transport

gnyiri: Thanks, fixed.

mutambaraf: Now working thanks.

mutambaraf (approved):
LGTM

PR #1732, opened 2026-08-19
fix: instrument requested time for multi-instrument of FAP page

zacharyjhankin:
## Description

This PR fixes a bug where the instrument requested time would show as zero when it was a multi instrument type. The problem was that the answer could be stored as an object or a list so it now can handle both cases. 

I decided to use database.raw because I think it is easier to read than using the database syntax.

Note was written by me, not an AI. (Though an AI might have done a better job)

## How Has This Been Tested

Manual testing


GrantDLS (approved):
Using param bindings for the query is probably safer (although not necessarily essential since the 2 values are just numbers). Happy either way.

The rest looks good!

TCMeldrum (approved):
👍

PR #1733, opened 2026-08-19
feat: add call, faps, fap meeting decisions and technical reviews to substitution data for proposal status change action emails

GrantDLS:
## Description
This is a rework of https://github.com/UserOfficeProject/user-office-core/pull/1690 to add extra data to the substitution data that's available to the pug email templates.

I've extended the `getEmailReadyArrayOfUsersAndProposals` function to add fapMeetingDecisions, technicalReviews, call and instruments. Instruments was already on `EmailReadyType` but was only being populated if the recipient was `USER_OFFICE`.

## Motivation and Context

Currently the data available to the templates is limited. This PR adds extra information. I've new properties keep to the existing types used within UOS. This will mean more javascript will need to be written in templates to extract and format the data but is probably the best approach for now, otherwise the handler will have to start making assumptions about how each org would present their data in each template which obviously isn't really feasible.


GrantDLS:
@yoganandaness I've added call, faps, fap meeting decisions and technical reviews to the substitution data for emails, but I'm still not sure this data is going to be very useful in its raw form. It'll still require some transformation to obtain usable data from it.

For example, to show a list of awarded shifts for each instrument in an email, I was using the following code that I think would be hard (or even impossible) to replicate in a pug template:

```ts
const instrumentIds = Array.from(
    new Set([
      ...fapProposals
        .map(({ instrumentId }) => instrumentId)
        .filter(
          (instrumentId): instrumentId is number => instrumentId !== null
        ),
      ...(technicalReviews ?? []).map(({ instrumentId }) => instrumentId),
    ])
  );
  const instruments = instrumentIds.length
    ? await instrumentDataSource.getInstrumentsByIds(instrumentIds)
    : [];

  const awardedShifts = instrumentIds.flatMap((instrumentId) => {
    const numberOfShifts =
      fapProposals.find(
        (fapProposal) => fapProposal.instrumentId === instrumentId
      )?.fapTimeAllocation ??
      technicalReviews?.find(
        (technicalReview) => technicalReview.instrumentId === instrumentId
      )?.timeAllocation;
    const instrument = instruments.find(({ id }) => id === instrumentId);

    return typeof numberOfShifts === 'number' && instrument
      ? [{ numberOfShifts, instrument.name }]
      : [];
  });
```

The only way around this I can see is to have code like this in `emailActionHandler.ts`, and add an `awarded_shifts` property to to the `substitution_data` that is passed to `sendMail`, however this can get messy very quickly - especially since awarded shifts isn't applicable to all emails.

There's already an issue with the `instruments` property only being populated if the email recipient is the `USER_OFFICE` email address. This seems like an odd choice and already illustrates how messy it can get having one huge object catering for all email types.

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
```diff
@@ -75,8 +83,61 @@ export type EmailReadyType = {
   proposalTemplate?: string;
   samples?: Answer[];
   hazards?: Answer[];
+  awardedShifts?: unknown[];
```
zacharyjhankin: I assume this is not actually necesserily shifts, so should be awardedTime? awardedAllocationUnits? something like that

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
```diff
@@ -75,8 +83,61 @@ export type EmailReadyType = {
...
+  commentsToUser?: string;
```
zacharyjhankin: Could we call this fapCommentsToUser, just so its clear as there are also admincommentstouser

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
```diff
@@ -75,8 +83,61 @@ export type EmailReadyType = {
...
+    }
+
+    return [
+      {
+        instrument: {
+          name: instrument.name,
+          description: instrument.description,
+        },
+        feasibility: technicalReview.status,
+        assessorsComment: stripHtml(technicalReview.publicComment ?? ''),
```
zacharyjhankin: Does assessorsComment appear in the UOS UI? call is publicComment instead

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
```diff
@@ -250,6 +358,10 @@ export const getEmailReadyArrayOfUsersAndProposals = async (
           proposalTemplate: proposalTemplateName,
           samples: sampleAnswers,
           hazards: hazardAnswers,
+          awardedShifts,
+          commentsToUser,
```
zacharyjhankin: I think going with fapReview as an object that contains commentsToUser is a better thing for expanding it in the future

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
zacharyjhankin, line 144: Is this comment still accurate if we are adding more?

zacharyjhankin, line 151: So just to confirm that right here it is impossible for us to know what email is being sent so we can't only gather the data we need for the emails that need it?

Its just very ugly kinda how the current architecture is

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
GrantDLS: Yes, actually for future flexibility it's probably better to just have the array of `FapMeetingDecision` here and rename the property to fapMeetingDecisions.

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
zacharyjhankin: So it seems like we might have access to the templateID, but I think we would need to check with the collaboration with exactly how that is used. Because they don't _have_ to match anything in hte backend source code. So I guess they might have multiple emails that all use different templateIDs but need the same data and that wouldn't actually work expandability if we need to add every templateID to source code  that needs the same data. 

My way around this would be to add a new property like emailDataType which you select with the UI when making an email so you get to decide which functions are ran when the email is sent. But obvs out of scope of this ticket and maybe needs to be raised to the collaboration

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
GrantDLS: I've looked at the shape of the `TechnicalReview` object and opted to just pass this directly as it is to the template. We should be able to link those to the relevant instrument using javascript's `filter()` function.

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
GrantDLS: Agreed. Done.

apps/backend/src/eventHandlers/workflowEntities/proposal/utils.ts
GrantDLS: I think for this stage, this is something that's best left for future development. Granted it's wasteful to gather data (or attempt to) for an email that doesn't need it, but if we did try to selectively do it by email template or event type, we'd be making assumptions based on what we think would be in the template, which obviously could be anything for each org.

GrantDLS: Good point. Even before this change the comment was kind of out of date due to techniques, samples and hazards being present for all email types. I think it's fine as it is though because the extra fields are still "proposal data" even if not directly on the proposal object.

zacharyjhankin (approved):
yeah looks good to me, all the data is in a good form. Happy with the excess data calls for now. Whole system needs rework

zacharyjhankin:
image

yoganandaness (approved):

PR #1734, opened 2026-08-21
fix: api tokens cannot resolve Questionary.isCompleted

simonfernandes:
## Description

Any API token selecting `isCompleted` on a `Questionary` failed with "Cannot return null for non-nullable field Questionary.isCompleted". It failed at the `@Authorized()` stage which checks if a specific API token permission exists for this query. But since we only make queries that begin with `get` grantable, it was never possible to grant it and so it always failed and returned null.

Rather than adding a new grantable permission for it, this removes the `@Authorized()` decorator instead because:

- `isCompleted` is only reachable via the `Questionary` field resolver, which is only reachable once already authorised to fetch the parent questionary
- The method already does its own authorisation inline checks.

## Motivation and Context

The error was noticed by another team consuming the UOP API.

## How Has This Been Tested

Manual testing

## Fixes

https://github.com/UserOfficeProject/issue-tracker/issues/1620

## Changes

Remove authorisation decorator

### DIFF
--- a/apps/backend/src/queries/QuestionaryQueries.ts
+++ b/apps/backend/src/queries/QuestionaryQueries.ts
@@ -108,7 +108,6 @@ export default class QuestionaryQueries {
     return this.dataSource.getCount(templateId);
   }
 
-  @Authorized()
   async isCompleted(agent: UserWithRole | null, questionaryId: number) {
     const hasRights =
       this.userAuth.isApiToken(agent) ||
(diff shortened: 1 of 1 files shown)


zacharyjhankin:
Do you think it is worth adding a comment, in case someone else thinks that they need to add it back to the source code?

simonfernandes:
@zacharyjhankin It can't hurt 👍

Scott-James-Hurley (approved):

jekabs-karklins (approved):

zacharyjhankin (approved):

PR #1735, opened 2026-08-21
fix: enforce question-level permissions for questionary reads and updates

janosbabik:
- Filter blank questionnaire steps according to the current user’s read permissions.  
- Prevent users from submitting answers to questions they are not permitted to access.  
- Ensure cleanup only affects omitted answers from questions the user can read.


apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -201,9 +201,37 @@ export default class QuestionaryMutations {
...
+      questionField.config.readPermissions.includes(
+        agent?.currentRole?.shortCode ?? ''
+      );
+
+    const unauthorizedSubmittedAnswer = answers.find((submittedAnswer) => {
+      const currentAnswer = currentAnswers?.find(
+        (answer) => answer.question.id === submittedAnswer.questionId
+      );
+
+      return currentAnswer != null && !hasQuestionReadPermission(currentAnswer);
```
jekabs-karklins: issue: we should change && to ||. I have attached the issue as md file.

Additionally what claude found it is much easier to get all the question ids even without guessing because `blankQuestionary` is not Authed at all and returns all questions with their IDs

[PR-1735-delete-hole-walkthrough.md](https://github.com/user-attachments/files/31370097/PR-1735-delete-hole-walkthrough.md)

janosbabik: You are right, this needs to be changed. I will update it. Thx!

jekabs-karklins (approved):

zacharyjhankin:
So if you make an answer required but also give the person no access to see the question it still lets you submit a proposal. (Even though the question is not answered). This seems sub-optimal.

zacharyjhankin:
Maybe we should have an indication to the user that some answers are hidden because of their roles. It feels a bit bad to just hide the questions so they have no idea that they are missing. If a user officer mis-configures this option then there should be some feedback to someone that some questions are hidden

apps/backend/src/mutations/QuestionaryMutations.ts
```diff
@@ -201,9 +201,37 @@ export default class QuestionaryMutations {
       await this.dataSource.getQuestionarySteps(questionaryId)
     ).find((step) => step.topic.id === topicId)?.fields;
 
-    const missingAnswers = currentAnswers?.filter(
-      (oldAnswer) =>
-        !answers.some((answer) => answer.questionId === oldAnswer.question.id)
+    const hasQuestionReadPermission = (questionField: {
```
zacharyjhankin, line 204: I'd put all your new code into its own function to be honest. I think its cleaner to have multiple smaller functions rather than one big one. like what uncle bob says

janosbabik:
> Maybe we should have an indication to the user that some answers are hidden because of their roles. It feels a bit bad to just hide the questions so they have no idea that they are missing. If a user officer mis-configures this option then there should be some feedback to someone that some questions are hidden

Thanks @zacharyjhankin  for the feedback. I wouldn’t show users an indication that some questions are hidden. The purpose of the role based questions is to display questions only to the roles they are intended for, so users shouldn’t be informed about questions that aren’t assigned to their role.

I agree that misconfiguration is a valid concern. However, I think it would be better addressed separately—for example, by validating the questionnaire when the User Office saves it and showing a warning about questions with missing or potentially incorrect role assignments.

Please feel free to refactor the new code into a separate function if you think that would make it cleaner, but keep the current behaviour: if a question has no user role assigned, the user should not see it. Thank you! : )

zacharyjhankin:
I think that the main important thing is that you shouldn't be able to submit a proposal with a required question unanswered, as that allows users to submit invalid proposals, causing issues downstream for the user office, once the call is closed. I guess we don't need to specify why the proposal can't be submitted to the user if that is a concern. 
 
I do accept that that would never be the case with a properly configured templates, but still a concern that we shouldn't let UO/users shoot themselves in the foot. Maybe we can raise this with the UO group to resolve?

janosbabik:
> I think that the main important thing is that you shouldn't be able to submit a proposal with a required question unanswered, as that allows users to submit invalid proposals, causing issues downstream for the user office, once the call is closed. I guess we don't need to specify why the proposal can't be submitted to the user if that is a concern.
> 
> I do accept that that would never be the case with a properly configured templates, but still a concern that we shouldn't let UO/users shoot themselves in the foot. Maybe we can raise this with the UO group to resolve?

I’ll take a closer look at the “required question unanswered” scenario today.

janosbabik:
So the current questionnaire design handles `required` constraints and role-based visibility independently. This allows a question to be required while hidden from the user completing the form, so the topic can still be marked complete and submitted without that answer.

This small fix: When the User Officer configure the questions, the "user-facing" templates now reject required questions unless they are visible to the `user` role or to all roles. "Staff-facing" review templates are excluded because they have different respondents.

Longer term, I think templates should explicitly define their completing roles, validate required-question visibility against those roles, and re-verify all required answers during final submission instead of relying only on topic-completion flags.

Let me know if you have any concerns about this fix or a better idea.
@zacharyjhankin @jekabs-karklins

jekabs-karklins (approved):
I think this is the smallest logical fix so looks good. If we want to make the underlaying system more robust/simple/maintainable we need to have a bigger  more overreaching plan

zacharyjhankin (approved):

zacharyjhankin (approved):
Yes I am happy with this, nice change

PR #1743, opened 2026-08-24
fix: allow teamLead field to be nullable in VisitResolver

jekabs-karklins:
## Description
This PR makes the 'teamLead' field nullable in 'VisitResolver'. 

## Motivation and Context
In the database the team lead always exists — the column is NOT NULL. The nullability here is about the read path, not the data model: teamLead resolves through getBasic, which returns null when the requesting user lacks permission to read that user's details.

While the field was non-nullable, that null triggered GraphQL's non-null propagation: the error bubbles up to the nearest nullable ancestor, Experiment.visit, and removes the entire visit object from the response. A single unreadable team lead therefore made the visit look as though it didn't exist, which is a far larger side effect than the missing field itself.

Making the field nullable confines the loss to the field that's actually restricted.

## Changes
1. Made 'teamLead' field nullable in 'VisitResolver' (apps/backend/src/resolvers/types/Visit.ts).
2. Adjusted related code in frontend components and tests to handle 'teamLead' being null (apps/e2e/cypress/e2e/visits.cy.ts, apps/frontend/src/components/experiment/ExperimentVisitsTable.tsx, apps/frontend/src/components/proposalBooking/CreateUpdateVisit.tsx).
3. Updated GraphQL queries to include 'teamLeadUserId' (apps/frontend/src/graphql/experiment/getAllExperiments.graphql, apps/frontend/src/graphql/experiment/getExperiment.graphql, apps/frontend/src/graphql/visit/fragment.visit.graphql).

## Fixes Jira Issue


yoganandaness (approved):

jekabs-karklins:
@SourangshuSTFC Could you take a look when you get a bit of time?

zacharyjhankin (approved):

PR #1745, opened 2026-08-25
build: raise frontend typescript to 6

shivoomiess:
## Description

Upgrades the frontend to **TypeScript 6.0** and bumps the `@typescript-eslint` toolchain to `^8.66.0` to support it. Includes the small source and config adjustments required for the codebase to type-check cleanly under TS 6.

## Motivation and Context

TypeScript 6 brings stricter type-checking (notably tighter generic variance) and changes to default behaviour (no automatic `@types/*` inclusion, and deprecation errors for options slated for removal in TS 7). Staying current keeps us on supported tooling and surfaces latent type issues early. This is the base of a larger upgrade stack (React 19 → MUI 9 → mobile-responsive).

## How Has This Been Tested

- `npm run generate:sdk` then `tsc --noEmit` — passes with no errors.
- `npm run lint` (frontend `tsc --noEmit && eslint .`) — clean.
- Verified no behavioural change in the touched components (the removed
  `|| 'DRAFT'` branch in `ProposalContainer` was provably unreachable dead code).

## Changes

- **`package.json`**
  - `typescript` `^4.7.4` → `^6.0.3`, moved from `dependencies` to `devDependencies`.
  - `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`, `typescript-eslint` → `^8.66.0` (TS 6 parser support).
- **`tsconfig.json`**
  - Added `"ignoreDeprecations": "6.0"` to silence TS 7 removal errors for `target`, `moduleResolution` and `baseUrl` (temporary; tracked separately).
  - Added `"types": ["node", "vite/client"]` since TS 6 no longer auto-includes `@types/*` globals.
- **`MultistepWizard.tsx`** — `validationSchema` type `Yup.AnyObjectSchema`
  → `Yup.ISchema<unknown>` (concrete object schemas are no longer assignable to `AnyObjectSchema` under TS 6's stricter variance checking; `unknown` avoids `any` and the eslint suppression).
- **`ProposalContainer.tsx`** — removed unreachable `|| 'DRAFT'` fallback (a JSX element is always truthy).
- **`QuestionaryDetails.tsx`** — `TableRowData` fields `JSX.Element | string | null` → `React.ReactNode`.

## Depends on

None — this is the base of the upgrade stack, based directly on `develop`. (PRs #1746, #1747, #1758 are stacked **on top** of this one.)

Follow-up: migrating the deprecated tsconfig options before TS 7 is tracked in UserOfficeProject/issue-tracker#1706.

### DIFF
--- a/apps/frontend/package.json
+++ b/apps/frontend/package.json
@@ -51,7 +51,6 @@
     "react-router-dom": "^6.30.3",
     "reactflow": "^11.11.4",
     "tinymce": "^7.5.1",
-    "typescript": "^4.7.4",
     "yup": "^1.7.1"
   },
   "proxy": "http://localhost:4000",
@@ -92,8 +91,8 @@
     "@types/node": "^22.13.10",
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.1",
-    "@typescript-eslint/eslint-plugin": "^8.46.2",
-    "@typescript-eslint/parser": "^8.46.2",
+    "@typescript-eslint/eslint-plugin": "^8.66.0",
+    "@typescript-eslint/parser": "^8.66.0",
     "@vitejs/plugin-react": "^4.3.0",
     "dotenv": "^16.4.5",
     "eslint": "^9.0.0",
@@ -105,7 +104,8 @@
     "eslint-plugin-react-hooks": "^7.0.1",
     "eslint-plugin-unused-imports": "^4.3.0",
     "prettier": "3.2.4",
-    "typescript-eslint": "^8.46.3",
+    "typescript": "^6.0.3",
+    "typescript-eslint": "^8.66.0",
     "vite": "^5.4.21"
   },
   "engines": {
--- a/apps/frontend/src/components/proposal/ProposalContainer.tsx
+++ b/apps/frontend/src/components/proposal/ProposalContainer.tsx
@@ -159,15 +159,14 @@ export default function ProposalContainer(props: ProposalContainerProps) {
 
   const { submitted, proposalId } = state.proposal;
 
-  let info: JSX.Element | string =
-    (
-      <CopyToClipboard
-        text={proposalId}
-        successMessage={`'${proposalId}' copied to clipboard`}
-      >
-        {proposalId ? `Proposal ID: ${proposalId}` : ''}
-      </CopyToClipboard>
-    ) || 'DRAFT';
+  let info: JSX.Element | string = (
+    <CopyToClipboard
+      text={proposalId}
+      successMessage={`'${proposalId}' copied to clipboard`}
+    >
+      {proposalId ? `Proposal ID: ${proposalId}` : ''}
+    </CopyToClipboard>
+  );
 
   if (!submitted && hasReferenceNumberFormat && proposalId) {
     info = (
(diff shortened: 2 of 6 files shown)


yoganandaness:
## NOTE: It is a Stacked PR

develop <-- typescript upgrade <-- react upgrade <-- mui, table upgrade <-- mobile view

1. Each PR is functional and can be reviewed individually. 
2. Merging is sequential. With the bottom PR being merged into develop, following with next bottom

apps/frontend/package.json
```diff
@@ -51,7 +51,6 @@
     "react-router-dom": "^6.30.3",
     "reactflow": "^11.11.4",
     "tinymce": "^7.5.1",
-    "typescript": "^4.7.4",
```
zacharyjhankin, line 54: do we need to move it out of the dependencies list into dev dependencies in the same PR as updating the version? I guess it is probably fine, but no reason to cause more potential problems for ourselves in one go.
 
Maybe someone in the colab uses it in a prod pipeline for example. 
 
I don't really mind, just worth flagging

apps/frontend/src/components/proposal/ProposalContainer.tsx
```diff
@@ -159,15 +159,14 @@ export default function ProposalContainer(props: ProposalContainerProps) {
...
-  let info: JSX.Element | string =
-    (
-      <CopyToClipboard
-        text={proposalId}
-        successMessage={`'${proposalId}' copied to clipboard`}
-      >
-        {proposalId ? `Proposal ID: ${proposalId}` : ''}
-      </CopyToClipboard>
-    ) || 'DRAFT';
+  let info: JSX.Element | string = (
```
zacharyjhankin: Can delete "| string "

apps/frontend/package.json
yoganandaness: I think the Dockerised app is being used and hence it wouldn't cause problems. I could be wrong as well.

Also I thought it would avoid overloading the node_modules. btw in backend, typescript is situated in devDependencies.

yoganandaness (approved):

jekabs-karklins:
The PR is good, but I did change the title to speciffically indicate that the upgrade is only for Frontend.
And that is what I would like to flag is that the backend still sits on TS5. Although not a big issue, I would suggest uppgrading BE to 6 in the near future to keep both code bases consistent.

Interstingly I found also that `validation` library includes TypeScript as a peer dependnecy from ts-jest
```bash
$npm why typescript
typescript@5.9.3 dev peer
node_modules/typescript
  peer typescript@">=4.3 <7" from ts-jest@29.4.11
  node_modules/ts-jest
    dev ts-jest@"^29.1.1" from the root project
```

jekabs-karklins (approved):

yoganandaness:
> The PR is good, but I did change the title to speciffically indicate that the upgrade is only for Frontend. And that is what I would like to flag is that the backend still sits on TS5. Although not a big issue, I would suggest uppgrading BE to 6 in the near future to keep both code bases consistent.
> 
> Interstingly I found also that `validation` library includes TypeScript as a peer dependnecy from ts-jest
> 
> ```shell
> $npm why typescript
> typescript@5.9.3 dev peer
> node_modules/typescript
>   peer typescript@">=4.3 <7" from ts-jest@29.4.11
>   node_modules/ts-jest
>     dev ts-jest@"^29.1.1" from the root project
> ```

Thanks. Here is the ticket https://github.com/UserOfficeProject/issue-tracker/issues/1707

TCMeldrum (approved):

PR #1746, opened 2026-08-25
build: upgrade React 18 to 19

shivoomiess:
## Description

React upgrade to v19. Includes other upgrades such as dnd, tinymce, react parser, notistack. Added emotion as direct dependency. 

## Motivation and Context

This is essential to facilitate other upgrades including Material Table and MUI. Keeps the stack upto date. 

## How Has This Been Tested

- e2e passed
- build passed
- manually

## Changes

- **`package.json`**
- Forward ref in is no more needed in React 19, hence removed
- React PropTypes is no longer needed in React 19
- No need to explicitly mention JSX.Element
- Latest notistack has out of the box css customization. so removed the definitions in the css file
- Added missing dependency in useEffect

## Depends on

PR 1745


apps/frontend/package.json
```diff
@@ -53,6 +55,10 @@
     "tinymce": "^7.5.1",
     "yup": "^1.7.1"
   },
+  "overrides": {
```
yoganandaness, line 58: Pls note. this will be reverted back on the subsequent [PR](https://github.com/UserOfficeProject/user-office-core/pull/1747/changes) in the stack.

yoganandaness (approved):

apps/frontend/src/components/common/MultistepWizard.tsx
```diff
@@ -24,6 +24,13 @@ interface WizardProps
   shouldCreate: boolean;
 }
 
+type WizardStepProps = {
+  title: string;
+  validationSchema: Yup.ISchema<unknown>;
```
jekabs-karklins: `Yup.ISchema<unknown>` is looser than what this field can accept. It admits `Yup.string()`, `Yup.number()`, `Yup.array()` and anything else implementing `ISchema`, none of which are valid for a Formik wizard step. The old `Yup.AnyObjectSchema` at least constrained this to object schemas.

Our TypeScript 6 migration notes call out this exact tradeoff: `Yup.AnyObjectSchema` stops accepting concrete object schemas under TS 6, and `Yup.ObjectSchema<Record<string, unknown>>` is the replacement that keeps the object constraint, unlike the looser `ISchema<unknown>`.

I applied the change locally and ran `tsc --noEmit`: it compiles with zero new errors, so we can keep the stricter bound for free.

```suggestion
  validationSchema: Yup.ObjectSchema<Record<string, unknown>>;
```

jekabs-karklins: `ISchema<unknown>` is too loose it admits `Yup.string()`, `Yup.number()`, `Yup.array()` and anything else implementing `ISchema`, none of which are valid for a Formik wizard step. For validation schema the object type would be better alternative 

```suggestion
  validationSchema: Yup.ObjectSchema<Record<string, unknown>>;
```

apps/frontend/src/components/App.tsx
```diff
@@ -16,6 +16,25 @@ import clearSession from 'utils/clearSession';
...
+    color: 'var(--PALETTE_PRIMARY_CONTRAST)',
+  },
+  '&.notistack-MuiContent-info': {
+    backgroundColor: 'var(--PALETTE_INFO_MAIN)',
+    color: 'var(--PALETTE_SECONDARY_CONTRAST)',
+  },
+}));
+
 function Root() {
   const notistackRef = React.createRef<SnackbarProvider>();
```
jekabs-karklins: Pre-existing rather than introduced here, but this file is already being modified and React 19 tightened ref semantics

`React.createRef()` is meant for class components. Called inside a function component it allocates a **brand new ref object on every render**, so the ref never persists across renders.

`useRef` is the correct hook

```suggestion
  const notistackRef = React.useRef<SnackbarProvider>(null);
```

jekabs-karklins:
Quiet a few tests are breaking even after re-running twice.

apps/frontend/src/components/common/MultistepWizard.tsx
yoganandaness: Great point.

jekabs-karklins (approved):

PR #1754, opened 2026-08-26
fix: await promises in VisitAuthorization hasReadRights

jekabs-karklins:
The two async checks in `VisitAuthorization.hasReadRights` were never awaited which caused issue that visitor could not load the questionaire.

```ts
return (
  visit.creatorId === agent.id ||                                  // sync
  this.proposalAuth.isMemberOfProposal(agent, visit.proposalPk) || // async, not awaited
  this.visitDataSource.isVisitorOfVisit(agent.id, visit.id)        // async, not awaited
);
```

image

The first line is sync, and the last two is async.

one is false → || moves on
two is a pending Promise → truthy → || short-circuits and returns two itself
three is never evaluated

The method was silently equivalent to:

```ts
return visit.creatorId === agent.id || (await this.proposalAuth.isMemberOfProposal(...));
```

The visitor-list check was dead code for everyone. Proposal members passed on the second check, so this only surfaced for a user on the visitor list who is **not** a proposal member — they saw `Failed to load questionary details` when opening the visit registration.

## How Has This Been Tested

Manually, as the reporting visitor: the visit registration questionary now loads.

## Fixes

SWAP-5726

## Changes

`apps/backend/src/auth/VisitAuthorization.ts` — await both async checks.


apps/backend/src/auth/VisitAuthorization.ts
```diff
@@ -68,8 +68,8 @@ export class VisitAuthorization {
      */
     return (
       visit.creatorId === agent.id ||
-      this.proposalAuth.isMemberOfProposal(agent, visit.proposalPk) ||
-      this.visitDataSource.isVisitorOfVisit(agent.id, visit.id)
+      (await this.proposalAuth.isMemberOfProposal(agent, visit.proposalPk)) ||
```
yoganandaness, line 71: Can we add some parallelization?

const [isMember, isVisitor] = await Promise.all([
  this.proposalAuth.isMemberOfProposal(agent, visit.proposalPk),
  this.visitDataSource.isVisitorOfVisit(agent.id, visit.id),
]);

return visit.creatorId === agent.id || isMember || isVisitor;

yoganandaness (changes requested):

jekabs-karklins:
@copilot please fix the merge conflicts in this pull request.

apps/backend/src/auth/VisitAuthorization.ts
jekabs-karklins: Thats a good tip!

jekabs-karklins:
@DChambersSTFC Could you check this PR when you get a bit of time?

yoganandaness (approved):

PR #1755, opened 2026-08-26
feat: add all feasibility reviews unfeasible state event

DLSValcarcel:
## Description

Adds a new status event used to trigger the change from feasibility review ot the next stage in the workflow.
PROPOSAL_ALL_FEASIBILITY_REVIEWS_UNFEASIBLE

image

## Motivation and Context

In cases where a proposal has multiple instruments associated with it, one of the possible workflow routes is for the proposal NOT to be rejected unless all instrument requests are deemed unfeasible.
In this case, the application needs to check for and trigger when all feasibility reviews are submitted and are unfeasible.

## How Has This Been Tested
Unit tested by creating proposals with multiple instruments and running feasibility reviews with combinations of all and some reviews unfeasible. Partially feasible states were also checked to ensure they behave as feasible.

## Changes
New event "PROPOSAL_ALL_FEASIBILITY_REVIEWS_UNFEASIBLE"
event.enum.ts
applictionEvents.ts

New event handler
customHandlers.ts

New event guard
isEveryFeasibilityReviewUnFeasibleForProposalGuard.ts
isEveryFeasibilityReviewUnFeasibleForProposalGuard.spec.ts


zacharyjhankin:
Manually tested and passed

apps/backend/src/workflowEngine/guards/isEveryFeasibilityReviewUnFeasibleForProposalGuard.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+      { status: TechnicalReviewStatus.UNFEASIBLE },
+    ]);
+
+    const result = await isEveryFeasibilityReviewUnFeasibleForProposalGuard({
+      id: 1,
+    });
+    expect(result).toBe(true);
+  });
+
+  it('returns false if any technical review is sfeasible', async () => {
```
zacharyjhankin, line 50: returns false if any technical review is feasible

apps/backend/src/workflowEngine/guards/isEveryFeasibilityReviewUnFeasibleForProposalGuard.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+    container.resolve = jest.fn((token) => {
+      if (token === Tokens.ReviewDataSource) return mockReviewDataSource;
+
+      return null;
+    }) as typeof container.resolve;
+  });
+
+  it('returns false if no technical reviews found', async () => {
```
zacharyjhankin, line 22: nitpick: 
it is TRUE that all technical reviews are unfeasible as there is no technical reviews
it is also true that all technical reviews are feasiable 

Image

zacharyjhankin: but to avoid a race condition it is probably ok? I don't mind really

zacharyjhankin: I actually do think it is best for the function to return true here. Because it should do what is logically valid, and changing it in the future would be difficult. If someone sets it up expecting it to be false and it is true, I think that is their fault for not understanding predicate logic

zacharyjhankin (approved):
Looks good to me, happy for you to merge or change with my comments. Either way

yoganandaness (approved):

PR #1774, opened 2026-09-03
chore: add dependabot config for backend/frontend dockerfiles

joeshannon:
This also sets the nginx base image to a fixed version but removes the specific alpine constraint.

## Description

Add dependabot config for frontend and backend Dockerfiles. The config broadly matches the existing in terms of schedule. Only minor/patch updates will be made otherwise e.g. it will attempt to update to the latest node major version.

## Motivation and Context

Container base images are only updated manually at the moment.

## How Has This Been Tested

Some manual testing in a different repository.

### DIFF
--- a/apps/frontend/Dockerfile
+++ b/apps/frontend/Dockerfile
@@ -23,7 +23,7 @@ COPY apps/frontend/ /app/frontend
 RUN npm run build
 
 # Stage 1, based on Nginx, to have only the compiled app, ready for production with Nginx
-FROM nginxinc/nginx-unprivileged:alpine3.23
+FROM nginxinc/nginx-unprivileged:1.31.5-alpine
 
 COPY --from=build-stage /app/frontend/build/ /usr/share/nginx/html
 COPY --from=build-stage /app/frontend/build-version.txt /usr/share/nginx/html
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -69,3 +69,21 @@ updates:
     versioning-strategy: increase-if-necessary
     cooldown:
       default-days: 7
+
+  - package-ecosystem: 'docker'
+    directories:
+      - '/apps/backend'
+      - '/apps/frontend'
+    schedule:
+      interval: 'weekly'
+      day: 'sunday'
+      time: '10:00'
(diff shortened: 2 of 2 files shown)


apps/frontend/Dockerfile
```diff
@@ -23,7 +23,7 @@ COPY apps/frontend/ /app/frontend
 RUN npm run build
 
 # Stage 1, based on Nginx, to have only the compiled app, ready for production with Nginx
-FROM nginxinc/nginx-unprivileged:alpine3.23
+FROM nginxinc/nginx-unprivileged:1.31.5-alpine
```
joeshannon, line 26: @jekabs-karklins This was previously pinned to alpine 3.23 with no specific nginx version. Was that for a particular reason/issue?

zacharyjhankin: 1.31.5-alpine3.24

Could be better to pin both? linux and nginx version

zacharyjhankin (approved):

apps/frontend/Dockerfile
jekabs-karklins: @joeshannon thanks for pointing it out. No there is no particular reason just oversight. This would be nice to pin the versions fixed

jekabs-karklins (approved):

apps/frontend/Dockerfile
joeshannon: The complication is that dependabot uses relatively basic [tag matching for docker.](https://github.com/dependabot/dependabot-core/tree/main/docker#supported-tag-schemas) It can only extract one sem ver from the string. That is to say "apline3.14" would be considered a constant suffix.

This is not immediately a problem but at some point nginx will stop building for alpine3.24 once it's unsupported and at that point we will remain on the old version until it's manually incremented.

Also note that currently we don't pin the OS base version for the backend, which is currently `node:22.14.0-alpine`.

Let me know which approach you prefer and we can make both consistent.

joeshannon: I think I will merge this as is for now.
We can look at rationalising the image tags later if we'd like.
I asked around at DLS and some teams have moved to Renovate which appears to be more flexible with update rules.

PR #1781, opened 2026-09-08
fix: visit RabbitMQ messages now use the visitor’s oidcSub

janosbabik:
The visit RabbitMQ messages now use the visitor’s oidcSub.

### DIFF
--- a/apps/backend/src/eventHandlers/messageBroker.ts
+++ b/apps/backend/src/eventHandlers/messageBroker.ts
@@ -319,16 +319,26 @@ export const getVisitMessageData = async (
     Tokens.ProposalDataSource
   );
 
-  const proposal = await proposalDataSource.getProposalByVisitId(
-    visitRegistration.visitId
+  const userDataSource = container.resolve<UserDataSource>(
+    Tokens.UserDataSource
   );
+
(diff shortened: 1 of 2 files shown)


yoganandaness (approved):

apps/backend/src/eventHandlers/messageBroker.ts
```diff
@@ -319,16 +319,26 @@ export const getVisitMessageData = async (
...
+  }
+
   const proposalPayload = await getProposalMessageData(proposal);
 
   const visitJsonMessage = JSON.stringify({
     id: visitRegistration.id,
     startAt: visitRegistration.startsAt,
     endAt: visitRegistration.endsAt,
-    visitorId: visitRegistration.userId.toString(),
+    visitorId: visitor.oidcSub,
```
jekabs-karklins, line 341: suggestion: it would be nice to get STFC sign off this line, as they have OIDC set to '' (empty string). If they are using this for something it could break

jekabs-karklins (approved):
Looks good

zacharyjhankin (approved):
Great change Diamond wants this asap :D

PR #1782, opened 2026-09-08
feat: add list of variables to email template editor

ellen-wright:
## Description
refs: https://github.com/UserOfficeProject/issue-tracker/issues/1616

## Motivation and Context
User Officers would need to know the codebase in order to know what variables were available to be used in an email template and their format. 
This list removes this need and gives a User Officers more autonomy in creating/editing email templates.  

## Changes
Tooltip added with list of variables 

image


zacharyjhankin:
Can we please not add a dependency for people to update this list when they change the object email object. I am sure there will be a way of in code taking the supplied object into that list you want.

Edit: it might require a graphql call to get data from backend

zacharyjhankin:
These variables are actually more complex than the UI makes it seem. Because lots are nested. proposals for example is a list that contains things, you can't actually inform the UO the full complexity of these items just by saying "proposals".
They would need to know it was proposals[0].title

zacharyjhankin:
The ID is not even available.
image

It is generated in the utils file that supplies info to the emails but the actual value isn't sent via mailService.sendMail. (which is where it matters)

ellen-wright:
Part of the Epic for the changes to email templates is to add documentation for User Officers to create templates, this might be best paired with that documentation when it's created. I will close this until that has happened

PR #1801, opened 2026-09-15
refactor: move role tag lookup to query layer

simonfernandes:
## Description
This pull request refactors the code to move the role tag lookup to the query layer. 

## Motivation and Context
This moves tag fetching for calls and instruments from the datasource layer to the query layer, for a better separation of concerns and to standardise the approach with proposal queries. Instead of datasources taking the agent's role ID and fetching its tags, the queries pass the tag IDs directly using the existing `@AgentTags` decorator.

## Changes
- Replaced agentId with tagIds in getInstruments and getUserInstruments methods in InstrumentDataSource.ts.
- Removed getTagsByRoleId method from CallDataSource.ts and InstrumentDataSource.ts.
- Updated the getCalls method in CallDataSource.ts to use tagIds for filtering.
- Added dummyUserOfficerWithDerivedRole and dummyInstrumentScientistWithDerivedRole in UserDataSource.ts.
- Removed unused import statements from CallDataSource.ts and InstrumentDataSource.ts.

## How Has This Been Tested?

Added unit tests, e2e already exist for filtering by tags.

## Fixes Jira Issue

https://github.com/UserOfficeProject/issue-tracker/issues/1641


apps/backend/src/queries/InstrumentQueries.ts
```diff
@@ -39,16 +39,13 @@ export default class InstrumentQueries {
...
+    callIds: number[],
+    @AgentTags tags?: number[]
+  ) {
     if (!callIds || callIds.length === 0) {
-      return await this.dataSource.getInstruments(
-        undefined,
-        undefined,
-        agentRoleId
-      );
+      return await this.dataSource.getInstruments(undefined, undefined, tags);
```
mutambaraf, line 48: Maybe we need to fix the ordering of this function to start with tags since they appear to be less optional here this.dataSource.getInstruments(undefined, undefined, tags) or make it an object.

apps/backend/src/resolvers/types/User.ts
```diff
@@ -206,11 +206,16 @@ export class UserResolver {
 
   @FieldResolver(() => [Instrument])
   async instruments(@Root() user: User, @Ctx() context: ResolverContext) {
-    const agentId = context.user?.currentRole?.id;
+    const agentRoleId = context.user?.currentRole?.id;
+    const tags = agentRoleId
+      ? await context.queries.instrument.roleDataSource.getTagsByRoleId(
+          agentRoleId
+        )
+      : [];
```
mutambaraf, line 219: If we already have tags in the frontend we can pass them as an optional variable here.

yoganandaness (commented):
Good work

yoganandaness (approved):

apps/backend/src/datasources/postgres/CallDataSource.ts
```diff
@@ -54,50 +52,24 @@ export default class PostgresCallDataSource implements CallDataSource {
...
-      if (tagIds.length != 0) {
-        const calls = await database<CallRecord>('call as c')
-          .join('tag_call as tc', 'c.call_id', 'tc.call_id')
-          .whereIn('tc.tag_id', tagIds)
-          .select('c.*');
+    if (tagIds?.length) {
+      const calls = await database<CallRecord>('call as c')
+        .join('tag_call as tc', 'c.call_id', 'tc.call_id')
+        .whereIn('tc.tag_id', tagIds)
+        .select('c.*');
```
jekabs-karklins: breaking: call can have several tags, so this would return duplicates
```suggestion
        .distinct('c.*');
```

apps/backend/src/datasources/postgres/CallDataSource.ts
```diff
@@ -54,50 +52,24 @@ export default class PostgresCallDataSource implements CallDataSource {
...
-
-    if (tags.length != 0) {
-      const tagIds = tags.map((tag) => tag.id);
-      if (tagIds.length != 0) {
-        const calls = await database<CallRecord>('call as c')
-          .join('tag_call as tc', 'c.call_id', 'tc.call_id')
-          .whereIn('tc.tag_id', tagIds)
-          .select('c.*');
+    if (tagIds?.length) {
+      const calls = await database<CallRecord>('call as c')
```
jekabs-karklins: suggestion: this function is badly implemented before you changed it. if tagIds are specified then filter, sortField and sortDirection is completely ignored. Can we just add tag filtering the `query` object we have defined already above?

apps/backend/src/datasources/postgres/CallDataSource.ts
simonfernandes: Good catch, thanks

apps/backend/src/datasources/postgres/CallDataSource.ts
simonfernandes: Very good idea, I failed to spot this one

apps/backend/src/resolvers/types/User.ts
simonfernandes: I had a look into this and the tags aren't coming from the frontend so must always be fetched here

mutambaraf (approved):
LGTM

PR #1808, opened 2026-09-17
fix: show PI name when last name is unavailable

yoganandaness:
## Description
This PR fixes the Principal Investigator column in proposal tables so it is populated when a principal investigator has no last name.

## Motivation and Context
The proposal table renderer previously returned an empty string whenever the principal investigator last name was unavailable, even when a preferred name or first name existed. This made the Principal Investigator column appear empty for valid user records.

## Changes
1. Proposal tables now display the preferred name, or first name, when the last name is unavailable.
2. The existing `Last name, Preferred name` format is preserved when a last name exists.
3. The officer proposal table continues to link the displayed name to the user profile.
4. An E2E regression setup covers a principal investigator without a last name.

## Visual Changes
There are no layout or styling changes. The Principal Investigator cell now displays the available name instead of being empty.

## How Has This Been Tested?

- `npm --prefix apps/frontend run lint` passed.
- `npm --prefix apps/e2e run lint` passed.
- `npm --prefix apps/e2e run cy:run -- --spec cypress/e2e/proposals.cy.ts --config retries=0` was attempted, but the local E2E database setup failed because PostgreSQL reported `relation "users" does not exist`.

## Fixes Jira Issue
Not applicable.


simonfernandes (approved):

jekabs-karklins (commented):
suggestion: can we just add the fix within the getPrefferedName instead of monkeypatching? Just check for the existence of the lastname in the utility method.

TCMeldrum (commented):
Could we move this into a new util function like `getPreferedNameWithLastname`?

